PR #222, opened 2021-05-10
EICNET-765: Hide features field from group creation form.

lramarojaona:
Test:

- Features field should not be visible when creating a new group
- It should be visible when editing an existing group


4hardfire (commented):
Instead of using FormAlter as a name, couldn't we use FormOperations instead? I suppose the class should handle all hooks related to forms.

lramarojaona:
@4hardfire , agreed, just renamed it.

PR #223, opened 2021-05-10
EICNET-760: Add button to create top level wiki page

4hardfire:
### Changes

- Add link to create a new wiki page at the current level;
- Hide book select list from wiki page forms;

### Tests

- Create a group and publish it;
- Go to /admin/content and click on the Book page that was automatically created (<Group name> - Wiki);
- From the book page you should see only 1 link to create the first wiki page: "Add a new wiki page";
- Create the first top level wiki page;
- When viewing the wiki page you should now see 2 links: " **Add a new wiki page at the current level**" and "**Add a new wiki page bellow this level**";

### DIFF
--- a/lib/modules/eic_groups/eic_groups.module
+++ b/lib/modules/eic_groups/eic_groups.module
@@ -7,6 +7,7 @@
 
 use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
 use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Form\FormStateInterface;
 use Drupal\eic_groups\Hooks\EntityOperations;
 
 /**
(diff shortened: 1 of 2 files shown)


didierdemaeyer (changes requested):
@4hardfire Looks good. 
The links to create wiki pages at current level and below this level work and don't show the book select list in the create/edit form.

I think the "Add child page" link (http://localhost:8080/test-group/node/add/book?parent=6) to create a Book as child page should still be removed because these pages don't work in the hierarchy.
Not sure if it is part of this ticket.

4hardfire:
> @4hardfire
> 
> Looks good. The links to create wiki pages at current level and below this level work and don't show the book select list in the create/edit form.
> 
> I think the "Add child page" link (http://localhost:8080/test-group/node/add/book?parent=6) to create a Book as child page should still be removed because these pages don't work in the hierarchy.
> Not sure if it is part of this ticket.

@didierdemaeyer 

I think that can be done in a different ticket. Already informed the PM.

PR #224, opened 2021-05-11
Feature/eicnet 748 group feature group flex roles

lramarojaona:
It's been decided that group flex controls visibility of group only and group overviews can only be viewed by members. This means there's no connection to be made regarding the group visibility.
It's also been decided that enabling/disabling features won't affect global Drupal roles (mainly SA/CA), so we don't set/unset any permissions for those.

However, this might change in the future, so we make this configurable, for both permissions and roles. There is no UI to configure this, just yml files.

Testing:
- Create a group
- Update the group and enable Discussions feature
- GO/GA/GM should have both 'access discussions overview' and 'create group_node:discussion entity' permissions
- Update the group and disable Discussions feature
- GO/GA/GM should lose both 'access discussions overview' and 'create group_node:discussion entity' permissions

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,8 +8,8 @@
     "require": {
         "php": ">=7.2",
         "composer/composer": "1.10.21 as 2.0.12",
-        "composer/semver": "1.7.2 as 3.2.4",
         "composer/installers": "^1.9",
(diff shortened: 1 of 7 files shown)


4hardfire (approved):
Seems ok to me!

PR #225, opened 2021-05-12
EICNET-750: open join method on Group

jaapjan:
**Task 1**
Trusted user should be able to join a Group with setting "Open". Non-trusted authenticated users should not be able to. By default the group_flex module provides an "Open" joining method. However this does not allow for any configuration on site-wide user roles.

Solution: build our own custom "Open" joining method plugin and disable the default open joining method.

**Task 2**
Group owner should receive e-mail notification when trusted user becomes a member.

Solution: react on entity_insert of type group_content and create a message of type `notify_new_member_joined` with the correct field content. I've also refactored slightly the code in eic_messages.module and the relevant Hook classes.

**How to test:**

Task 1:
- [x] Create a group with Open joining method as administrator. Make sure it's published.
- [x] Login as authenticated user.
- [x] Open the group page and notice the group does not have a "join" button in the local actions menu.
- [x] Assign the role "trusted user" to the authenticated user.
- [x] Re-open the group page and notice you can join the group.

Task 2:
- [x] Join the group.
- [x] Run the cron.
- [x] Notice the group owner received a notification e-mail.


lib/modules/eic_messages/eic_messages.module
```diff
@@ -6,20 +6,32 @@
...
 
 /**
  * Implements hook_ENTITY_TYPE_insert().
  */
 function eic_messages_group_insert(EntityInterface $entity) {
-  \Drupal::classResolver(EntityOperations::class)
-    ->groupInsert($entity);
+  /** @var GroupMessageCreator $class */
+  $class = \Drupal::classResolver(GroupMessageCreator::class);
+  $class->groupInsert($entity);
```
4hardfire, line 18: Hi @jaapjan. Is there any reason for this change?

lib/modules/oec_group_flex/src/Plugin/GroupJoiningMethod/TuOpenMethod.php
```diff
@@ -0,0 +1,81 @@
...
+   *
+   * @param \Drupal\group\Entity\GroupTypeInterface $groupType
+   *   The group type.
+   *
+   * @return string
+   *   The group role id.
+   */
+  private function getTrustedUserRoleId(GroupTypeInterface $groupType) {
+    /** @var \Drupal\group\GroupRoleSynchronizer $groupRoleSync */
+    $groupRoleSync = \Drupal::service('group_role.synchronizer');
```
4hardfire, line 76: I think this service can be injected using dependency injection.

lib/modules/eic_messages/eic_messages.module
jaapjan: It's mostly to make the module more future proof. Since I expect (see also PR #226) more code in those classes. I think it's better to separate. Also discussed with Ludovic and he was also ok with this

4hardfire (changes requested):
@jaapjan I left some feedback. Everything looks pretty fine to me, although there are some tiny remarks that you probably need to fix.

lib/modules/eic_messages/eic_messages.module
4hardfire: Makes sense to me. Thanks!

lib/modules/oec_group_flex/src/Plugin/GroupJoiningMethod/TuOpenMethod.php
jaapjan: It was solved in the last commit abfdf31

4hardfire (approved):
Seems fine to me!

PR #226, opened 2021-05-13
EICNET-657: moderated join method on Group

jaapjan:
**Task 1**
Group creator should be able to create a group with joining method "Request membership" so trusted users will be able to request access.

Solution: build our own custom "Request" joining method plugin and disable the default request joining method.

**Task 2**
Group owner should receive e-mail notification when trusted user request membership.

Solution: react on entity_insert of type group_content and create a message of type `notify_new_membership_request` with the correct field content. 

**How to test:**

Task 1:
- [x] Create a group with Request joining method as administrator. Make sure it's published.
- [x] Login as authenticated user.
- [x] Open the group page and notice the group does not have a "request" button in the local actions menu.
- [x] Assign the role "trusted user" to the authenticated user.
- [x] Re-open the group page and notice you can request membership to the group.

Task 2:
- [x] Request membership to the group.
- [x] Run the cron.
- [x] Notice the group owner received a notification e-mail with links to either rejection or accepting.
- [x] Notice the behaviour of these links is correct and the user is either accepted or rejected. See /group/x/members-pending for the membership request overview.


lib/modules/oec_group_flex/src/Plugin/GroupJoiningMethod/TuGroupMembershipRequest.php
```diff
@@ -0,0 +1,91 @@
...
+   *
+   * @param \Drupal\group\Entity\GroupTypeInterface $groupType
+   *   The group type.
+   *
+   * @return string
+   *   The group role id.
+   */
+  private function getTrustedUserRoleId(GroupTypeInterface $groupType) {
+    /** @var \Drupal\group\GroupRoleSynchronizer $groupRoleSync */
+    $groupRoleSync = \Drupal::service('group_role.synchronizer');
```
4hardfire, line 86: Same as PR #225.

4hardfire (changes requested):
There is just one remark that is concerning the dependency injection but besides that everything seems fine to me.

lib/modules/oec_group_flex/src/Plugin/GroupJoiningMethod/TuGroupMembershipRequest.php
jaapjan: Should be done now

jaapjan:
Solved the merge conflicts and moved the eic_messages_group_content_insert to the correct class. 

Also added dependency injection for the group role synchronizer.

4hardfire (approved):
Everything is working fine. Thanks!

PR #227, opened 2021-05-14
EICNET-767: Fix search group no results found

didierdemaeyer:
### Test

* Set "Access permission" for Group - Overviews > Search group page to "Unrestricted" (group permission doesn't work yet)
* "No results found." should not be displayed when there are search results


4hardfire (approved):
It seems fine to me.

PR #230, opened 2021-05-14
EICNET-766: Show fallback icon for group teaser, updates storybook example

toolbarthomas:
This PR updates the group teaser component and will show the group icon as fallback image.
Storybook has been updated and the example can be found at: `Storybook > Group > Group Overview Public`.


4hardfire (approved):
Sorry... I was just in the wrong page. Everything works fine.

PR #232, opened 2021-05-17
EICNET-747: Adapt allowed bundles for each flag

DutchYoda:
This is not finished.
Some flags and settings need to be checked.


config/sync/field.field.user.user.field_full_name.yml
```diff
@@ -0,0 +1,20 @@
+uuid: 0cf58c2c-9355-4dae-aab3-449419e1768e
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.user.field_full_name
```
lramarojaona, line 6: This is old stuff, we don't use field_full_name anymore, can you remove it?

lramarojaona (changes requested):
@DutchYoda ,
Some of those configs need to be removed (not needed).
You have the overview of flags here: https://citnet.tech.ec.europa.eu/CITnet/confluence/display/EICNET/FS+-+Allowed+bundles+in+flags

DutchYoda:
> @DutchYoda ,
> Some of those configs need to be removed (not needed).
> You have the overview of flags here: https://citnet.tech.ec.europa.eu/CITnet/confluence/display/EICNET/FS+-+Allowed+bundles+in+flags

I made the changes.

PR #234, opened 2021-05-19
EICNET-774: custom restricted visibility plugins

jaapjan:
This contains a plugin system for the custom restricted visibility options.

When a group creator selects "Custom restriction" for Group visibility you have the option to restrict on:

- Email domains
- Selected trusted user

Since there will be more custom restrictions in the future this is a perfect opportunity to create a plugin system.

What is done:

- Plugin system for custom restricted visibility options.
- Email domains and restricted users is moved to the new plugin system.
- Update hook to update existing data from existing groups to the new data structure.

Data structure of the options array should now be:
```
plugin_id:
  - key: value
  - key: value
plugin_id: 
  - key: value
  - key: value
```

The plugin system takes care of:

- Checking if the user viewing the group has view access.
- Getting plugin form field.
- Validating plugin form field.
- Setting default values for the form field (based on value in database).
- Getting the form field names which will be used for the private temporary storage in multi-step Group form (membership creation after group creation).

Improvements for the future:
- If we need more form fields (we now assume there is always one addition field with the name '[plugin-id]_conf' we need to improve CustomRestrictedVisibility::getFormStateValues().

**How to test**

- [x] Create some custom restriction groups in develop branch.
- [x] Switch to this branch and run database update hooks.
- [x] Verify it all works as expected on creation, edit and viewing of the group.
- [x] Check the code changes.


lib/modules/oec_group_flex/src/Plugin/CustomRestrictedVisibility/EmailDomains.php
```diff
@@ -0,0 +1,93 @@
...
+use Drupal\group\Entity\GroupTypeInterface;
+use Drupal\oec_group_flex\Annotation\CustomRestrictedVisibility;
+use Drupal\oec_group_flex\GroupVisibilityRecordInterface;
+use Drupal\oec_group_flex\Plugin\CustomRestrictedVisibilityBase;
+
+/**
+ * Provides a 'email_domains' custom restricted visibility.
+ *
+ * @CustomRestrictedVisibility(
+ *  id = "email_domains",
```
4hardfire, line 19: Hi @jaapjan, because the plugin ID doesn't contain the prefix "**restricted_**" the migration with hook_update doesn't work for this plugin. Although, it works for "**restricted_user**" because it contains the prefix from the previous data structure.
One way of fixing this is just add the prexis "**restricted_**" in the plugin annotation or change the hook_update.

lib/modules/oec_group_flex/src/Plugin/CustomRestrictedVisibility/EmailDomains.php
```diff
@@ -0,0 +1,93 @@
+<?php
+
+namespace Drupal\oec_group_flex\Plugin\CustomRestrictedVisibility;
+
+use Drupal\Core\Access\AccessResultNeutral;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\group\Access\GroupAccessResult;
+use Drupal\group\Entity\GroupInterface;
+use Drupal\group\Entity\GroupTypeInterface;
```
4hardfire, line 10: This can be removed.

lib/modules/oec_group_flex/src/Plugin/CustomRestrictedVisibility/EmailDomains.php
```diff
@@ -0,0 +1,93 @@
+<?php
+
+namespace Drupal\oec_group_flex\Plugin\CustomRestrictedVisibility;
+
+use Drupal\Core\Access\AccessResultNeutral;
```
4hardfire, line 5: This can be removed.

4hardfire (changes requested):
To fix:

- [ ] Fix hook_update issue by updating the logic or changing the plugin_id of EmailDomains class;
- [ ] Remove unused namespaces;

lib/modules/oec_group_flex/src/Plugin/CustomRestrictedVisibility/EmailDomains.php
jaapjan: Good catch. I've renamed the plugin id.

jaapjan:
I've made the changes in the last commit. Also made sure to include the StringTranslationTrait 🙈  Small thing, but much neater than just using the t() function.

4hardfire (approved):
@jaapjan Everything looks nice. Thanks!

PR #236, opened 2021-05-20
EICNET-795: Alter the url aliases for nodes that belong to a group

4hardfire:
### Description

This PR contains back-end changes in order to uninstall group_purl module and create a custom token to prefix the group url in the pathauto patterns of all nodes that belong to groups.

Bellow there is an overview list of what has been done:

- [ ] EICNET-796: Uninstall purl and group_purl modules
- [x] EICNET-795: Install subpathauto module
- [ ] EICNET-796: Re-adapt code (block plugins and services)
- [ ] EICNET-796: Re-adapt group related views;
- [ ] EICNET-795: Create custom token **node_group_url** to be used in pathauto patterns in order to prepend the group url to the node url;
- [ ] EICNET-795: Updated URL pattern of Content Type discussions and added new patterns for ~Books and~ Wiki pages.
- [ ] EICNET-795: Custom patch to fix subpathauto installation dependencies (see https://www.drupal.org/project/subpathauto/issues/3188768);
- [ ] EICNET-807: Implement hook_pathauto_alias_alter in order to fix url alias of book nodes that belong to groups.

### Tests

- [x] Create a group and publish it;
- [x] Create a new Discussion in group from the group detail page and check if the Discussion URL is prefixed with the group url;
- [ ] Test if group overview pages can be accessed via the group url alias like: Discussions (**group-url/discussions**), Members (**group-url/members**), Search group (**group-url/search-group**).
- [x] Test if group wiki is accessible via **group-url/wiki**
- [ ] Create a new Wiki page from the group wiki and check if the Wiki page URL is prefixed with **group-url/wiki**;

### Procedures to deploy in STG

- [ ] Before merging this PR, make sure you remove **purl.purl_provider.group_purl_provider.yml** configuration and run **drush cim**;
- [ ] Then you can merge this PR in staging and run the deploy script like usual.


lib/modules/eic_groups/src/Hooks/GroupTokens.php
```diff
@@ -0,0 +1,105 @@
...
+    if ($type == 'eic_groups_tokens') {
+      foreach ($tokens as $name => $original) {
+        switch ($name) {
+          case 'node_group_url':
+            if (isset($data['node'])) {
+              if (($node = $data['node']) && $node instanceof NodeInterface) {
+                if ($group_contents = $this->entityTypeManager->getStorage('group_content')->loadByEntity($node)) {
+                  /** @var \Drupal\group\Entity\GroupContentInterface $group_content */
+                  $group_content = reset($group_contents);
+                  if ($group_content->hasTranslation($node->language())) {
```
eelkeblok, line 83: Should fix: hasTranslation expects a language code (e.g. 'en'), while $node->language() returns a \Drupal\Core\Language\LanguageInterface. I suspect this is the reason I am seeing warnings of the form `Warning: Illegal offset type in Drupal\Core\Entity\ContentEntityBase->hasTranslation()`. I think it can be solved by calling getId() on the language object.

eelkeblok:
Note that the purl and group_purl are still present in the composer.json/composer.lock. They will need to be removed, but I would recommend to do that in a follow-up PR, after this has been deployed, so that the code is still available when uninstalling.

eelkeblok:
Should fix: config/sync/core.entity_form_display.group.group.default.yml still contains a reference to purl (I *think* this is the group prefix..?)

lib/modules/eic_groups/src/EventSubscriber/SearchApiSubscriber.php
```diff
@@ -2,7 +2,7 @@
 
 namespace Drupal\eic_groups\EventSubscriber;
 
-use Drupal\group_purl\Context\GroupPurlContext;
+use Drupal\eic_groups\EICGroupsHelperInterface;
```
eelkeblok, line 5: Could fix: Line 21 below contains a reference to purl.

lib/modules/eic_groups/src/Hooks/Pathauto.php
```diff
@@ -0,0 +1,109 @@
...
+          /** @var \Drupal\group\Entity\GroupContentInterface $group_content */
+          $group_content = reset($group_contents);
+          if ($group_content->hasTranslation($node->language()->getId())) {
+            $group = $group_content->getGroup()->getTranslation($node->language()->getId());
+          }
+          else {
+            $group = $group_content->getGroup();
+          }
+
+          // If the group has an alias we use it has a prefix.
```
eelkeblok, line 93: Could fix: Typo: "...use it *as* a prefix..."

lib/modules/eic_groups/src/Hooks/Pathauto.php
```diff
@@ -0,0 +1,109 @@
...
+          if ($group_alias = $this->aliasStorageHelper->loadBySource($group->toUrl()->getInternalPath(), $group->language()->getId())) {
+            $alias = "{$group_alias['alias']}/wiki";
+            return;
+          }
+
+          // We use pathauto generator to retrieve the group alias so we can
+          // use it has a prefix.
```
eelkeblok, line 100: Could fix: Typo: "...use it *as* a prefix..."

eelkeblok (changes requested):
Should fix (?): When visiting the discussions overview, I do not actually see my created discussion and I do not have any indication this is the overview for my group. However, this may all be unrelated to this PR, so we could mark this as Won't fix too, depending on your judgement. The same basically goes for the search-group suffix.
Should fix: Maybe I misunderstand how this is supposed to work, but when I create a wiki page, it seems to take in the location of the top level; when visiting /group-url/wiki, I get redirected to the URL for the book page, which is /test-group-20210601/test-group-20210601-wiki/my-new-wiki-page-20210601 (/test-group-20210601 is the URL to my group, but the rest was automatically generated; I guess this needs some more work to fix, although if David needs all this too, we could choose to do that in a follow-up).

lib/modules/eic_groups/src/EventSubscriber/SearchApiSubscriber.php
4hardfire: Fixed.

lib/modules/eic_groups/src/Hooks/Pathauto.php
4hardfire: Fixed.

lib/modules/eic_groups/src/Hooks/Pathauto.php
4hardfire: Fixed.

lib/modules/eic_groups/src/EventSubscriber/BookPageRedirectSubscriber.php
```diff
@@ -33,13 +33,13 @@ class BookPageRedirectSubscriber implements EventSubscriberInterface {
   /**
    * Constructs a new BookPageRedirectSubscriber instance.
    *
-   * @param \Drupal\group_purl\Context\GroupPurlContext $group_purl_context
```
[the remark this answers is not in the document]
4hardfire: Fixed.

4hardfire:
> Should fix (?): When visiting the discussions overview, I do not actually see my created discussion and I do not have any indication this is the overview for my group. However, this may all be unrelated to this PR, so we could mark this as Won't fix too, depending on your judgement. The same basically goes for the search-group suffix.
> Should fix: Maybe I misunderstand how this is supposed to work, but when I create a wiki page, it seems to take in the location of the top level; when visiting /group-url/wiki, I get redirected to the URL for the book page, which is /test-group-20210601/test-group-20210601-wiki/my-new-wiki-page-20210601 (/test-group-20210601 is the URL to my group, but the rest was automatically generated; I guess this needs some more work to fix, although if David needs all this too, we could choose to do that in a follow-up).

@eelkeblok thanks for catching those issues :).

- Concerning the issue with the discussion overview, I've tried to create a group + discussion and it shows up in the list of discussions of the group. You can check the image attached. Do you think it might be an issue with your SOLR? In any case I guess it's not related to the PR. In fact we just needed to make sure we could access the URL.
- The issue with the wiki page URL it needs to be fixed (indeed). But I guess we can do it in a follow-up ticket.

image

4hardfire:
> Should fix: config/sync/core.entity_form_display.group.group.default.yml still contains a reference to purl (I _think_ this is the group prefix..?)

This has been fixed. For some reason it was not removed from the configuration after uninstalling group_purl.

eelkeblok:
> * Concerning the issue with the discussion overview, I've tried to create a group + discussion and it shows up in the list of discussions of the group. You can check the image attached. Do you think it might be an issue with your SOLR? In any case I guess it's not related to the PR. In fact we just needed to make sure we could access the URL.

Ah. Missing piece of the puzzle, didn't know this came from SOLR. Never mind then.

eelkeblok (approved):
Approved. Feel free to merge once the conflicts are resolved.

PR #237, opened 2021-05-20
EICNET-750/657: Joining methods available for all visibility types (except private)

jaapjan:
As found in the internal demo meeting the joining method should be available for all visibility types, except private. This solves it.

Test:

- [x] Create group
- [x] Change visibility of the group. 
- [x] Notice the joining methods are always available, except if the group visibility is private.


4hardfire (approved):
It works pretty fine!

PR #238, opened 2021-05-20
EICNET-667: Started matching drupal output with storybook.

xwero:
Filters styling
Teasers styling


didierdemaeyer (approved):
Filters and group teaser look better on the Groups overview page. 

Some functionality of the filters should be implemented in next PR.

PR #240, opened 2021-05-20
Feature/eicnet 637 as an anonymous user, i can't add a comment

fonteijne:


lib/modules/eic_comments/src/Plugin/Block/LoginBlock.php
```diff
@@ -0,0 +1,179 @@
...
+
+    $field_comments = $node->get('field_comments')->first()->getValue();
+    $status = intval($field_comments['status']);
+
+    // Status 0: Comments are "hidden".
+    // Status 1: Comments are "closed".
+    // Status 2: Comments are "open".
+    if ($status !== 2) {
+      return FALSE;
+    }
```
4hardfire, line 174: @fonteijne I recommend using constants here. The comment module already provides it for you to use in **Drupal\comment\Plugin\Field\FieldType\CommentItemInterface**:

- **CommentItemInterface::HIDDEN** - 0
- **CommentItemInterface::CLOSED** - 1
- **CommentItemInterface::OPEN** - 2

This way, if the status number changes you don't need to worry about changing everywhere :).

4hardfire (approved):
Seems ok to me.

PR #244, opened 2021-05-24
EICNET-776: Autocomplete user fields may filter untrusted users

jaapjan:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-776

I noticed we use autocomplete fields for reference user entities. I think we should use a generic view for this. The view should have two displays:

* Full name (all roles)
* Full name (trusted users)

You can then use one of the View displays if you want creators to only select trusted users or the other if you want the creator to be able to select all authenticated users. Also this is an opportunity to remove the entity reference display from the user_admin_people view. 

Because the realname module is overriding all the autocomplete endpoints for users it didn't respect the View settings. I've applied a patch for that. We still want the realname module to do it's magic when we have a `default` or `default:user` entity reference select field. But if we have a working view we should use the Entity Reference View settings and thus fallback on the default autocomplete behaviour which makes sure the View is used. 

See: https://www.drupal.org/project/realname/issues/3059311

How to test:
- Do a `composer install` before testing locally.
- Check code changes.
- Make sure all the changed fields still work.
- Test the custom restricted visibility feature on a group. Make sure you can't select untrusted users.
- Test the author field on a Node.
- Test the quote user field on Paragraph.
- Test the user reference on the Contributor paragraph.


fonteijne (approved):
Approved with one minor bug:
~User 0 is still being shown in `full_name_all`, even tho the filter `user id != 0` (respectively) is active.~

This was a user error. No bugs found.

PR #247, opened 2021-05-26
Feature/eicnet 788 sa scm can edit a comment from anyone else

fonteijne:
Updated the logic for showing the buttons to edit and remove the comments.

### DIFF
--- a/lib/themes/eic_community/includes/preprocess/comments/comment.inc
+++ b/lib/themes/eic_community/includes/preprocess/comments/comment.inc
@@ -65,17 +65,13 @@ function eic_community_preprocess_comment(&$variables) {
 
   // Stored in render array as twig does some logic with it.
   $variables['comment']['can_edit_comment'] =
+  $variables['comment']['can_remove_comment'] =
     (
       $current_user->hasPermission('edit own comments')
       && $current_user->id() == $comment->getOwnerId()
       && $num_of_replies === 0
-    );
-
-  $variables['comment']['can_remove_comment'] =
-    $variables['comment']['can_edit_comment']
-    || (
+    ) || (
       $current_user->hasPermission('administer comments')
-      && $num_of_replies === 0
     );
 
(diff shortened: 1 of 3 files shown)


lib/themes/eic_community/includes/preprocess/comments/comment.inc
```diff
@@ -65,17 +65,13 @@ function eic_community_preprocess_comment(&$variables) {
...
-
-  $variables['comment']['can_remove_comment'] =
-    $variables['comment']['can_edit_comment']
-    || (
+    ) || (
       $current_user->hasPermission('administer comments')
-      && $num_of_replies === 0
     );
 
   if ($variables['comment']['can_edit_comment']) {
```
4hardfire, line 77: In the future we should consider using the getOperations method to retrieve the comment links as follow:

`$operations = \Drupal::entityTypeManager()->getListBuilder('comment')->getOperations($comment);

  foreach ($operations as $key => $operation) {
    $variables['comment']['can_' . $key . '_comment'] = TRUE;
    $variables['comment'][$key . '_path'] = $operation['url']->toString();
  }`

Note that we also need to replace the keys "**can_remove_comment**" and "**remove_path**" (in **/eic_community/patters/compositions/comment/comment-base** with "**can_delete_comment**" and "**delete_path**" respectively otherwise the delete link will never show up.

This way the comment links are retrieved based on the user access (that I suppose it will be implemented in https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-728).

4hardfire (approved):
Since the comment accesses will be handled in a different ticket I will just approve this one :).

lib/themes/eic_community/includes/preprocess/comments/comment.inc
fonteijne: Thanks for reviewing! Indeed, the way we retrieve the user access is changed in 728 since this also implements the group permissions. Will take note of your snippet!

PR #248, opened 2021-05-26
EICNET-777: Implement search suggestions (Did you mean?)

didierdemaeyer:
Installed the `search_api_spellcheck` module and added the "Did you mean" spellcheck in the header of the global and group search pages.

### Issues

The spellcheck block is displayed on the group search page (/{group}/search-group), but not on the global search page (/search).


4hardfire (commented):
@didierdemaeyer, I've added the missing header in the twig template used in the global view.

composer.json
```diff
@@ -48,7 +48,7 @@
         "drupal/scheduler": "^1.3",
         "drupal/scheduler_content_moderation_integration": "^1.3",
         "drupal/search_api_solr": "^4.1",
-        "drupal/search_api_sorts": "^1.0",
```
4hardfire, line 51: @didierdemaeyer Did you remove the module **search_api_sorts** by mistake? I checked all the configurations and the module seems to be used.
Can you have a look?

4hardfire (changes requested):

composer.json
didierdemaeyer: Yes, I must've removed that by accident. I've re-added it.

4hardfire (approved):
Seems ok to me!

PR #249, opened 2021-05-26
EICNET-762 : Restricted group follow-ups

jaapjan:
Fixes:

- Validation of the custom restricted visibility option. At least one needs to be selected. 
bd775c6
- Checks if the Group is published before giving access. c7b6344 3f71c71
- Retrieve the group outsider roles via a service provided by Group. 
e8295dc
- Small notice when saving a group with community member visibility. 
e0b270c
- We forgot to change the key email_domains_conf to restricted_email_domains_conf . The access is not working for this plugin right now. See: b8e603a
- Making visibility plugin work on Group Type level again.
4420c45
da4f495
14ed336
- Added Group Visibility settings on `/admin/group/visibility-settings`.


4hardfire (approved):
Everything working like a charm. Thanks :)

PR #250, opened 2021-05-26
EICNET-808: Create banner fragment inline in page.

didierdemaeyer:
Changed field_banner form field so Banner fragments can be create inline or referenced in the Banner paragraph.

Test:
* /node/add/page


4hardfire (approved):
Seems ok to me.

PR #252, opened 2021-05-27
EICNET-809: Update text link of buttons to create wiki pages.

4hardfire:
### Changes

- [ ] EICNET-809: Update text link of buttons to create wiki pages.

### Test

- [ ] Create a new group
- [ ] Navigate to the book page of the group (you can use the **/admin/content** view) and create a new wiki page from there
- [ ] Check if the text links to add wiki pages are as follow:  "Add a new page on the current level" and "Add a new wiki page below this page".


DutchYoda (approved):
Looks good to me.

PR #254, opened 2021-05-28
EICNET-811: Overview groups - split up search input and items per page selection

4hardfire:
### Description

This PR introduces a solution to split up search input and items per page selection in overview pages. Note that this is only working in Groups overview page (/groups).

- [ ] ~Place duplicated views exposed form in groups overview (Groups overview context - admin/structure/context/groups_overview)~
- [ ] ~Make sure views exposed forms have unique id attributes when shown multiple times on the page.~
- [ ] Create custom block plugin (**eic_theme_helper_views_filter_items_per_page**) to expose views "Item per page" filter in a block and turn it as a list of links;
- [ ] Place block **eic_theme_helper_views_filter_items_per_page** in groups overview page.
- [ ] Remove "items_per_page" filter from views_exposed_form (see hook **eic_groups_form_alter**).

### Tests

- [ ] Go to /groups and check

### DIFF
--- a/lib/modules/eic_theme_helper/src/Plugin/Block/ViewsFilterItemsPerPageBlock.php
+++ b/lib/modules/eic_theme_helper/src/Plugin/Block/ViewsFilterItemsPerPageBlock.php
@@ -1,0 +1,182 @@
+<?php
+
+namespace Drupal\eic_theme_helper\Plugin\Block;
+
+use Drupal\Core\Block\BlockBase;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Link;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Url;
+use Drupal\views\ViewExecutable;
+use Drupal\views\Views;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\HttpFoundation\RequestStack;
+
+/**
+ * Provides a ViewsFilterItemsPerPageBlock block.
+ *
+ * @Block(
+ *   id = "eic_theme_helper_views_filter_items_per_page",
+ *   admin_label = @Translation("Views filter items per page block"),
+ *   category = @Translation("OpenEuropa")
+ * )
+ */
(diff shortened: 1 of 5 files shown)


lib/modules/eic_theme_helper/src/Plugin/Block/ViewsFilterItemsPerPageBlock.php
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    if (empty($links)) {
+      return $build;
+    }
+
+    $build['content'] = [
+      '#theme' => 'item_list',
+      '#list_type' => 'ul',
+      '#title' => $this->t('Showing'),
```
lramarojaona, line 146: @4hardfire , is there any way to get the total number of results in this block?
The design shows: "Showing X of 83"

lib/modules/eic_theme_helper/src/Plugin/Block/ViewsFilterItemsPerPageBlock.php
```diff
@@ -0,0 +1,182 @@
...
+
+    // Get current route name.
+    $current_route_name = $this->routeMatch->getRouteName();
+
+    // We check if the current route is actually a view page.
+    if ($current_route_name !== "view.$view_id.$display_id") {
+      return $build;
+    }
+
+    // // Get the view by id and initializes the right display.
```
lramarojaona, line 117: Double comments here ;)

lramarojaona (changes requested):

lib/modules/eic_theme_helper/src/Plugin/Block/ViewsFilterItemsPerPageBlock.php
4hardfire: Done.

lib/modules/eic_theme_helper/src/Plugin/Block/ViewsFilterItemsPerPageBlock.php
4hardfire: @lramarojaona, In fact I didn't realize we had the total number of results in front of if. If we need to have everything in the same wrapper this has to be done in the header of the view and transform this block in a view header plugin or something like that. Otherwise there is no way of showing it without executing the view twice.

PR #255, opened 2021-05-28
Feature/eicnet 799 author of a discussion is also shown in the list of contributors

fonteijne:


lib/modules/eic_content/modules/eic_content_discussion/eic_content_discussion.module
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Implements hook_ENTITY_TYPE_insert().
+ */
+function eic_content_discussion_node_insert(\Drupal\node\NodeInterface $entity) {
+  if ('discussion' === $entity->bundle()) {
+    \Drupal::classResolver(EntityOperations::class)
+      ->discussionInsert($entity);
+  }
+}
```
4hardfire, line 22: Hi @fonteijne,
This works like a charm, however I would suggest implementing the logic using the **hook_entity_presave()**. This way you avoid saving the entity twice.

4hardfire (changes requested):

lib/modules/eic_content/modules/eic_content_discussion/eic_content_discussion.module
fonteijne: Nice! Just a note from our call:
use the `isNew()` to determine whether the entity is created or updated.

fonteijne: - [ ] Use `hook_entity_presave()`
- [ ] Use `isNew()` to on run on the "insert" operation.

4hardfire (approved):
This seems ok now.

PR #256, opened 2021-05-28
Feature/eicnet 765 hide group features on creation form

lramarojaona:
Testing procedure:

- Create a group
- The "Discussions" group menu item should be displayed
- Edit the group => Discussions feature should be enabled
- Edit group permissions => "Access Discussions overview" permission should be enable for Member role

- [ ] Need to check why config doens't have uuid


4hardfire (changes requested):
Hi @lramarojaona, I get an error when trying to create a group:

`TypeError: Argument 1 passed to Drupal\oec_group_features\GroupFeaturePluginBase::removeRolePermissionsFromGroup() must be an instance of Drupal\group_permissions\Entity\GroupPermission, null given, called in /var/www/html/lib/modules/eic_groups/src/Plugin/GroupFeature/GroupDiscussions.php on line 76 in Drupal\oec_group_features\GroupFeaturePluginBase->removeRolePermissionsFromGroup() (line 234 of /var/www/html/lib/modules/oec_group_features/src/GroupFeaturePluginBase.php)`

Can you have a look?

4hardfire (approved):
The group feature is working fine now. Thanks!

PR #259, opened 2021-05-31
EICNET_805: Made facts and figures block more compact.

xwero:


4hardfire (approved):
It seems ok to me.
Side note: I've tested changing the value to 1,5 and 6 and apparently doesn't work for those cases.

xwero:
Only 3 and 4 are created in css. It is a small change to create the other column classes.

xwero:
Forgot to mention this branch is based on the EICNET-795 branch, so don't merge this before that branch is merged.

PR #260, opened 2021-05-31
EICNET-801: Refactor logic that adds contributors based on comments

4hardfire:
### Description

Whenever a user leaves a comment on a discussion, the same user would automatically become a contributor. This logic has been implemented by updating the field_related_contributors of a discussion node everytime a new comment was inserted in the database. This workflow is working however, it assumes the owner cannot edit his discussion at the same time a user is commenting otherwise an error would be thrown.

This PR removes the logic responsible for updating the field_related_contributors (when a user leaves a comment on a discussion) and dinamically adds the comment contributors to the renderable array that will be sent to the theme variables.
The field_related_contributors is still part of the discussion content type which still allows the owner to manually add contributors.

### Merge procedures:

- [x] This PR should only be merged after #255 

### Tests

Visually we won't see any changes because the front-end hasn't been implemented yet, but this PR can be tested by debugging lines 90 and 91 of the preprocess function **eic_community_preprocess_node__discussion** (see file **node--discussion.inc**):

- [ ] Create a new discussion
- [ ] Debug eic_community_preprocess_node__discussion and check if variable **$variables['contributors']** has 1 contributor (the owner)
- [ ] Create a new trusted user and comment the discussion at least twice.
- [ ] Debug eic_community_preprocess_node__discussion again and check if variable **$variables['contributors']** has 2 contributors (the owner + the user who commented twice)
- [ ] Test adding more comments with different users and see if it works as expected.


lib/modules/eic_content/modules/eic_content_discussion/src/Hooks/EntityOperations.php
```diff
@@ -0,0 +1,217 @@
...
+      ->fields('c', ['uid']);
+    $query->condition('c.entity_id', $this->discussion->id());
+    $query->condition('c.entity_type', 'node');
+    // Skip anonymous users.
+    $query->condition('c.uid', 0, '<>');
+    // We group by uid to avoid duplicated results.
+    $query->groupBy('c.uid');
+    $results = $query->execute()->fetchAll(\PDO::FETCH_ASSOC);
+
+    if (empty($results)) {
```
Dromadoros, line 210: What do you think about removing this "if" and directly return value ?

```suggestion
   return $results ?: FALSE;
```

4hardfire: Nice one! Thanks :)

Dromadoros (approved):

PR #261, opened 2021-06-01
EICNET-728: Define group permissions for CRUD operations on comments

fonteijne:
Changes:

- Created new custom module `oec_group_comments`.
- Created new group permissions for viewing, posting and editing (own/all) comments.
- Created `OecGroupCommentsAccessControlHandler` to check for accessing operations on comments.
- Implement Field Formatter to remove the display of the post comment form and/or comments on a node. See Field Formatter description below.
- Added the new field formatter on all the nodes as view display (of field comments).

**Access control handling**
Access control is implemented according to specifications. With the exception of requesting to delete or reporting a comment.

To summarise the access control is as follows:

1. User has site-wide permission "Administer comment" everything is allowed. Returned early.
2. Comment on Entity not posted in Group = Default Drupal Core access control handling. Returned early.
3. View operation: Allowed if user has group-wide permission "view comments" on any of the Groups the given Commented Node was posted in.
4. Update operation: Allowed if user has group-wide permission "edit own comments" on any of the Groups the given Commented Node was posted in AND there are no replies on the comment.
5. Update operation: Allowed if user has group-wide permission "edit all comments" on any of the Groups the given Commented Node was posted in.
6. Delete operation: always forbidden (except if the access allowed was returned in step 1 or step 2 already).

**Field Formatter**
We were running into the issue that the comments access checks are not fully implemented in D8/D9 as they are "supposed to". See: https://www.drupal.org/project/drupal/issues/2879087. 

This is now fixed by creating a custom comment field formatter. This formatter checks for the "post comments" and "view comments" permission in a Group and if the current user does not have the necessary permission we show the "There are currently no comments." message.

**How to test**
To test this through, add a user to a group and start toying around with the permissions and see what you can achieve with the comments. Make sure to test all the access control situations as described above.


lib/modules/oec_group_comments/src/OecGroupCommentsAccessControlHandler.php
```diff
@@ -0,0 +1,125 @@
...
+        $access = AccessResult::neutral();
+    }
+
+    return $access;
+  }
+
+  /**
+   * Helper function to fetch comment reply count.
+   */
+  protected function commentReplyCount($cid, $nid, $entity_type): int {
```
Dromadoros, line 115: missing params in docblock

lib/modules/oec_group_comments/src/Plugin/Field/FieldFormatter/CommentGroupContentFormatter.php
```diff
@@ -0,0 +1,186 @@
...
+      $third_party_settings,
+      $current_user,
+      $entity_type_manager,
+      $entity_form_builder,
+      $route_match,
+      $entity_display_repository
+    );
+
+    $this->renderer = $renderer;
+    $this->isXmlHttpRequest = $is_xml_http_request;
```
Dromadoros, line 142: Not sure but it seems that this variable is never used ?

lib/modules/oec_group_comments/src/Plugin/Field/FieldFormatter/CommentGroupContentFormatter.php
```diff
@@ -0,0 +1,186 @@
...
+    $output = parent::viewElements($items, $langcode);
+    $entity = $items->getEntity();
+
+    // Exclude entities without the set id.
+    if (!empty($entity->id())) {
+      $group_contents = GroupContent::loadByEntity($entity);
+    }
+
+    // Check if the content is posted in groups.
+    if (!empty($group_contents)) {
```
Dromadoros, line 159: What do you think about doing the inverse logic ? For example

```suggestion
    if (empty($group_contents)) {
        return $output;
    }
```

I found it more readable but it's a personal opinion I guess

lib/modules/oec_group_comments/src/Plugin/Field/FieldFormatter/CommentGroupContentFormatter.php
fonteijne: Good catch! I'll remove this.

lib/modules/oec_group_comments/src/Plugin/Field/FieldFormatter/CommentGroupContentFormatter.php
fonteijne: Agreed. Will update this.

lib/modules/oec_group_comments/src/OecGroupCommentsAccessControlHandler.php
fonteijne: Done.

Dromadoros (approved):

PR #263, opened 2021-06-02
Feature/eicnet 773 show group menu

xwero:
This branch is based on EICNET-795.


lib/modules/eic_groups/src/Hooks/GroupTokens.php
```diff
@@ -0,0 +1,104 @@
...
+      ],
+    ];
+  }
+
+  /**
+   * Implements hook_tokens().
+   */
+  public function tokens($type, $tokens, array $data, array $options, BubbleableMetadata $bubbleable_metadata) {
+    $replacements = [];
+    if ($type == 'eic_groups_tokens') {
```
Dromadoros, line 74: better to use strict equality to be consitant I guess

lib/modules/eic_groups/src/Hooks/GroupTokens.php
```diff
@@ -0,0 +1,104 @@
...
+      foreach ($tokens as $name => $original) {
+        switch ($name) {
```
Dromadoros, line 76: It seems that we have only one case in our switch, isn't it a bit overkill ? 

maybe better to use directly an if 

```suggestion
        if ('node_group_url' === $name && isset($data['node']) && ($node = $data['node']) && $node instanceof NodeInterface)
```

So we avoid the if-tree, what do you think ?

lib/modules/eic_groups/src/Hooks/GroupTokens.php
```diff
@@ -0,0 +1,104 @@
...
+          case 'node_group_url':
+            if (isset($data['node'])) {
+              if (($node = $data['node']) && $node instanceof NodeInterface) {
+                // If node belongs to a group.
+                if ($group_contents = $this->entityTypeManager->getStorage('group_content')->loadByEntity($node)) {
+                  /** @var \Drupal\group\Entity\GroupContentInterface $group_content */
+                  $group_content = reset($group_contents);
+
+                  if ($group_content->hasTranslation($node->language()->getId())) {
```
Dromadoros, line 85: I see that we have this group get translation logic in 2 different places for now, isn't it beter to have a service for it ? What do you think ? Maybe there is an existing group service

lib/themes/eic_community/includes/preprocess/blocks/block.inc
```diff
@@ -50,4 +50,8 @@ function eic_community_preprocess_block(array &$variables) {
     // Adds cache tags.
     $variables['#cache']['tags'] = $cache_tags;
   }
+
+  if($variables['base_plugin_id'] == 'group_content_menu') {
```
Dromadoros, line 54: Strict equal to be consistant is better and space after if  so code can be more readable

lib/themes/eic_community/includes/preprocess/blocks/block--group-menu.inc
```diff
@@ -0,0 +1,28 @@
...
+use Drupal\Core\Cache\Cache;
+use Drupal\eic_community\ValueObject\ImageValueObject;
+
+/**
+ * Implements hook_preprocess_block() for group menu.
+ */
+function eic_community_preprocess_block__group_menu(&$variables, &$cache_tags = []) {
+  $items = [];
+
+  foreach($variables['content']['#items'] as $link) {
```
Dromadoros, line 17: Same fo the space after foreach

xwero: Changed

lib/themes/eic_community/includes/preprocess/blocks/block.inc
xwero: Changed

lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
```diff
@@ -81,7 +81,9 @@ public function build() {
 
     // Do nothing if no group was found in the context or in the current route.
     if ((!$group = $this->getContextValue('group')) || !$group->id()) {
-      return $build;
+      if (!$group = $this->eicGroupsHelper->getGroupFromRoute()) {
```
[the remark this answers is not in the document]
xwero: This should be picked up by the person who wrote the code.

lib/modules/eic_groups/src/Hooks/GroupTokens.php
xwero: This should be picked up by the person who wrote the code.

lib/modules/eic_groups/src/Hooks/GroupTokens.php
xwero: This should be picked up by the person who wrote the code.

lib/modules/eic_groups/src/Hooks/GroupTokens.php
xwero: This should be picked up by the person who wrote the code.

PR #264, opened 2021-06-02
EICNET-768: Fixes padding for card group teaser.

toolbarthomas:
This PR should fix the issue where the teaser had incorrect padding for related groups within a group template. This bug was only visible for the overview sections with a coloured background.


xwero (approved):
The changes work in drupal too.

PR #265, opened 2021-06-03
feature/EICNET-782-updates-editorial-header

toolbarthomas:
This PR updates the editorial header component that is used within the Group Overview pages; and introduces the flags option that will display links in the sidebar of the component. The statistics have been moved within the new sidebar and the viewport for mobile devices has been finetuned.

Storybook has been updated to include the examples within `Storybook > Compositions > Editorial Header` and `Storybook > Bundles > Group > Group Overview Public/Private`.


lib/themes/eic_community/patterns/compositions/editorial-header.html.twig
```diff
@@ -28,113 +28,172 @@
...
+                        type: action.icon.type,
+                        name: action.icon.name,
+                        size: 'xs',
+                      } %}
+                    {% endif %}
+
+                    {% set trigger = '' %}
+
+                    {% if action.icon is not empty %}
+
```
korhankale, line 78: Aren't we missing something here? Nothing happens when action.icon is empty

toolbarthomas: No, this was a leftover of the development for this ticket, I've removed it since I already use the fallback during the usage of action.icon.

korhankale: Awesome, thank you for the fix 😄

korhankale (approved):

PR #266, opened 2021-06-03
EICNET-784: Implements state banner composition.

toolbarthomas:
This PR implements the state banner composition that should be used within the group overview page.
I've included an example withing Storybook: `Storybook > Compositions > State Banner` and `Storybook > Bundles > Group > Group Overview Pending`.

The actual dismiss button can be displayed optionally and can be constructed as an ECL button or link component. The current example uses an anchor element for this toggle since I'm not sure where we should define the logic to dismiss messages. Doing this in combination with a button is also possible by simply not defining the link object within the dismiss context.


4hardfire (approved):
Seems ok to me!

PR #271, opened 2021-06-04
[EICNET-847] feat: create overview search component

Dromadoros:
Introducing React framework with Laravel mix as assets compiler. The goal of this PR is to create a generic search block for most of search of the project. 

Here are different points that have been developped to achieve this feature (mainly in the new module eic_search) :

- Create a block with custom config, so we can select the source (entity type) of the search, for example : **Group**. It will automatically update available config on the block with depending facets/sorting
- Create different services tagged and a collector that will automatically collect all tagged sources so we can use easily different sources in the block config
- This block will be decoupled in react, so we have the main component "Overview/index.js" who contains different methods to get/update data.
- A new endpoint via controller has been created to get data from solr through search API 
- Some complex data (like group visibility) have been calculated before sending to solr (hook_search_api_solr_documents_alter)
- Enable the interface translation to edit all labels from front-end easily

You need to have at least the node version 12.14 to run the asset compiling (npm run react-production) in the theme root

image
image


config/sync/search_api.index.global.yml
```diff
@@ -171,6 +178,8 @@ field_settings:
     datasource_id: 'entity:node'
     property_path: status
     type: boolean
+    indexed_locked: true
+    type_locked: true
```
lramarojaona, line 182: What's this for?
Just trying to understand what it does.

config/sync/search_api.index.global.yml
```diff
@@ -198,36 +207,6 @@ field_settings:
...
-      module:
-        - group
-  group_content_type:
-    label: '[Group content] Group content type'
-    datasource_id: 'entity:group_content'
-    property_path: type
-    type: string
-    dependencies:
-      module:
-        - group
```
lramarojaona, line 230: Is there a reason to remove the 4 items above?

- flag_recommend_group_count
- group_content_entity_id
- group_content_gid
- group_content_type

Not sure if the last 3 are still being used, but let's keep them to be sure.
The first one (recommend count) AFAIK is needed.

config/sync/search_api.index.global.yml
```diff
@@ -260,11 +239,19 @@ field_settings:
     dependencies:
       config:
         - field.storage.group.field_vocab_topics
+  group_id:
+    label: ID
```
lramarojaona, line 243: Please prefix all fields with their entity type for consistency.
`[Group] ID` in this case

config/sync/search_api.index.global.yml
```diff
@@ -260,11 +239,19 @@ field_settings:
...
+    type: integer
+    dependencies:
+      module:
+        - group
   group_label:
     label: '[Group] Title'
     datasource_id: 'entity:group'
     property_path: label
-    type: text
+    type: string
```
lramarojaona, line 254: I wouldn't change this, please create a new one with String (we already had the case where we needed to have the same field in both modes, string and text)
`[Content] Body - Fulltext => content_field_body_fulltext`

config/sync/search_api.index.global.yml
```diff
@@ -277,6 +264,28 @@ field_settings:
     dependencies:
       module:
         - group
+  group_teaser:
+    label: 'Thumbnail image » Media » Image » File » URI » Root-relative file URL'
```
lramarojaona, line 268: Same here, please prefix all fields with their entity type :)

config/sync/search_api.index.global.yml
```diff
@@ -383,29 +467,31 @@ field_settings:
...
+        - field.storage.user.field_media
+      module:
+        - file
+        - media
 datasource_settings:
   'entity:content_moderation_state':
     languages:
       default: true
       selected: {  }
-  'entity:flagging':
```
lramarojaona, line 391: I guess this is also related to the recommend count

config/sync/search_api.index.global.yml
```diff
@@ -443,6 +529,10 @@ processor_settings:
   language_with_fallback: {  }
   rendered_item: {  }
   reverse_entity_references: {  }
+  role_filter:
```
lramarojaona, line 532: Do we need this?
We might have content that belong to the anonymous user in the future (if a user is being deleted from the platform)

lib/modules/eic_search/eic_search.routing.yml
```diff
@@ -0,0 +1,7 @@
+eic_groups.solr_search:
+  path: '/api/solr/search'
+  defaults:
+    _title: 'Solr search'
+    _controller: '\Drupal\eic_search\Controller\SolrSearchController::search'
+  requirements:
+    _access: 'TRUE'
```
lramarojaona, line 7: Maybe use the "access content" permission here for more flexibility (in case they want to restrict access to logged in users only, which was still the case on D7 some time ago)

lib/modules/eic_search/src/Plugin/Block/SearchOverviewBlock.php
```diff
@@ -0,0 +1,336 @@
...
+ *
+ * @Block(
+ *   id = "eic_search_overview",
+ *   admin_label = @Translation("EIC Search Overview"),
+ *   category = @Translation("European Innovation Council"),
+ * )
+ */
+class SearchOverviewBlock extends BlockBase implements ContainerFactoryPluginInterface {
+
+  const AVAILABLE_TYPES = ['group', 'story', 'event', 'organisation', 'global'];
```
lramarojaona, line 28: Is this used somewhere?

lib/modules/eic_search/src/Search/Sources/GroupSourceType.php
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  /**
+   * @inheritDoc
+   */
+  public function getAvailableFacets(): array {
+    return [
+      'ss_group_topic_name',
+      'ss_group_label',
+      'ss_group_user_fullname',
```
lramarojaona, line 40: We should integrate the labels here, as the current isn't ideal (translating variables and the need to enable locale module just for that)

lib/themes/eic_community/react/components/Block/Overview/Search/Field/SearchField.js
```diff
@@ -0,0 +1,35 @@
...
+
+  handleChange(e) {
+    this.props.searchSolr(e.target.value)
+  }
+
+  render() {
+    return (
+      <div className="ecl-filter-sidebar__item">
+                          <span className="ecl-filter-sidebar__item-label" tabIndex="-1">
+                            {this.props.translations.search_text}
```
lramarojaona, line 18: I think we don't need this, but a placeholder in the input directly.
Also, it would be nice to configure the placeholder text from the block configuration (this will also allow to be translated in the block config if needed in the future)

lramarojaona (changes requested):

config/sync/search_api.index.global.yml
Dromadoros: This one was needed for users overview which show by default a user 0 without any data but I will figured it out another way to filter this one for users list

lib/modules/eic_search/src/Plugin/Block/SearchOverviewBlock.php
Dromadoros: It was old stuff, will remove that

lib/modules/eic_search/src/Search/Sources/GroupSourceType.php
Dromadoros: I see what you mean for those fields but there are a lot of translated labels that needs to be editable in the admin and not a good idea to place all labels editable in the block config cause that means that for each block we will need to set again all labels one by one. 

For example the label : "clear all" or other labels. I don't think it's an issue to enable the translation interface, I'm sure in long term it will be good to have this for our complex features

lib/themes/eic_community/react/components/Block/Overview/Search/Field/SearchField.js
Dromadoros: In the design we have a label above but you're right for the placeholder. 

But it's strange to have a placeholder AND a label

config/sync/search_api.index.global.yml
Dromadoros: I will re-add it, I just didn't need it for now

lib/themes/eic_community/react/components/Block/Overview/Search/Field/SearchField.js
lramarojaona: Ok, I was based on the wireframe and not the design

lib/modules/eic_search/src/Search/Sources/GroupSourceType.php
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * @inheritDoc
+   */
+  public function getAvailableSortOptions(): array {
+    return [
+      'timestamp' => t('Created date', [], ['context' => 'eic_search']),
+      'ss_group_label_string' => t('Group label', [], ['context' => 'eic_search']),
+      'ss_group_user_fullname' => t('Fullname', [], ['context' => 'eic_search']),
+    ];
```
lramarojaona, line 52: It seems the labels for the sort options are not taken into account yet.

lramarojaona (changes requested):

lib/modules/eic_search/src/Search/Sources/GroupSourceType.php
Dromadoros: Yes for the admin but on the FE is more complex, I need 2 translations for each sort cause there is ASC and DESC option. Look at my eic_search.module the trick I did

lramarojaona: Ok, then I suggest the following:
```php
t('@sort_option DESC', ['@sort_option' => $sort_option], ['context' => 'eic_search']);
```
So we avoid using variables in `t()` function and we can disable the locale module.

lib/modules/eic_search/src/Plugin/Block/SearchOverviewBlock.php
```diff
@@ -0,0 +1,334 @@
...
+   */
+  public function updateSourceConfig(array &$form, FormStateInterface $form_state) {
+    $current_source_value = $form_state->getValue('settings')['search']['source_type'] ?: GroupSourceType::class;
+    $response = new AjaxResponse();
+
+    /** @var SourceTypeInterface $current_source */
+    $current_source = $this->getCurrentSource($current_source_value);
+
+    $css = ['display' => 'none'];
+
```
korhankale, line 176: Seems you have to much line breaks here

lib/modules/eic_search/src/Search/Sources/GroupSourceType.php
```diff
@@ -0,0 +1,76 @@
...
+   */
+  public function getSourceId(): string {
+    return 'group';
+  }
+
+  /**
+   * @inheritDoc
+   */
+  public function getLabel(): string {
+    return t('Group', [], ['context' => 'eic_search']);
```
korhankale, line 23: When possible we try to use StringTranslationTrait, isn't it better if you add it and use $this->t() ? We'll also be more consistent with the rest of the codebase

lib/modules/eic_search/src/Search/Sources/UserSourceType.php
```diff
@@ -0,0 +1,65 @@
...
+   */
+  public function getSourceId(): string {
+    return 'user';
+  }
+
+  /**
+   * @inheritDoc
+   */
+  public function getLabel(): string {
+    return t('User', [], ['context' => 'eic_search']);
```
korhankale, line 23: Same comment for StringTranslationTrait here and everywhere else actually

config/sync/locale.settings.yml
```diff
@@ -0,0 +1,15 @@
+cache_strings: true
```
lramarojaona, line 1: Do we still need locale module to be enbaled?

config/sync/search_api.index.global.yml
```diff
@@ -198,12 +209,6 @@ field_settings:
     dependencies:
       module:
         - node
-  flag_recommend_group_count:
```
lramarojaona, line 201: We should keep this

lramarojaona (changes requested):

config/sync/search_api.index.global.yml
```diff
@@ -293,6 +396,12 @@ field_settings:
     dependencies:
       module:
         - user
+  url:
```
lramarojaona, line 399: @Dromadoros , can you explain which kind of URL this is?

Dromadoros: relative url of the entity for eg : 

image

lramarojaona: Ok, just renamed the label for clarity

lramarojaona (approved):

PR #275, opened 2021-06-07
EICNET-848: html tags shown in teaser components as plain text

4hardfire:
### Changes

Fix description text in group header and discussion teaser components.

### Tests

- [ ] Create a group and check if the body field is now shown as full html when viewing the group header;
- [ ] Create a group discussion and check if the body text tags have been stripped when viewing group discussions overview

### DIFF
--- a/lib/themes/eic_community/includes/preprocess/content/node--discussion.inc
+++ b/lib/themes/eic_community/includes/preprocess/content/node--discussion.inc
@@ -7,6 +7,7 @@
 
 use Drupal\taxonomy\Entity\Term;
 use Drupal\comment\Entity\Comment;
+use Drupal\Component\Utility\Xss;
 
 /**
  * Implements hook_preprocess_node() for discussion node.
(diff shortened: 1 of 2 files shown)


lramarojaona (approved):

lramarojaona:
Looks good to me

lib/themes/eic_community/includes/preprocess/content/node--discussion.inc
```diff
@@ -45,7 +46,7 @@ function eic_community_preprocess_node__discussion(array &$variables) {
         ],
       ],
       'author' => eic_community_get_teaser_user_display($node->getOwner()),
-      'description' => $node->get('field_body')->first()->getValue()['value'],
+      'description' => Xss::filter($node->get('field_body')->first()->getValue()['value']),
```
korhankale, line 49: what about using the same way of accessing value as above? Will spare you some characters actually

4hardfire: @korhankale like this you mean? `Xss::filter($node->get('field_body')->value)`

korhankale: @4hardfire Yes I would stay the important thing is that we keep being consistent within the project, however since I never understood in what case it's better to use this getter I find better to use this `Xss::filter($node->get('field_body')->value)`

4hardfire: @korhankale indeed. I usually do it that way but in this case I wasn't the one who created this preprocesss function and that's why I kept as it is. But I will change it since we both agree on that. Thanks!

4hardfire: Done.

PR #278, opened 2021-06-07
Feature/EICNET-832: Admin overview + CTA for groups

korhankale:


config/sync/field.field.flagging.request_delete_group.field_deletion_status.yml
```diff
@@ -0,0 +1,23 @@
...
+description: ''
+required: false
+translatable: true
+default_value:
+  -
+    value: 0
+default_value_callback: ''
+settings:
+  on_label: 'On'
+  off_label: 'Off'
```
lramarojaona, line 22: @korhankale ,
Would it be more future-proof to have this field as a list options so we can have different statuses (requested/denied/deleted/archived)

lib/modules/eic_flags/src/DeleteRequestListBuilder.php
```diff
@@ -0,0 +1,178 @@
...
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Routing\RedirectDestinationInterface;
+use Drupal\eic_flags\Service\DeleteRequestManager;
+use Drupal\flag\FlaggingInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Defines a class to build a listing of flags with type 'delete_request_*'.
+ */
+class DeleteRequestListBuilder extends EntityListBuilder {
```
lramarojaona, line 20: @korhankale ,
I think it would be much more extendable to create a view for this, instead of a hard-coded list builder, do you agree?
As needs might evolve in the future, and also we should think of re-usability in sibling sites (like C4D)

lramarojaona (commented):
@korhankale ,
As SA, I can request deletion for a group, but then when I navigate back to the group and request deletion again I'm getting those messages after filling the form:
- You are not allowed to do this
- The request has been made 
I think the request deletion feature should not be available anymore until some action has been made on the previous request

lib/modules/eic_flags/src/Routing/RequestDeleteRoutes.php
```diff
@@ -33,6 +33,16 @@ public function routes() {
       $route_collection->add('entity.' . $entity_type . '.request_delete_form', $route);
     }
 
+    // Define the route displaying entities flagged for removal
+    $route = (new Route('/admin/content/request/delete'))
```
lramarojaona, line 37: Also, we might gather all EIC admin related pages under one specific menu item, as there will be more pages like this, but this can be a more global thought

config/sync/field.field.flagging.request_delete_group.field_deletion_status.yml
korhankale: Indeed, was planning to rework this field with another branch which will allow to have multiple flags. As out of the box the flag module doesn't allow multiple flags for the entity_type + id + account/session combination. So we must take care of this ourselfs

lib/modules/eic_flags/src/DeleteRequestListBuilder.php
korhankale: For the views part I don't agree at all, in my opinion that's how we end up having pages which doesn't follow designs, are hard to evolve (especially for frontends) without writing hundreds of lines of codes to alter/extend what views is intended for. <br> We can't really say this is "hard coded" since it is an instance of EntityListBuilder which is implemented for every Drupal entity. For the reusability nothing prevents someone extending this class actually

config/sync/field.field.flagging.request_delete_group.field_deletion_status.yml
korhankale: Gonna remove it from this PR and implement it in another one :)

lib/modules/eic_flags/src/Routing/RequestDeleteRoutes.php
korhankale: What path should we follow? /admin/eic/* ?

lramarojaona: Let's use a generic one: `admin/community/*`

lramarojaona (approved):
Ok, it looks good for now, I will create a follow-up ticket to see if we can replace the list with a view

PR #279, opened 2021-06-07
EICNET-850: Public group content should be available to trusted users and anonymous user

4hardfire:
### Changes

- Create class that overrides public group visibility plugin in order to allow anonymous and trusted users to view group content of a public group.

### Tests

- [ ] Create a discussion
- [ ] As anonymous user when I go to the group page I should be able to view the group header and the discussion.


lramarojaona (approved):
It works fine, but we must keep in mind that we'll need similar stuff for Search API based views.
This will be treated in antother ticket.

PR #280, opened 2021-06-08
Feature/eicnet 770 group discussions storybook to drupal

xwero:


lib/themes/eic_community/templates/facets/facets-item-list--checkbox--discussions-regions-countries.html.twig
```diff
@@ -0,0 +1,7 @@
+{#
+/**
+ * @file
+ * Theme override for the taxonomy facet.
+ */
+#}
+{% include '@eic_community/facets/facets-item-list--checkbox--groups.html.twig' %}
```
4hardfire, line 7: Hi @xwero, couldn't we simply override the template **facets-item-list--checkbox.html.twig** and copy paste the code that is in **facets-item-list--checkbox--groups.html.twig**? I think there is no need to create multiple templates with the same code when you can use a suggestion already provided by the theme.

4hardfire (changes requested):

lib/themes/eic_community/templates/facets/facets-item-list--checkbox--discussions-regions-countries.html.twig
xwero: This pull request should be removed, because most of the things are going to be moved to react.

xwero:
Changes made here will be overwitten when the group discussions overview will have a react.js part.

PR #281, opened 2021-06-08
Feature/eicnet 787 implements global search

toolbarthomas:
This PR implements the global search example for Storybook and introduces the `Searchform` and `Tooltip` components.

The tooltip component is visible within `Storybook > Components > Tooltip` and is also used within the actual Searchform component: `Storybook > Compositions > Searchform`.

The actual example of the overview is visible at `Storybook > Pages > Search`.

The actual results are dynamic, meaning that the content and icons are defined from the defined data. Some icons currently don't look the same since they need to be converted by the designer (but this can be fixed in future pr's).


4hardfire (approved):
Seems ok to me.

PR #282, opened 2021-06-08
EICNET-829: Book page sidebar navigation and add book page links (current level / child)

didierdemaeyer:
* Add book navigation in sidebar for books that do not belong to a group
* Add links to add book page on current level or child page
* Fix circular redirect issue for book pages outside groups (BookPageRedirectSubscriber)
* Hide author and publish date information from wiki pages

### Test

* Create a new book page not linked to a group - /node/add/book
* SA/SCM can see and use add book page links ("Add a new book page" on top level book page) ("Add a new page on the current level" and "Add a new page below this page" on other book pages)
* All site users can see the book navigation in the sidebar for all book pages
* Book and wiki pages that belong to a group are not affected by these changes

### DIFF
--- a/lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
+++ b/lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
@@ -92,7 +92,7 @@ class EntityOperations implements ContainerInjectionInterface {
             return;
           }
 
-          if (isset($entity->book) && $entity->book['pid'] === '0') {
+          if (isset($entity->book) && empty($entity->book['pid'])) {
             // If top level book page.
             $build['link_add_child_book_page'] = $add_book_page_urls['add_child_book_page']->toString();
             $build['link_add_child_book_page_renderable'] = Link::fromTextAndUrl($this->t('Add a new book page'), $add_book_page_urls['add_child_book_page'])->toRenderable();
--- a/lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
+++ b/lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
@@ -5,13 +5,12 @@ namespace Drupal\eic_content_book\Hooks;
 use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
 use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
 use Drupal\Core\Entity\EntityInterface;
-use Drupal\Core\Entity\EntityTypeManagerInterface;
-use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\Link;
 use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\Core\Session\AccountProxyInterface;
 use Drupal\Core\StringTranslation\StringTranslationTrait;
 use Drupal\Core\Url;
+use Drupal\eic_content\EICContentHelperInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
@@ -24,25 +23,11 @@ class EntityOperations implements ContainerInjectionInterface {
   use StringTranslationTrait;
 
   /**
-   * The entity type manager.
-   *
-   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
-   */
-  protected $entityTypeManager;
-
-  /**
    * The current route match service.
    *
    * @var \Drupal\Core\Routing\RouteMatchInterface
    */
   protected $routeMatch;
-
-  /**
-   * The module handler service.
-   *
-   * @var \Drupal\Core\Extension\ModuleHandlerInterface
-   */
-  protected $moduleHandler;
 
   /**
    * The current user.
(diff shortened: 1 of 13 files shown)


lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
```diff
@@ -0,0 +1,146 @@
...
+          unset($build['book_navigation']);
+
+          $add_book_page_urls = $this->getBookPageAddFormUrls($entity);
+
+          // If user can't access the route.
+          if (!$add_book_page_urls['add_child_book_page']->access($this->currentUser)) {
+            return;
+          }
+
+          if (isset($entity->book) && $entity->book['pid'] === '0') {
```
Dromadoros, line 95: the pid really returns a string and not an integer ?

lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
```diff
@@ -0,0 +1,146 @@
...
+
+  /**
+   * Implements hook_node_view().
+   */
+  public function nodeView(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display, $view_mode) {
+    switch ($this->routeMatch->getRouteName()) {
+      case 'entity.node.canonical':
+        if ($entity->bundle() === 'book') {
+          // Ignore book page that belongs to a group.
+          if ($this->eicContentHelper->getGroupContentByEntity($entity)) {
```
Dromadoros, line 81: I'm not sure but is this not handle by default with group module ? I'm surprised that a node linked to a group can be by default accessible by everyone

lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
```diff
@@ -0,0 +1,146 @@
...
+            return;
+          }
+          // Unsets book navigation since we already have that show in the
+          // eic_content_book_navigation block plugin.
+          unset($build['book_navigation']);
```
Dromadoros, line 86: Can we avoid this ? I find it strange to remove the navigation via this preprocess and not config

lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
4hardfire: Hi @Dromadoros, I've created this method in EICGroupsHelper service so that we could retrieve a group from any given entity. The method itself internally uses the **GroupContent::loadByEntity** to get the group content of a node and then the respective group. In any case I did this to avoid duplicated code.

4hardfire: But indeed... in this case we could simply check:
`if (GroupContent::loadByEntity($entity))`

lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
4hardfire: @Dromadoros, The book_navigation is not something you can control via configuration unfortunatelly. The book module actually adds it to the renderable array:
```php
/**
 * Implements hook_ENTITY_TYPE_view() for node entities.
 */
function book_node_view(array &$build, EntityInterface $node, EntityViewDisplayInterface $display, $view_mode) {
  if ($view_mode == 'full') {
    if (!empty($node->book['bid']) && empty($node->in_preview)) {
      $book_node = Node::load($node->book['bid']);
      if (!$book_node->access()) {
        return;
      }
      $build['book_navigation'] = [
        '#theme' => 'book_navigation',
        '#book_link' => $node->book,
        '#weight' => 100,
        // The book navigation is a listing of Node entities, so associate its
        // list cache tag for correct invalidation.
        '#cache' => [
          'tags' => $node->getEntityType()->getListCacheTags(),
        ],
      ];
    }
  }
}
```

lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
4hardfire, line 81: @didierdemaeyer can you check if it is possible to use the condition as follow? -> `if (GroupContent::loadByEntity($entity))`

4hardfire (changes requested):

lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
didierdemaeyer: @Dromadoros Yes it does. I thought it was weird too

lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
4hardfire: Changes not needed after all. All good!

lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
4hardfire: @Dromadoros I replaced with an empty function. It works fine.

lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
4hardfire: Changes not needed after all. All good!

4hardfire (approved):
It looks fine to me now.

PR #284, opened 2021-06-09
EICNET-857: Breadcrumb doesn't show the correct group title

4hardfire:
### Improvements

- Create breadcrumb builder for groups and nodes that belong to groups.

### Tests

- [ ] Create a group A
- [ ] Navigate to the group detail page and check if the breadcrumb shows the right group title and url
- [ ] Create a new group B
- [ ] Navigate to the group B detail page and check if the breadcrumb shows the right group title and url
- [ ] Navigate to the group B Wiki (you can access it via **/admin/content**) and check if the breadcrumb **Home > group-title > group-wiki-title**
- [ ] Create a new wiki page from the group B Wiki page
- [ ] Navigate to the wiki page previously created and check if the breadcrumb is **Home > group-title > group-wiki-title > wiki-page-title**

### DIFF
--- a/lib/modules/eic_groups/eic_groups.services.yml
+++ b/lib/modules/eic_groups/eic_groups.services.yml
@@ -12,3 +12,8 @@ services:
     arguments: ['@eic_groups.helper', '@book.manager']
     tags:
       - { name: event_subscriber }
+  eic_groups.breadcrumb:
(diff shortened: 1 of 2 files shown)


lramarojaona (approved):
Although we missing breadcrumbs block on some pages, this seems to work correctly.

PR #285, opened 2021-06-09
Feature/EICNET-831: Notification MT38 + Flagging entity tokens

korhankale:


lramarojaona (approved):
Looks good!

docker-compose.yml
```diff
@@ -33,8 +34,8 @@ services:
     command: --innodb-log-file-size=1G --max_allowed_packet=1G --innodb-buffer-pool-size=512M --wait_timeout=3000 --net_write_timeout=3000 --log_error_verbosity=3
     environment:
       MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
-#    ports:
-#      - 3306:3306
```
lramarojaona, line 37: @korhankale , do we need those docker-compose changes?

lramarojaona (changes requested):

docker-compose.yml
korhankale: Actually yes,
Don't know if there are people using docker but without the first change there are no chances xdebug 3 will work and the second one is to expose the mysql port to allow any database GUI connecting to it

PR #288, opened 2021-06-10
Feature/eicnet 826 group about page

lramarojaona:


4hardfire (changes requested):
@lramarojaona, because you are injecting services from **oec_group_flex** we need to add the module dependency in **eic_groups.info.yml** otherwise the site install process will fail.

lib/modules/eic_groups/src/EICGroupsHelper.php
```diff
@@ -130,4 +175,104 @@ public function getGroupOperationLinks(GroupInterface $group, $limit_entities =
...
+    $joining_methods = $this->groupFlexGroup->getDefaultJoiningMethods($group);
+    $settings = [];
+    foreach ($joining_methods as $joining_method) {
+      if (isset($joining_method_plugins[$joining_method])) {
+        $settings[] = $joining_method_plugins[$joining_method]->getLabel();
+      }
+    }
+    return $settings;
+  }
+
```
4hardfire, line 277: @lramarojaona, what do you think if we create a new helper service for the oec_group_flex module and move these methods?

4hardfire (approved):
Seems fine to me now. @lramarojaona could you just do a final check before merging?

lramarojaona:
@4hardfire , as discussed I removed the interface for OECGroupFlexHelper class.

PR #289, opened 2021-06-11
EICNET-864: Implements title overflow for teaser.

toolbarthomas:
This PR adjusts the existing teaser component that will truncate the teaser title within a column overview items. 
The optional `title_before` and `title_after` arguments can be used in order to show extra text within the truncated titles.

An example has been implemented within `Storybook > Bundles > Group > Overview Private`


4hardfire (commented):
@toolbarthomas it works well when viewing the example you provided. However, I don't know if I am missing something but I tried to test with other content like organisations or members and I couldn't see the text being truncated for those cases. Can you just let me know if that was supposed to happen?
Thanks!

toolbarthomas:
The logic will be applied for teasers that exists within the container that has the `.ecl-teaser-overview--has-columns` class. This classname is currently used within the Group overview templates, we need to append to other component classes in order to get the text truncated within other overview.

4hardfire (approved):

PR #292, opened 2021-06-11
Feature/eicnet 862 replace deletion requests list builder with views

lramarojaona:
- Replaces the /admin/community/request/delete listbuilder with a view
- Creates a menu entry in admin menu


Dromadoros:
Do we need to keep this PR ?

PR #295, opened 2021-06-14
Feature/EICNET-833: Deny admin flow

korhankale:
This PR implements the deny reponse to delete requests but also introduces some revamps to make things more generic for other request types


lib/modules/eic_flags/eic_flags.module
```diff
@@ -73,34 +76,59 @@ function eic_flags_field_request_status_allowed_values() {
  * Implements hook_entity_type_alter.
```
lramarojaona, line 76: The hook is no correct, also should be followed by parenthesis:
_Implements hook_entity_type_build()._

lib/modules/eic_flags/eic_flags.module
```diff
@@ -73,34 +76,59 @@ function eic_flags_field_request_status_allowed_values() {
...
     if (!isset($entity_types[$entity_id])) {
       continue;
     }
 
     $entity_types[$entity_id]
       ->setFormClass('request-delete', RequestDeleteForm::class)
-      ->setLinkTemplate('request-delete-form', '/{entity_type}/{entity_id}/request-delete');
+      ->setFormClass('close-request', RequestCloseForm::class)
+      ->setLinkTemplate('request-delete-form', '/{entity_type}/{entity_id}/request-delete')
+      ->setLinkTemplate('close-request', '/admin/community/{entity_type}/{entity_id}/request/{request_type}/close');
```
lramarojaona, line 92: @korhankale 
I wonder if we should have /admin/community prefix set somewhere as a constant or config.
What do you think?

lib/modules/eic_flags/eic_flags.module
```diff
@@ -73,34 +76,59 @@ function eic_flags_field_request_status_allowed_values() {
...
  * Implements hook_entity_operation.
  */
 function eic_flags_entity_operation(EntityInterface $entity) {
-  $class_name = strtolower((new \ReflectionClass($entity))->getShortName());
+  /** @var \Drupal\eic_flags\Service\HandlerInterface $delete_request_handler */
+  $delete_request_handler = \Drupal::service('eic_flags.handler_collector')
+    ->getHandlerByType(RequestTypes::DELETE);
+
+  $entity_type_id = $entity->getEntityTypeId();
   if (!$entity instanceof ContentEntityInterface
```
lramarojaona, line 105: Can we split this in multiple if conditions?
For better readability and make debigging easier
https://www.drupal.org/docs/develop/standards/coding-standards#linelength

lib/modules/eic_flags/eic_flags.module
```diff
@@ -110,13 +138,24 @@ function eic_flags_entity_operation(EntityInterface $entity) {
...
+  if (!in_array(
+    $entity_type_id,
+    array_keys($delete_request_handler->getSupportedEntityTypes()))
+  ) {
     return AccessResult::neutral();
   }
 
-  // We allow delete requests only for certain roles.
-  if (empty(array_intersect(['administrator', 'trusted_user', 'content_administrator'], $account->getRoles()))) {
+  // We allow requests only for certain roles.
```
lramarojaona, line 153: I think we need to implement a new permission for this and use permission for the condition.
In that way we can easily change who can make delete requests

lramarojaona (changes requested):

lib/modules/eic_flags/eic_flags.module
korhankale: Actually reworked this with branch feature/EICNET-834 let's check it there

lib/modules/eic_flags/eic_flags.module
korhankale: Ok with it but maybe we should handle this in another PR since the routing will get more work (we want routes to be dynamically generated based on services having the 'request_handler' tag. This will make adding new request types easier)

lramarojaona: Ok, nice!

PR #298, opened 2021-06-15
Feature/EICNET-834: Implement the accept request response

korhankale:


lib/modules/eic_flags/src/Service/DeleteRequestHandler.php
```diff
@@ -38,4 +66,124 @@ public function getSupportedEntityTypes() {
     ];
   }
 
+  /**
+   * @param \Drupal\group\Entity\GroupInterface $group
+   */
+  private function deleteGroup(GroupInterface $group) {
+    $related_stories = $group->get('field_related_news_stories')
```
lramarojaona, line 73: We should not delete related stories and news, as they only exist on a global level, they never belong to a group.

lib/modules/eic_flags/src/Service/DeleteRequestHandler.php
```diff
@@ -38,4 +66,124 @@ public function getSupportedEntityTypes() {
...
+    // Retrieve group content entities linked to the group
+    $group_contents = $group->getContent();
+    $batch_builder = (new BatchBuilder())
+      ->setFinishCallback(
+        [
+          DeleteRequestHandler::class,
+          'deleteGroupContentFinished',
+        ]
+      )
+      ->setTitle($this->t('User updates'));
```
lramarojaona, line 88: User updates seem a bit confusing, can we have a more appropriate title here?
Such as "Deleting group content"?

lramarojaona (changes requested):

lib/modules/eic_flags/src/Service/DeleteRequestHandler.php
```diff
@@ -38,4 +66,118 @@ public function getSupportedEntityTypes() {
...
+        return;
+      }
+
+      if ($target_entity instanceof ContentEntityInterface) {
+        $target_entity->delete();
+      }
+
+    } catch (\Exception $exception) {
+      $context['results']['errors'][] = t(
+        'Something went wrong during content removal @error',
```
lramarojaona, line 132: Can we use $this->t() instead?

korhankale: Actually it is a static function so the answer is no without any negociation possible 😅

lramarojaona (approved):

PR #300, opened 2021-06-16
Feature/EICNET-837: Archive response for delete requests

korhankale:


config/sync/workflows.workflow.groups.yml
```diff
@@ -40,6 +40,12 @@ type_settings:
       from:
         - draft
         - published
+    unpublish:
+      label: Unpublish
+      from:
+        - published
+      to: pending
```
lramarojaona, line 47: @korhankale ,
We should use a new "Archived" status instead

lib/modules/eic_flags/src/Form/RequestDeleteForm.php
```diff
@@ -67,7 +69,9 @@ public function getFormId() {
    * {@inheritdoc}
    */
   public function getDescription() {
-    return $this->t('You\'re about to request a deletion for this entity. Are you sure?');
+    return $this->t(
+      'You\'re about to request a deletion for this entity. Are you sure?'
```
lramarojaona, line 73: Would be better to use double quotes here to avoid having escape characters in the string:
`"You're about to request a deletion for this entity. Are you sure?"`

lramarojaona (changes requested):

lramarojaona:
@korhankale ,
When I select "Archive" on a delete request, I'm seeing the following message on the confirm form:
`Are you sure you want to apply response "denied" to the group Group A?`

Where does the user message go when archiving the entity?

When I request a deletion, I'm getting this message, even thought the deletion request is correctly created:
`You are not allowed to do this`

lib/modules/eic_flags/src/Form/RequestDeleteForm.php
korhankale: Personnaly I prefer to use single quotes everywhere (whether or not escape is required) for multiple reasons: consistency, they are not parsed and are the go to quotes. Also it seems on the project this is the type we use

lramarojaona: I understand, but in this project we use single quotes, unless we have single quotes in the sting itself.
See https://www.drupal.org/docs/develop/standards/coding-standards#quotes

korhankale:
> @korhankale ,
> When I select "Archive" on a delete request, I'm seeing the following message on the confirm form:
> `Are you sure you want to apply response "denied" to the group Group A?`
> 
> Where does the user message go when archiving the entity?
> 
> When I request a deletion, I'm getting this message, even thought the deletion request is correctly created:
> `You are not allowed to do this`

The "You are not allowed to do this" message is fixed, I also changed the message for archive responses

config/sync/workflows.workflow.groups.yml
```diff
@@ -40,8 +40,8 @@ type_settings:
       from:
         - draft
         - published
-    unpublish:
```
lramarojaona: @korhankale ,
Can you also create the workflow state?
It doesn't exist yet

lramarojaona (changes requested):

config/sync/workflows.workflow.groups.yml
lramarojaona: I mean the archived

korhankale: As discussed, will be handled in another PR. Changing the state to a new one (archived) makes the edit page inaccessible (should be a permission issue or something like this)

lramarojaona (approved):

PR #302, opened 2021-06-16
EICNET-870: Implement solution for private / public stories & news articles

4hardfire:
### Description

- Created new custom module eic_private_content that exposes a new property "private" on Nodes;
- Enabled private content for CTypes News and Story.

### Todo

- Create custom view access check to allow/restrict access based on private field.

### Tests

- [ ] Enable permission to access admin content overview page for Trusted users;
- [ ] Disable view private content permission for Trusted users;
- [ ] As a content admin, create a new story and mark it as private;
- [ ] As a trusted user navigate to admin content overview page and check if the node is **not** shown in the results;
- [ ] As a content admin, uncheck the private field so the node is now public
- [ ] As a trusted user navigate to admin content overview page and check if the node is shown in the results;
- [ ] Enable permission to view private content for Trusted users;
- [ ] As a content admin, umark the story as private again;
- [ ] As a trusted user navigate to admin content overview page and check if the node is shown in the results;

Homepage:

- [ ] As a content admin, add some private and public stories/news to the homepage latest news and stories block;
- [ ] As anonymous, check if I can only see public news/stories


lib/modules/eic_private_content/eic_private_content.module
```diff
@@ -0,0 +1,131 @@
...
+ * Entity builder for the node type form with private option.
+ *
+ * @see eic_private_content_form_node_type_form_alter()
+ */
+function eic_private_content_form_node_type_form_builder($entity_type, NodeTypeInterface $type, &$form, FormStateInterface $form_state) {
+  $existing = $type->getThirdPartySetting('eic_private_content', PrivateContentConst::FIELD_NAME);
+  $new = $form_state->getValue('private');
+  $type->setThirdPartySetting('eic_private_content', PrivateContentConst::FIELD_NAME, $new);
+  if ($new != $existing) {
+    // @todo Mark all nodes as non private.
```
didierdemaeyer, line 106: @4hardfire Should this still be done in this PR?

didierdemaeyer (changes requested):
Looks good, but there's still one todo in the code.

lib/modules/eic_private_content/eic_private_content.module
4hardfire: @didierdemaeyer In my opinion this should be done with cron in a follow-up ticket. Do you agree?

didierdemaeyer: @4hardfire Yes, that's ok.

didierdemaeyer (approved):

PR #303, opened 2021-06-17
[EICNET] fix: Do not give an opinion for unsupported operations

korhankale:


4hardfire (approved):
Seems to fix the issue! Thanks!

PR #306, opened 2021-06-17
Fixes some issues when deleting comments

korhankale:
### Fixes
Error on approuval of a comment delete request
Count of open requests
Hook being not triggerred when request is created

### How to test

- [x] Introduce 2 delete request for content entities (e.g: a comment)
- [x] An email should be sent and available in mailhog when the queue runs
- [x] Go to /admin/community/request/delete, count of requests should match to the number of open requests
- [x] Approuve the delete request for an entity, the entity should be deleted


4hardfire (approved):
Tests:

- Approve deletion of entities with multiple requests
- Deny works
- Archive works

PR #307, opened 2021-06-18
Feature/869 add storybook styling to discussion detail page

xwero:
The contributors are now next to the comments. 

The contributor grid is still not correct.


didierdemaeyer:
@xwero Seems like the titles "Comments" and "Contributors" are not correctly aligned. 

Seems like the Contributors title uses a h2 which has margin and the comments uses a span with a class specific for comments.

Screenshot 2021-06-18 at 14 12 20

didierdemaeyer (changes requested):
This seems like the correct styling for the comments + contributors in aside. Perhaps you can use this

https://eic-d8.stg.blue4you.be/themes/custom/eic_community/storybook-static/index.html?path=/story/compositions-comments--overview-private

xwero:
The pull request is updated.

didierdemaeyer (approved):

PR #314, opened 2021-06-21
EICNET-806: Fixed frontend forms

xwero:
This change will fix the form field that use a table theme, like topics and countries & regions.
It wil also fix submitting the forms. 

I tested the submit change as a trusted user with the add group form, /group/add/group. And adding a discussion to the group.
I tested the topic fields on the two previous forms and also on the member tab of the user.

The field layout of the forms is still messy.


4hardfire (approved):
Everything seems to be working like a charm. No issues with autocomplete fields, paragraphs, submit buttons. Good job!

PR #315, opened 2021-06-21
Feature/EICNET-835: Notification MT39 - Delete request denied

korhankale:
Sends a notification when a delete request is denied

### To Test

- [ ] Make one or more delete requests on an entity
- [ ] Go to /admin/community/request/delete and deny it with a SA/SCM
- [ ] A message should be sent for each existing requests to the requester and author


4hardfire (approved):
Seems ok to me!

PR #316, opened 2021-06-21
Feature/EICNET-836: Notification MT40 - Delete request accepted

korhankale:
Sends a notification when a delete request is accepted

### To Test

- [ ] Make one or more delete requests on an entity
- [ ] Go to /admin/community/request/delete and accept it with a SA/SCM
- [ ] A message should be sent for each existing requests to the requester and author


4hardfire (approved):
Seems ok to me!

PR #317, opened 2021-06-22
EICNET-983: Problem with author picture fixed

xwero:
The author picture is now visible again, the css bug affected the display everywhere.

It was noticed on the discussion detail page contributor grid. But I also saw it in the discussion teaser on the group page and in the comments on the discussion detail page.


4hardfire (approved):
Seems ok to me!

PR #319, opened 2021-06-23
EICNET-958/956: Align prefix in URL with breadcrumb (singular / plural naming)

4hardfire:
### Changes

- Update group pathauto pattern so that it uses the plural form;
- Replace breadcrumb slug of group book page with "Wiki"

### Tests

- [ ] Create a group
- [ ] Go to the group detail page and check if the URL is `groups/<group-title>`
- [ ] Navigate to the group book page and check if the URL is `groups/<group-title>/wiki` and the breadcrumb is `Home > Groups > Group title > Wiki`


lib/themes/eic_community/includes/preprocess/navigation/breadcrumb.inc
```diff
@@ -5,14 +5,32 @@
...
 /**
  * Implements hook_preprocess_breadcrumb().
  */
 function eic_community_preprocess_breadcrumb(array &$variables): void {
   $request = \Drupal::request();
   $route_match = \Drupal::routeMatch();
-  $title = \Drupal::service('title_resolver')
-    ->getTitle($request, $route_match->getRouteObject());
+
+  switch ($route_match->getRouteName()) {
```
Dromadoros, line 17:
```suggestion
  if ('entity.node.canonical' === $route_match->getRouteName() && $node = $route_match->getParameter('node') && GroupContent::loadByEntity($node) && $node->bundle() == 'book') {
```

Isn't something that can works in one line ?

4hardfire: > Isn't something that can works in one line ?

@Dromadoros I am doing this way to leave it open for other possible routes. And in fact there is an open PR with another exception for route **eic_groups.about_page** (see https://github.com/EIC-EA/EIC-community-D8/pull/331/files#diff-6a3dcaaa7fe058dbc48841c381f8bce67855543a1103fad95c22760ad32390c5). Also... if we structure the code this way I guess it makes it easier to read right?

Dromadoros (approved):

PR #321, opened 2021-06-23
EICNET-871: Implement solution to select "related stories and news" based on permissions

4hardfire:
### New implementations

- Create a new entity reference view to show all News/Stories in related stories field from group. Also, show the privacy label in front of each result from the autocomplete;

### Tests

- [ ] Create a couple of News and Stories. Create at least 1 public and 1 private.
- [ ] Go to the group add form and try to select some news and stories. Check if there is a label **(Private)** or **(Public)** depending on the privacy set for the News/Story.


4hardfire (commented):
PR is ready to review.

Dromadoros (approved):

PR #324, opened 2021-06-23
Feature/eicnet 840 admin logs list

lramarojaona:
Test:
- Create log messages (currently there's only one of that type and it's not triggered, I had to create one programmatically :/ )
- Go to /admin/community
- Select Logs
- Check the results

Notes:
- Now all message templates have one of those types: stream/log/notification/subscription
- The message type can be set to multiple in the future if needed


4hardfire (approved):
Seems ok to me!

PR #326, opened 2021-06-24
EICNET-986: Started with the redesign of the group header.

xwero:
This includes subtickets 996, 997, 998, 999, 1000, 1001.

Not all the scenarios are done because it needs backend work. This pull request is made to test the backend work.


4hardfire:
**EICNET-996**:

- Flags: there is an issue with the flag token. Apparently when I try to like I get an access denied. Not really sure why that is happening but with lazy_builder it works properly.

4hardfire:
**EICNET-997**:

- @xwero, the button to "Request group membership" is shown in the place as flags + statistics. Shouldn't this be replacing the button "Join group"? (see attached images)

image
image

4hardfire:
**EICNET-998**:

- Link to "Invite user" is not yet available for GM, GO and GA because there will be limitations depending on the group visibility and this was not yet developed;

lib/themes/eic_community/includes/preprocess/blocks/block--eic-group-header-block.inc
```diff
@@ -41,21 +40,102 @@ function eic_community_preprocess_eic_group_header_block(array &$variables) {
...
+
+      $actions[]['link'] = [
+        'label' => $action['title'],
+        'path' => $action['url'],
+      ];
+    }
+  }
+
+  $variables['group_values']['actions'] = $actions;
+  $flags = eic_community_get_flag_links_display($variables['group']->flags);
```
4hardfire, line 56: I am having an error when navigating to the group wiki section. I guess we are missing a condition when flags are not in the variables:

```php
$flags = [];

if (!empty($variables['group']->flags)){
  $flags = eic_community_get_flag_links_display($variables['group']->flags);
}
```

4hardfire (changes requested):

lib/themes/eic_community/includes/preprocess/blocks/block--eic-group-header-block.inc
4hardfire: Done.

4hardfire:
**EICNET-999**:

- Link to "Invite user" is not yet available for GM, GO, GA and SCM because there will be limitations depending on the group visibility and this was not yet developed;
- Missing link to add a new Wiki page;

4hardfire:
**EICNET-1000**:

- "Post content" button is still visible. Question: shouldn't the user be able to add content while the group is still in pending state?
- Button "Delete this group" is not shown;
- Cog wheel button is not shown;
- Flags are still visible;

4hardfire:
**EICNET-1001**:

- "Publish" button is missing (needs back-end);
- Cog wheel button is missing (needs back-end);
- Flags are still visible (needs back-end);

4hardfire (approved):
For now I will approve this since we already have most of the components shown in the group header. New follow-up tickets should be created according to the notes I wrote in this PR.

PR #327, opened 2021-06-24
Feature/EICNET-1017: Admin list improvements

korhankale:


lib/modules/eic_flags/src/FlaggedEntitiesListBuilder.php
```diff
@@ -105,9 +117,12 @@ public static function createInstance(
   public function buildHeader() {
     // Enable language column and filter if multiple languages are added.
     $header = [
-      'title' => $this->t('Title'),
+      'title' => [
+        'data' => $this->t('Title'),
+        'class' => [RESPONSIVE_PRIORITY_MEDIUM],
```
4hardfire, line 122: @korhankale where is this **RESPONSIVE_PRIORITY_MEDIUM** come from?

4hardfire: My IDE is complaining about it.

korhankale: Hmm it shouldn't, this is one of the constants drupal define within the global namespace. What IDE are you using?

4hardfire: I'm using VS code with WSL. This is fine... I just wanted to make sure this was not an issue :).

4hardfire:
@korhankale Are we taking into account pagination in the list of requests?

korhankale:
> @korhankale Are we taking into account pagination in the list of requests?

Currently we don't, let's not approve this PR. I'll add it this afternoon

4hardfire (changes requested):
- [ ] Add missing pager in the list

4hardfire (approved):
Seems ok to me.

PR #329, opened 2021-06-25
EICNET-970: Create custom descriptions for the visibility and joining…

lramarojaona:
… methods plugins on the About page.


4hardfire (commented):
@lramarojaona, do you think it would be better to manage the variables within a view mode? Let's say we create a new view mode "About page" in the group, and the build function of the AboutPageController would return the renderable array of the group in the view mode "About page". This way the logic for the front-ender would be in the preprocess_group instead of the preprocess_page which is more for global for all pages.

A little example of how to get a renderable array of an entity using Entity view builder:

```php
$view_builder = \Drupal::entityManager()->getViewBuilder('group');
$full_output = $view_builder->view($group);
$about_page_output = $view_builder->view($group, 'about_page');
```

lramarojaona:
@4hardfire,
I implemented your suggestion, it works well and is also cleaner. :)

Dromadoros (approved):

PR #332, opened 2021-06-28
Feature/EICNET-885: As a TU/SA/SCM I request archival of a certain entity type

korhankale:
Allows users to send archival requests
### To Test

- [x] Go to a group/comment/node page
- [x] Withing the operations links a new one "Request archival" must exists
- [x] After sending it, the request should be displayed here: /admin/community/request/archive
- [x] A new request notification should be sent


4hardfire (commented):
@korhankale, some remarks and questions:

- ~When viewing the list of archive requests, I see the operation links "Accept", "Deny", "Archive", "Edit". Do we actually need the "Archive" one? I suppose "Accept" or "Archive" is the same thing right?~ ("Archived" removed in #335)
- The ~confirmation label~ (fixed in #335) + page title don't seem correct for the "Accept" operation.
- The confirmation label doesn't seem correct for the "Deny" and "Archive" operations.
- The button to request archival is not visible in the group header although I can still access the form via URL.
- Page title in the list of archive request is wrong (`/admin/community/request/archive`)

lib/modules/eic_flags/src/Controller/FlagRequestController.php
```diff
@@ -6,6 +6,7 @@
 use Drupal\eic_flags\FlaggedEntitiesListBuilder;
 use Drupal\eic_flags\FlaggingListBuilder;
 use Drupal\eic_flags\RequestTypes;
+use http\Env\Request;
```
4hardfire, line 9: It doesn't seem to be used anywhere.

4hardfire (changes requested):

4hardfire (approved):
Seems ok to me. Thanks!

PR #339, opened 2021-06-30
EICNET-928: As a GO I want to edit my group after creating it (when it's not reviewed by a SA/SCM yet)

4hardfire:
### Improvements

- Implement hook_entity_field_access() to restrict access on group fields (**field_related_groups** and **field_related_news_stories**) when group moderation state is pending.

### Test

- [ ] As a trusted user try to create a new group `/group/add/group` and check if the fields **field_related_groups** and **field_related_news_stories** are not visible;
- [ ] Save the group
- [ ] Try to edit the group and make sure you can edit the fields **field_related_groups** and **field_related_news_stories**;

### DIFF
--- a/lib/modules/eic_groups/eic_groups.module
+++ b/lib/modules/eic_groups/eic_groups.module
@@ -5,7 +5,6 @@
  * Primary module hooks for EIC Groups module.
  */
 
-use Drupal\content_moderation\Entity\ContentModerationState;
 use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
 use Drupal\Core\Entity\EntityInterface;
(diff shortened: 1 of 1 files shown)


lib/modules/eic_groups/eic_groups.module
```diff
@@ -159,3 +164,46 @@ function eic_groups_cron() {
...
+function eic_groups_entity_field_access($operation, FieldDefinitionInterface $field_definition, AccountInterface $account, FieldItemListInterface $items = NULL) {
+  $access = AccessResult::neutral();
+
+  $group_restricted_fields = [
+    'field_related_groups',
+    'field_related_news_stories',
+  ];
+
+  // Access restrictions for group fields.
+  if (in_array($field_definition->getName(), $group_restricted_fields)) {
```
Dromadoros, line 180: It's just my opinion but for me it's easier to read it like this 

`if (!in_array($field_definition->getName(), $group_restricted_fields)) { return }`
So we avoid the "if" tree at maximum

Dromadoros (approved):

lib/modules/eic_groups/eic_groups.module
4hardfire: should be fine now :)

PR #340, opened 2021-06-30
EICNET-986: Closer to group header redesign

xwero:
Changes:

- The recommend group flag is working now
- The login button is visible when anonymus (EICNET-1039, EICNET-1047)
- Moved link manipulation to the `EICGroupHeaderBlock` class to make the `eic_community_preprocess_eic_group_header_block` cleaner

Problems:

- The group follow flag is displayed with the lib/themes/eic_community/patterns/components/flag.html.twig template file which is meant to display a country flag. That flag does not render with the `eic_flag_count` theme like the recommend group flag. 
- When a user with a site admin role is not part of a group both 'join group' and 'request membership group' buttons are shown. This could be confusing. 
- There are still open tickets that block some of the scenarios in the subtickets of EICNET-986


lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
```diff
@@ -138,13 +138,48 @@ public function build() {
...
+        unset($user_operation_links[$key]);
+
+        $operation_links[$key] = $action;
+      }
+    }
+
+    $create_operations = [];
+
+    foreach ($node_operation_links as $key => $link) {
+      if (strpos($key, 'create') !== false) {
```
4hardfire, line 166: @xwero how many times is this IF condition TRUE? If  it's something that happens once I guess we can do a `break;` at the end of the IF. Also, do you think it would be nice if we add a comment in the beginning of the foreach just to understand what's happening?

lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
```diff
@@ -138,13 +138,48 @@ public function build() {
...
+    $operation_links = [];
+    // Get login link for anonymous users.
+    if ($login_link = $this->getAnonymousLoginLink($group)) {
+      $operation_links['anonymus_user_link'] = $login_link;
+    }
+
+    foreach ($user_operation_links as $key => $action) {
+      if (in_array($action['url']->getRouteName(), ['entity.group.group_request_membership', 'entity.group.join'])) {
+        unset($user_operation_links[$key]);
+
```
4hardfire, line 158: Can we delete this line?

4hardfire (changes requested):

lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
```diff
@@ -138,13 +138,48 @@ public function build() {
...
-    if (isset($operation_links['gnode-create-wiki_page'])) {
-      unset($operation_links['gnode-create-wiki_page']);
+    if (isset($node_operation_links['gnode-create-wiki_page'])) {
+      unset($node_operation_links['gnode-create-wiki_page']);
+    }
+
+    $operation_links = [];
+    // Get login link for anonymous users.
+    if ($login_link = $this->getAnonymousLoginLink($group)) {
+      $operation_links['anonymus_user_link'] = $login_link;
```
4hardfire, line 152: Typo in `anonymus_user_link`. It should be `anonymous_user_link`.

4hardfire (changes requested):

4hardfire:
> Problems:
> 
> * The group follow flag is displayed with the lib/themes/eic_community/patterns/components/flag.html.twig template file which is meant to display a country flag. That flag does not render with the `eic_flag_count` theme like the recommend group flag.
> * When a user with a site admin role is not part of a group both 'join group' and 'request membership group' buttons are shown. This could be confusing.
> * There are still open tickets that block some of the scenarios in the subtickets of EICNET-986

- The flag group follow needs to be changed in the backend. We should be using the option **EIC Count link** and link type;
- I guess we will create a new role for site admins because right now we are assuming the Drupal administrator role and that role can bypass everything. This is the problem;

lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
xwero: All the create links that come from the `$this->eicGroupsHelper->getGroupOperationLinks($group, ['node'], $cacheable_metadata)` are matched, so at the moment it wil happen three times.

Before the rewrite I had to hope that the links in the `operations_links` array where the create links. Now we are sure they are and we can do al lot more link manipulation because the link handling in the preprocess function is now generic.

I added a comment in my last update.

lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
xwero: The unset function can not be removed, that stops the link for appearing a button and as a link in the grey section below the buttons.

lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
xwero: Fixed.

lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
4hardfire: I mean the empty line :)

4hardfire (approved):
It seems fine to me!

PR #344, opened 2021-06-30
EICNET-1021: Show user links in site header.

xwero:
This is a PR for EICNET-1020 and EICNET-1021.

Tests:

- When you are not logged in you should see a login link in the site header.

- When you are logged in you should the user image or image placeholder and name.
-- When you click on the user you should see the profile and log out link.

Improvement FE:

When the browser is slow you first see the links without the user trigger. This is the case for all the links in a dropdown.


lib/themes/eic_community/includes/preprocess/layout/page.inc
```diff
@@ -21,4 +21,42 @@ function eic_community_preprocess_page(array &$variables): void {
...
+      'alt' =>  t('European Commission logo'),
+      'path' => \Drupal::urlGenerator()->generateFromRoute('<front>'),
+      'src_desktop' => $variables['eic_logo_path'] . '/logo--' . $variables['language']->getId() . '.svg',
+    ]
+  ];
+
+  if ($variables['user']->id() === 0) {
+    $variables['site_header']['login']['link'] = [
+      'label' => t('Log in'),
+      'path' => '/user/login'
```
Dromadoros, line 38: Hey, it's better to use `Url::fromRoute('user.login');`, use id instead of direct path

lib/themes/eic_community/includes/preprocess/layout/page.inc
```diff
@@ -21,4 +21,42 @@ function eic_community_preprocess_page(array &$variables): void {
...
+    'icon_file_path' =>  $variables['eic_icon_path'],
+    'logo' => [
+      'title' =>  t('European Commission'),
+      'alt' =>  t('European Commission logo'),
+      'path' => \Drupal::urlGenerator()->generateFromRoute('<front>'),
+      'src_desktop' => $variables['eic_logo_path'] . '/logo--' . $variables['language']->getId() . '.svg',
+    ]
+  ];
+
+  if ($variables['user']->id() === 0) {
```
Dromadoros, line 35: It's working like this but I feel that a `\Drupal::currentUser()->isAnonymous()` is cleaner

Dromadoros: or in this case `$variables['user']->isAnonymous();`

lib/themes/eic_community/includes/preprocess/layout/page.inc
xwero: Changed.

lib/themes/eic_community/includes/preprocess/layout/page.inc
xwero: Changed.

lib/themes/eic_community/includes/preprocess/layout/page.inc
```diff
@@ -21,4 +22,42 @@ function eic_community_preprocess_page(array &$variables): void {
...
+      [
+        'link' => [
+          'label' => t('My profile'),
+          'path' => $account->toUrl()->toString(),
+        ],
+      ],
+      [
+        'link' => [
+          'label' => t('Log out'),
+          'path' => '/user/logout',
```
Dromadoros, line 56: Same for the logout, needs to generate from Route::

Dromadoros (approved):

PR #336, opened 2021-06-28
EICNET-1047: As a GA I can edit my group via a cog wheel in the group header

4hardfire:
### Improvements

- Rename method **getGroupContentOperationLinks()** from **EICGroupsHelper** service class;
- Add group edit link to the renderable array of group header block.

### Tests

- [ ] Create a group as a trusted user;
- [ ] From the group landing page, check if a new link to edit the group is available in the group header;

PR #347, opened 2021-07-02
EICNET-942: As a GO I can delete my group when it's in pending state

4hardfire:
### Changes

- New group permission helper methods added to eic_groups.helper service;
- Add/remove **group delete** permission from group owner when moderation state is **pending**.

### Tests

- [ ] Go to the group edit page and check if the delete link is available; Also, validate if the user has access to -> `groups/<group-title>/delete`
- [ ] As a SA/SCM update the group status to "Draft";
- [ ] As a trusted user, go to the group edit page and check if the delete link is not visible; Also, validate if the access is denied in -> `groups/<group-title>/delete`.

### Todo

- [ ] The delete button in the group header block is not yet available and needs to be added after we merge #336

### DIFF
--- a/lib/modules/eic_groups/src/Hooks/EntityOperations.php
+++ b/lib/modules/eic_groups/src/Hooks/EntityOperations.php
@@ -159,7 +159,7 @@ class EntityOperations implements ContainerInjectionInterface {
    */
   public function groupPermissionInsert(GroupPermissionInterface $group_permissions) {
     $group = $group_permissions->getGroup();
-    // Adds or removes delete group permission from group owner based on the
+    // Adds or removes "delete group" permission from group owner based on the
     // group moderation state.
     if ($group->get('moderation_state')->value === 'pending') {
       $this->eicGroupsHelper->addRolePermissionsToGroup($group_permissions, EICGroupsHelper::GROUP_OWNER_ROLE, ['delete group']);
@@ -368,7 +368,7 @@ class EntityOperations implements ContainerInjectionInterface {
   }
 
   /**
-   * Updates group owner permissions after based on moderation state.
+   * Updates group owner permissions based on moderation state.
    *
    * @param \Drupal\group\Entity\GroupInterface $entity
    *   The Group entity.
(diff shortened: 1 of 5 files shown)


lib/modules/eic_groups/src/Hooks/EntityOperations.php
```diff
@@ -103,6 +123,25 @@ public function groupUpdate(GroupInterface $entity) {
...
+  }
+
+  /**
+   * Implements hook_group_permission_insert().
+   */
+  public function groupPermissionInsert(GroupPermissionInterface $group_permissions) {
+    $group = $group_permissions->getGroup();
+    // Adds or removes delete group permission from group owner based on the
+    // group moderation state.
+    if ($group->get('moderation_state')->value === 'pending') {
```
lramarojaona, line 137: Maybe introduce constants for the moderation states?

lib/modules/eic_groups/src/Hooks/EntityOperations.php
```diff
@@ -263,4 +302,28 @@ protected function createGroupAboutPageMenuLink(EntityInterface $group) {
...
+   */
+  protected function updateGroupOwnerPermissions(GroupInterface $entity) {
+    /** @var \Drupal\group_permissions\Entity\GroupPermissionInterface $group_permissions */
+    $group_permissions = $this->groupPermissionsManager->loadByGroup($entity);
+
+    // Get moderation states.
+    $old_moderation_state = $entity->original->get('moderation_state')->value;
+    $new_moderation_state = $entity->get('moderation_state')->value;
+
+    if ($old_moderation_state !== $new_moderation_state && $new_moderation_state === 'pending') {
```
lramarojaona, line 319: Could we have a small comment to explain this condition?

lib/modules/eic_groups/src/Hooks/EntityOperations.php
```diff
@@ -263,4 +302,28 @@ protected function createGroupAboutPageMenuLink(EntityInterface $group) {
...
+      $this->eicGroupsHelper->addRolePermissionsToGroup($group_permissions, EICGroupsHelper::GROUP_OWNER_ROLE, ['delete group']);
+      $this->eicGroupsHelper->saveGroupPermissions($group_permissions);
+    }
+    elseif ($old_moderation_state !== $new_moderation_state && $new_moderation_state !== 'pending') {
```
lramarojaona, line 323: Could we have a small comment to explain this condition?

lramarojaona (changes requested):

lib/modules/eic_groups/src/Hooks/EntityOperations.php
4hardfire: Done. Can you have a look?

lib/modules/eic_groups/src/Hooks/EntityOperations.php
4hardfire: Done. Can you have a look?

lib/modules/eic_groups/src/Hooks/EntityOperations.php
4hardfire: Done.

PR #348, opened 2021-07-02
Feature/eicnet 1024 truncate group alias

lramarojaona:
- Create a group with a long title (e.g. "Group with with with with with with with with with long title, and some more lenghty description")
- Check that the group title token is truncated at 40 characters

### DIFF
--- a/lib/modules/eic_groups/src/Hooks/GroupTokens.php
+++ b/lib/modules/eic_groups/src/Hooks/GroupTokens.php
@@ -2,12 +2,12 @@
 
 namespace Drupal\eic_groups\Hooks;
 
-use Drupal\Component\Utility\Unicode;
 use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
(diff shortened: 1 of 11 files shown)


lib/modules/eic_groups/src/Hooks/GroupTokens.php
```diff
@@ -116,12 +123,21 @@ public function tokens($type, $tokens, array $data, array $options, BubbleableMe
...
-        if (isset($data['group']) && $data['group'] instanceof GroupInterface) {
-          // Provide replacements for truncated title tokens.
-          foreach ($this->tokenService->findWithPrefix($tokens, 'eic_groups_truncated_title') as $value => $token_name) {
-            $replacements[$original] = Unicode::truncate($data['group']->label(), $value);
-          }
+      }
+    }
+
+    // Group tokens.
+    if ($type == 'group' && !empty($data['group']) && $data['group'] instanceof GroupInterface) {
```
Dromadoros, line 130: better to be consistent i guess and use strict equal

lramarojaona: This is done

PR #349, opened 2021-07-02
EICNET-1061: Change the link type for the follow group flag.

lramarojaona:
- You should now see the follow group flag with a counter


4hardfire (approved):
Seems ok to me.

PR #350, opened 2021-07-02
EICNET-942: Add group "delete" operation to the group header block.

4hardfire:
### Changes

- Add group "delete" operation to the group header block.

### Tests

- [ ] Create a group
- [ ] Go to the group overview page and check if there is a button to delete the group in the group header block.


lramarojaona:
Looks good to me

lramarojaona (approved):

PR #351, opened 2021-07-02
EICNET-954: User profile - add validation to social channel links (Part 2)

4hardfire:
### Changes

- Apply patch to fix with limited values being randomised (see https://www.drupal.org/project/social_link_field/issues/3155628).
- Limit field to only 3 social network links (LinkedIn, Twitter and Facebook).

### Tests

- [ ] Try to edit the admin profile -> /profile/1/edit
- [ ] Check if linkedin, twitter and facebook are already set as default networks. Also it shouldn't be allowed to add more than 3 or change the the network types.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -5266,6 +5266,9 @@
                         "status": "not-covered",
                         "message": "Alpha releases are not covered by Drupal security advisories."
                     }
+                },
+                "patches_applied": {
+                    "#3155628 - Limited values are randomised": "https://www.drupal.org/files/issues/2020-09-02/3155628-10.patch"
                 }
             },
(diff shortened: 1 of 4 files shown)


lramarojaona (approved):
Looks good but the validation is quite cumbersome:
_Just add the part of your profile. Do not include preceding slash._

Let's create a followup ticket.

4hardfire:
> Looks good but the validation is quite cumbersome:
> _Just add the part of your profile. Do not include preceding slash._
> 
> Let's create a followup ticket.

I guess if you add `https://example.org` it should not pass. It's just a matter of improving the description.

PR #355, opened 2021-07-05
EICNET-1062: As a TU I can reply to a comment on discussion in a public group without being a GM

4hardfire:
### Changes

- Improve code to update group permissions in` \Drupal\eic_groups\Hook\EntityOperations` class.
- Deny access to reply to comments when user does not have permission to post comments in groups.

### Tests

- [ ] Create a new group discussion
- [ ] Add a comment
- [ ] Log in as a Trusted user, and navigate to the discussion page of the group
- [ ] Try to reply to the comment added previously. You should get an access denied.

### Todo

Hide Reply button when the user doesn't have access to reply. @xwero we need to add a condition in pattern `comment-base.html.twig` to check if the reply_path exists (**line 177**):
```twig
{% if reply_path %}
  <div class="ecl-comment__action">
    {% include "@ecl-twig/ec-component-link/ecl-link.html.twig" with {
      link: {
        label: reply_label|default('Reply'),
        path: reply_path,
      },
      extra_classes: 'ecl-comment__reply ecl-link--button ecl-link--button-ghost',
      extra_attributes: action_attributes,
    } %}
  </div>
{% endif %}
```

### DIFF
--- a/lib/modules/eic_groups/src/Hooks/EntityOperations.php
+++ b/lib/modules/eic_groups/src/Hooks/EntityOperations.php
@@ -151,8 +151,16 @@ class EntityOperations implements ContainerInjectionInterface {
     if (!$entity->original->isPublished() && $entity->isPublished()) {
       $this->publishGroupWiki($entity);
     }
-    // Updates group owner permissions.
-    $this->updateGroupOwnerPermissions($entity);
+
+    // Get moderation states.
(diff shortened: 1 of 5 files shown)


lramarojaona:
Looks good

lramarojaona (approved):

PR #356, opened 2021-07-05
Remove book.manager service from Drupal\eic_groups\Hooks\FormOperations::create() method.

4hardfire:
### Tests

- Try to access group creation form and check if there is not PHP error.


lramarojaona:
Looks good

PR #359, opened 2021-07-07
EICNET-904: Added storybook styling to group wiki pages.

xwero:
Testing:

* Check the wiki section on the group edit page
* Create wiki (sub)pages

Problems:

* The levels of subpages is set to nine, so the add page below button is visible on the third level
* At the moment I only added the wiki page creator. There are no additional contributors.
* The hide sidebar button isn't working as on storybook, but i couldn't figure out why.


4hardfire:
@xwero,

### Book page:

There is an issue when viewing the wiki homepage (CType book page) when there are no wiki pages. The width is kinda broken (see image).
image

### Wiki page

- Issue with the button to hide/show the sidebar. (TO BE FIXED)
- Wiki page navigation is working a bit strange. When I click on the arrow to open the child links it works ok but when I try to close it works as if I clicked on the link which redirects me to the top level instead of closing the child links. The arrow should always open/close the dropdown. (TO BE FIXED)
- Missing caption from the media (both images or remote videos). This needs to be discussed with @JuliePost as well.
- Remote video element looks very small (see image at the bottom)
- Missing elements from download medias: image + tag + author. Needs to be reviewed with @JuliePost
- PHP error when preprocessing media documents (uploaded video): `Error: Call to a member function get() on null in eic_community_preprocess_field__field_related_downloads() (line 26 of /var/www/html/lib/themes/eic_community/includes/preprocess/field/field--media--eic-document.inc)`

image

4hardfire (changes requested):
It needs some changes. Check previous comment.

PR #361, opened 2021-07-07
EICNET-1051: As a GO I shouldn't be able to add content when my group is in pending state

4hardfire:
### Improvements

- Override access_check.group_content.create_entity service class with a custom class GroupContentCreateEntityAccessCheck in order to have extra access conditions and prevent GO from creating content while the group is in pending state;
- Add missing conditions to check group operation links access before adding them to the group header;
-  Check access to delete group by implementing hook_group_access() instead of adding/removing permissions to the group owner.

### Tests

- [ ] From the group overview page, check if the **Post content** dropdown is not available;
- [ ] Try to manually access the group content creation form through the URL and check if you get an access denied;
- [ ] As a SA/SCM, approve the group (change to draft or published);
- [ ] As a GO, you should now be able to create group content and the "Post content" dropdown should now be available in the group header;

### Test regressions

- [ ] As a Trusted user and GO, make sure you can't delete your group when it's in draft or published state.

### Todo

~- Change the way we add/remove permissions to delete a group as a GO when group has been approved. Right now, we add/remove permission to delete based on the group moderation state but we should be doing an access check on top of the default permissions. It should be done after we merge #355~


lib/modules/eic_groups/eic_groups.module
```diff
@@ -215,9 +216,25 @@ function eic_groups_entity_field_access($operation, FieldDefinitionInterface $fi
...
+  $access = AccessResult::neutral();
+
+  $moderation_state = $entity->get('moderation_state')->value;
+
+  switch ($operation) {
+    case 'delete':
+      // Deny access to delete group if the group is NOT in pending state and
+      // the user is not an "administator" or a "content_administrator".
+      if ($moderation_state !== GroupsModerationHelper::GROUP_PENDING_STATE) {
+        $access = AccessResult::forbiddenIf(!in_array('administrator', $account->getRoles(TRUE)) && !in_array('content_administrator', $account->getRoles(TRUE)))
```
lramarojaona, line 231: Don't know if we have a permission that applies here to do the check, but we should at least use the role constants rather than hard-code them: https://github.com/EIC-EA/EIC-community-D8/blob/develop/lib/modules/eic_user/src/UserHelper.php

lib/modules/eic_groups/src/Access/GroupContentCreateEntityAccessCheck.php
```diff
@@ -0,0 +1,42 @@
...
+
+    // If access is allowed, we also need to check if the user can create group
+    // content based on the current group moderation state.
+    if ($access->isAllowed()) {
+      switch ($group->get('moderation_state')->value) {
+        case GroupsModerationHelper::GROUP_PENDING_STATE:
+          // Deny access to the group content node creation form if the group
+          // is in pending state and the user is not an "administator" or a
+          // "content_administrator".
+          $access = AccessResult::forbiddenIf(!in_array('administrator', $account->getRoles(TRUE)) && !in_array('content_administrator', $account->getRoles(TRUE)))
```
lramarojaona, line 31: Same here we should use role constants: https://github.com/EIC-EA/EIC-community-D8/blob/develop/lib/modules/eic_user/src/UserHelper.php

lramarojaona (changes requested):

lramarojaona:
Also, it seems I can delete the group as trusted user/GO while it's in pending state, but I guess that's the correct behavior?
It is however wrong if we need to comply to your statement: _"As a Trusted user and GO, make sure you can't delete your group when it's in pending state."_

lib/modules/eic_groups/src/Access/GroupContentCreateEntityAccessCheck.php
4hardfire: Done!

lib/modules/eic_groups/eic_groups.module
4hardfire: Done!

4hardfire:
> Also, it seems I can delete the group as trusted user/GO while it's in pending state, but I guess that's the correct behavior?
That is correct.

> It is however wrong if we need to comply to your statement: _"As a Trusted user and GO, make sure you can't delete your group when it's in pending state."_
The statement was wrong. The right one is: _As a Trusted user and GO, make sure you can't delete your group when it's in draft or published state._. I updated the PR description accordingly.

lramarojaona (approved):
Looks good now!

PR #366, opened 2021-07-09
EICNET-1059: update @ecl-twig/ec-components to version 2.35.2 to fix …

filipvt1:
Fix for double escaping bug in breadcrumb attributes


xwero (approved):
This fix works for the Drupal breadcrumb.

PR #367, opened 2021-07-09
EICNET-1072: Fix 'no wiki pages' notification styling

filipvt1:


xwero (approved):
The problem is fixed in stroybook.

But my pull request, https://github.com/EIC-EA/EIC-community-D8/pull/359, will need more work after the merge.

PR #368, opened 2021-07-09
EICNET-873: Added storybook styling to group about page

xwero:
Tests:

Change the group settings to test all the possible output.

Blocker:

The topics and region fields are not multi value. This means hierarchical lists have not been tested.


lib/themes/eic_community/includes/preprocess/groups/group.inc
```diff
@@ -0,0 +1,192 @@
...
+      'content' => [
+        '#theme' => 'eic_community_extended_list',
+        '#title' => t('Group Permissions'),
+        '#items' => $admin_items,
+      ]
+    ];
+  }
+  // output
+  $variables['sections'] = [
+    'as_tabs' => true,
```
Dromadoros, line 189: drupal coding standards always in CAPS boolean value and NULL

lib/themes/eic_community/includes/preprocess/groups/group.inc
```diff
@@ -0,0 +1,192 @@
...
+      '#theme' => 'eic_community_extended_list',
+      '#title' => t('Group details'),
+      '#items' => $about_items,
+    ]
+  ];
+  // permissions and visibility tab
+  if (isset($variables['elements']['visibility'])) {
+    $admin_items = [];
+
+    if ($variables['elements']['visibility']['plugin_id'] === 'custom_restricted') {
```
Dromadoros, line 133: Can we save this value into a variable so we dont need to repeat to whole array, I see that it's used 5 times

I mean save this one : $variables['elements']['visibility']

lib/themes/eic_community/includes/preprocess/groups/group.inc
```diff
@@ -0,0 +1,192 @@
...
+        ];
+      }
+
+      $admin_items[] = [
+        'title' => t('Group access'),
+        'content' => [
+          '#theme' => 'eic_community_well',
+          '#content' => [
+            '#theme' => 'eic_community_harmonica',
+            '#title' => 'Custom restriction',
```
Dromadoros, line 149: Can we also use the translate function here ?

lib/themes/eic_community/includes/preprocess/groups/group.inc
```diff
@@ -0,0 +1,192 @@
...
+
+    $about_items[] = [
+      'title' => t('Group topics'),
+      'items' => $topics,
+    ];
+  }
+
+  if (isset($variables['elements']['field_vocab_geo']['#theme'])) {
+    $render_array_geo = array_filter($variables['elements']['field_vocab_geo'], function($key) {
+      return is_int($key);
```
Dromadoros, line 100: Just wondering why we are doing this ? Can we not just get the field from the group entity and the do the array_map like you did ?

lib/themes/eic_community/eic_community.theme
```diff
@@ -10,6 +10,47 @@
...
+ */
+function eic_community_theme() {
+  $theme_path = drupal_get_path('theme', 'eic_community');
+
+  return [
+    'eic_community_extended_list' => [
+      'template' => 'extended-list',
+      'path' =>  $theme_path . '/patterns/compositions/extended-list',
+      'variables' => [
+        'extra_classes' => NULL,
```
Dromadoros, line 24: Better to respect the type of variable, so if "items" is an array, the default should be [], for title : '', etc

lib/themes/eic_community/includes/preprocess/groups/group.inc
xwero: changed.

lib/themes/eic_community/includes/preprocess/groups/group.inc
xwero: changed

lib/themes/eic_community/includes/preprocess/groups/group.inc
xwero: changed

lib/themes/eic_community/includes/preprocess/groups/group.inc
xwero: I am not sure if `$group->get('field_vocab_geo')->referencedEntities();` does not generate another database query or not. That is why I am using what already is in the render array and manipulate those values.

lib/themes/eic_community/eic_community.theme
xwero: I'm using the templates from storybook as themes, so to be on the safe side i will use NULL so that the frontenders can use simple if conditions.

Dromadoros (approved):

PR #369, opened 2021-07-09
EICNET-1055: No need for flags in group header when group is in draft or pending state

4hardfire:
### Improvements

- Decorate FlagAccessCheck and UnFlagAccessCheck services in order to provide extra access check logic for group flags;
- Improve EICGroupHeaderBlock class in order to show/hide group flags depending on the access (the acces check was not being triggered);
- Decorate RequestAccessCheck service in order to provide extra access check logic for group flags "request delete" and "request archival".

### Tests

- [ ] Create a new group and set the state to pending;
- [ ] Go to the group overview page and check if the group flags are hidden;
- [ ] Check if you get an access denied when trying to flag the group via URL;
- [ ] Change the group state from pending to draft and check if the access to flags is still denied;
- [ ] Change the group state to published and check if you are now able to flag the group via the group header.


lib/modules/eic_groups/src/Access/FlagRequestAccessCheck.php
```diff
@@ -0,0 +1,89 @@
...
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The user account.
+   *
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   An AccessResult object.
+   */
+  public function access(Route $route, RouteMatchInterface $route_match, AccountInterface $account) {
+    $access = $this->requestAccessCheck->access($route, $route_match, $account);
+
+    if ($access->isAllowed()) {
```
lramarojaona, line 57: Same here:
```php
if (!$access->isAllowed()) {
  return $access;  
}

// My code...
```

lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
```diff
@@ -265,4 +275,47 @@ private function getAnonymousLoginLink(GroupInterface $group) {
...
+   * @param \Drupal\group\Entity\GroupInterface $group
+   *   The group entity.
+   *
+   * @return array
+   *   A renderable array of flag links.
+   */
+  private function getGroupFlagLinks(GroupInterface $group) {
+    $group_flags = [];
+
+    if (isset($group->flags)) {
```
lramarojaona, line 290: Again :)
```php
if (empty($group->flags)) {
  return $group_flags;
}

// Code here.
```

lramarojaona (changes requested):

lib/modules/eic_groups/eic_groups.module
```diff
@@ -221,3 +222,38 @@ function eic_groups_group_permission_insert(EntityInterface $entity) {
   \Drupal::classResolver(EntityOperations::class)
     ->groupPermissionInsert($entity);
 }
+
+/**
+ * Implements hook_ENTITY_TYPE_access().
+ */
+function eic_groups_flagging_access(EntityInterface $entity, $operation, AccountInterface $account) {
```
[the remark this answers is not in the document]
4hardfire: There is no need to create a class for this hook because we don't need to make use of any service. This will make the code run faster because there is no need to construct a new class everytime the hook_access is triggered.
Usually I create a class when I need to make use of a service instead of using \Drupal::service().

lib/modules/eic_groups/src/Access/FlagAccessCheck.php
```diff
@@ -0,0 +1,89 @@
...
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The user account.
+   *
+   * @return \Drupal\Core\Access\AccessResult
+   *   An AccessResult object.
+   */
+  public function access(RouteMatchInterface $route_match, FlagInterface $flag, AccountInterface $account) {
+    $access = $this->flagAccessCheck->access($route_match, $flag, $account);
+
+    if ($access->isAllowed()) {
```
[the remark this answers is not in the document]
4hardfire: Done!

lib/modules/eic_groups/src/Access/FlagRequestAccessCheck.php
4hardfire: Done!

lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
4hardfire: Done!

lib/modules/eic_groups/eic_groups.module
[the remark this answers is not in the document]
lramarojaona: Ok understood

PR #371, opened 2021-07-09
Feature/eicnet 1105 switch access check to permissions

lramarojaona:


lib/modules/eic_flags/src/Routing/EntityRequestRouteProvider.php
```diff
@@ -68,10 +68,7 @@ public function getRoutes(EntityTypeInterface $entity_type) {
       $close_request_route = $this->getRouteByTemplate($entity_type, $handler, 'close-request');
       if ($close_request_route) {
         $close_request_route
-          ->setRequirement(
-            '_permission',
-            'close ' . $handler->getType() . ' request'
-          )
+          ->setRequirement('_permission','manage archival deletion requests')
```
korhankale, line 71: just a tiny space missing after the coma

lramarojaona: Done, thanks!

PR #372, opened 2021-07-09
EICNET-1054: A GO shouldn't be able to invite members in a group in draft state

4hardfire:
### Improvements

- Extend Group content enabler plugin "group_invitation" in order to provide extra access check logic for group invitations.

### Tests

- [ ] From the group overview page, check if the link to invite users is not available from the group header;
- [ ] Try to manually go to the group invite user form via URL `/groups/<group-title>/content/add/group_invitation` and check if you get an access denied;
- [ ] As an SA/SCM user, try to publish the group;
- [ ] As a GO, go back to the group overview page and heck if the link to invite users is now available in the group header;
- [ ] Try to access the group invite user form and check if you have access.

### Todo

- [x] Add missing permissions to invite users.


lib/modules/eic_groups/src/Plugin/GroupContentEnabler/GroupInvitation.php
```diff
@@ -0,0 +1,60 @@
...
+    return $operations;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function createAccess(GroupInterface $group, AccountInterface $account) {
+    $access = parent::createAccess($group, $account);
+
+    if ($access->isAllowed()) {
```
lramarojaona, line 40:
```php
if (condition) {
  return;
}
// My code here?
```
:)

lib/modules/eic_groups/src/Plugin/GroupContentEnabler/GroupInvitation.php
```diff
@@ -0,0 +1,60 @@
...
+      switch ($group->get('moderation_state')->value) {
+        case GroupsModerationHelper::GROUP_PENDING_STATE:
+        case GroupsModerationHelper::GROUP_DRAFT_STATE:
```
lramarojaona, line 43: Here I think this is wrong, a GO should be able to invite users in draft mode.

I'll ask on Slack.

lramarojaona (changes requested):

lib/modules/eic_groups/src/Plugin/GroupContentEnabler/GroupInvitation.php
4hardfire: Done!

lib/modules/eic_groups/src/Plugin/GroupContentEnabler/GroupInvitation.php
4hardfire: Julie's answer:

> I’ve always written down - only when published. I know Chris mentioned last time that he wants to add GAs already sooner but we advised against it.
> The reason why is that sometimes the GO forgets the group is unpublished when there are already members in the group and some content. So this would be a big reminder that if the GO actually wants a group with members, they have to put some content in there and publish it.

lramarojaona (approved):

PR #374, opened 2021-07-12
EICNET-1052: As a GO I can publish my draft group after it's been approved by a SA/SCM

4hardfire:
### Improvements

- New custom controller + route to publish a group;
- Create confirmation form to prevent GO from publishing a group by mistake;

### Tests

- [ ] As a trusted user, create a new group
- [ ] As SA/SCM change the group moderation state to **draft**;
- [ ] As GO I should be able to see the link "Publish" in the group header;
- [ ] Click on "Publish". You should be redirected to the group homepage. Note: make sure you check if the group was published successfully;
- [ ] The "Publish" link should not be available anymore in the group header;
- [ ] Make sure you get an access denied if you try to access the confirmation form (`/groups/<group-title>/publish/confirm`) or the route that is responsible for publishing the group (`/groups/<group-title>/publish`)


lib/modules/eic_groups/src/Access/PublishGroupAccessChecker.php
```diff
@@ -0,0 +1,84 @@
...
+          }
+          else {
+            // For group members we only allow access if the user is the group
+            // owner or an "administrator" or a "content_administrator".
+            $access = AccessResult::allowedIf(in_array(UserHelper::ROLE_SITE_ADMINISTRATOR, $account->getRoles(TRUE)) || in_array(UserHelper::ROLE_CONTENT_ADMINISTRATOR, $account->getRoles(TRUE)) ||
+              in_array('group-owner', array_keys($membership->getRoles())))
+              ->addCacheableDependency($account)
+              ->addCacheableDependency($group);
+          }
+        }
```
lramarojaona, line 72: Here it would be better to check:
- The permission/group role
- And always allow SA/CA

For now you're testing 3 times for SA/CA roles, I think it can be done only once.

lib/modules/eic_groups/src/Controller/GroupOperationsController.php
```diff
@@ -0,0 +1,75 @@
...
+        '@group-type' => $group->bundle(),
+        '%label' => $group->label(),
+      ]
+    );
+  }
+
+  /**
+   * Publishes a given group and redirects back to the group homepage.
+   */
+  public function publish(GroupInterface $group) {
```
lramarojaona, line 57: Do we need a route for that?
Can't we just handle this in the submitForm() of the PublishGroupConfirmForm.php?

lramarojaona (changes requested):

lib/modules/eic_groups/src/Controller/GroupOperationsController.php
4hardfire: I'm doing this way because in the future if we change to a more decoupled solution at least that part is done. This is something I learned from past projects. If this is not ok for you I can change that, but in my opinion this is how things should be done.

lramarojaona: No, makes sense, it's ok then.

lib/modules/eic_groups/src/Access/PublishGroupAccessChecker.php
4hardfire: It should look much better. Can you have a look?

lib/modules/eic_groups/src/Access/PublishGroupAccessChecker.php
```diff
@@ -0,0 +1,87 @@
...
+        $membership = $group->getMember($account);
+
+        // If the user is not a member of the group, we deny access.
+        if (!$membership) {
+          break;
+        }
+
+        // For group members we only allow access if the user is the group
+        // owner.
+        $access = AccessResult::allowedIf(in_array('group-owner', array_keys($membership->getRoles())))
```
lramarojaona, line 77: Didn't see that one, we should use th constant for that: https://github.com/EIC-EA/EIC-community-D8/blob/develop/lib/modules/eic_groups/src/EICGroupsHelper.php#L27

lramarojaona (changes requested):

lib/modules/eic_groups/src/Access/PublishGroupAccessChecker.php
4hardfire: Done!

lramarojaona (approved):

PR #376, opened 2021-07-12
EICNET-235: Migrate "document types" taxonomy vocabulary

didierdemaeyer:


lib/modules/eic_migrate/eic_migrate.module
```diff
@@ -0,0 +1,10 @@
+<?php
```
Dromadoros, line 1: Do we really need this whole module for just one config file ? If we decide to keep this module for certain reasons we do not need an empty .module, we can remove this file

didierdemaeyer: The module will contain all of the config files and custom code needed for the migrations. But I will add them in separate PRs.
I will remove the empty .module file

Dromadoros (approved):

PR #377, opened 2021-07-13
[EICNET-1069] chore: Wording change for the close request form

korhankale:


lib/modules/eic_flags/src/Form/RequestCloseForm.php
```diff
@@ -94,6 +94,14 @@ public function getQuestion() {
     );
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getDescription() {
+    return $this->getRequest()
+      ->get('request_type') === RequestTypes::DELETE ? 'This action cannot be undone.' : '';
```
lramarojaona, line 102: Should the text be wrapped inside t() function?

lramarojaona (changes requested):

lramarojaona (approved):

PR #383, opened 2021-07-14
EICNET-1068: When a TU becomes a GM, issue that page needs reloading

4hardfire:
### Improvements

- Add destination URL to redirect the user back to the current page after joining the group.

### Tests

- [ ] Create a group public group
- [ ] As a trusted user go to the group homepage and click join
- [ ] Click join in the confirmation form and check if you are redirected back to the group homepage
- [ ] Leave the group
- [ ] Try to access the join form manually without adding a destination URL `/groups/<group-title>/join`
- [ ] Click on join and check if you are redirected back to the group homepage


lib/modules/eic_groups/eic_groups.module
```diff
@@ -41,9 +41,9 @@ function eic_groups_theme($existing, $type, $theme, $path) {
         'edit_link' => NULL,
         'delete_link' => NULL,
         'actions' => NULL,
-        'has_content' => NULl,
+        'has_content' => NULL,
```
Dromadoros, line 44: This value is a boolean I guess ? Better to put it as FALSE, it will also be better for a dev to understand it directly

lib/modules/eic_groups/src/Hooks/FormOperations.php
```diff
@@ -161,6 +177,26 @@ public function formGroupSubmit(&$form, FormStateInterface $form_state) {
     $this->enableDefaultFeatures($group);
   }
 
+  /**
+   * Custom submit handler to redirect the user to the group homepage.
+   *
+   * This submit handler can be used in in group or group_content forms.
+   */
+  public function formRedirectToGroupHomepage(&$form, FormStateInterface $form_state) {
```
Dromadoros, line 185: missing parameters doc

lib/modules/eic_groups/src/Hooks/FormOperations.php
```diff
@@ -161,6 +177,26 @@ public function formGroupSubmit(&$form, FormStateInterface $form_state) {
...
+    $entity = $form_state->getFormObject()->getEntity();
+
+    switch ($entity->getEntityTypeId()) {
```
Dromadoros, line 188: Are we sure that we will always have group_content or group entity type ? we don't need to put a default redirect ?

lib/modules/eic_groups/eic_groups.module
4hardfire: To be honest... I don't know if that will have any impact in the front-end. @xwero any idea?

Dromadoros: It will have 0 impacts of course, just for better reading I'm saying that

lib/modules/eic_groups/src/Hooks/FormOperations.php
4hardfire: Done!

lib/modules/eic_groups/src/Hooks/FormOperations.php
4hardfire: Yes... this is supposed to be used in the context of a group. Either you are on a group or group_content forms and if not, then we don't need to do anything because the form already has a default redirect.

Dromadoros: you're the best

Dromadoros (approved):

lib/modules/eic_groups/eic_groups.module
4hardfire: @xwero will take care of this in one of his branches. Thanks!

PR #384, opened 2021-07-14
[EICNET-1074] As a TU I can create content by clicking on the button "add content" on the homepage

korhankale:
Introduces an "Add content" button for TU and SA/SCM roles


lib/modules/eic_content/eic_content.module
```diff
@@ -8,3 +8,13 @@
  * This file is no longer required in Drupal 8.
  * @see https://www.drupal.org/node/2217931
  */
+
+function eic_content_theme() {
```
Dromadoros, line 12: missing the implement docblock

Dromadoros (approved):

PR #385, opened 2021-07-14
[EICNET-843]feature: Create search input next to group menu

Dromadoros:
New block/form injected directly next to the group_menu via preprocess. The submit will redirect the user to a route with a query parameter


lib/modules/eic_groups/src/Plugin/Block/SearchMenuGroupBlock.php
```diff
@@ -0,0 +1,32 @@
...
+ *   }
+ * )
+ */
+class SearchMenuGroupBlock extends BlockBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    $form = \Drupal::formBuilder()->getForm(SearchMenuGroupForm::class);
```
4hardfire, line 27: Can you use dependency injection here instead of calling `\Drupal::formBuilder()`? Usually grumphp should throw an error when you try to use drupal services within a class without using proper dependency injection.
In order to avoid spending time searching about how to implement dependency injection in plugins, I already did that for you :):

- Your class should implement the `Drupal\Core\Plugin\ContainerFactoryPluginInterface`;
- Implement the following create method:
```php
/**
 * {@inheritdoc}
 */
public static function create(
  ContainerInterface $container,
  array $configuration,
  $plugin_id,
  $plugin_definition
) {
  return new static(
    $configuration,
    $plugin_id,
    $plugin_definition,
    $container->get('form_builder')
  );
}
```

- Override the constructor:

```php
public function __construct(
  array $configuration,
  $plugin_id,
  $plugin_definition,
  FormBuilderInterface $form_builder,
) {
  parent::__construct($configuration, $plugin_id, $plugin_definition);
  $this->formBuilder = $form_builder;
}
```

lib/modules/eic_groups/src/Plugin/Block/SearchMenuGroupBlock.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Drupal\eic_groups\Plugin\Block;
+
+use Drupal\Core\Block\Annotation\Block;
+use Drupal\Core\Block\BlockBase;
+use Drupal\eic_groups\Form\SearchMenuGroupForm;
+
+/**
+ * Provides an SearchMenuGroupBlock block.
```
4hardfire, line 10: I know it sucks but can you replace "an" by "a"? Thanks ;)

lib/modules/eic_groups/src/Form/SearchMenuGroupForm.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+namespace Drupal\eic_groups\Form;
+
+use Drupal\Core\Form\FormBase;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Class SearchMenuGroupForm
```
4hardfire, line 9: Usually GrumPHP will throw an error because it expects a better description for your class other thatn "Class ClassName". Could we have something like: "Provides a class for the group search form."?

4hardfire (changes requested):
- Check previous comments;
- There is an issue where the icon next to the search input text is not shown.

lib/modules/eic_groups/src/Form/SearchMenuGroupForm.php
Dromadoros: Of course my dear deer

lib/modules/eic_groups/src/Plugin/Block/SearchMenuGroupBlock.php
Dromadoros: Haha thanks for doing my job 🍸 🥥

lib/modules/eic_groups/src/Plugin/Block/SearchMenuGroupBlock.php
```diff
@@ -0,0 +1,63 @@
...
+    );
+  }
+
+  /**
+   * SearchMenuGroupBlock constructor.
+   *
+   * @param array $configuration
+   * @param $plugin_id
+   * @param $plugin_definition
+   * @param \Drupal\Core\Form\FormBuilderInterface $form_builder
```
4hardfire, line 47: Some parameters descriptions are missing:

```php
   * @param array $configuration
   *   The plugin configuration, i.e. an array with configuration values keyed
   *   by configuration option name. The special key 'context' may be used to
   *   initialize the defined contexts by setting it to an array of context
   *   values keyed by context names.
   * @param string $plugin_id
   *   The plugin_id for the plugin instance.
   * @param mixed $plugin_definition
   *   The plugin implementation definition.
   * @param \Drupal\Core\Form\FormBuilderInterface $form_builder
   *   The Form builder service.
```

lib/modules/eic_groups/src/Plugin/Block/SearchMenuGroupBlock.php
```diff
@@ -0,0 +1,63 @@
...
+ *   admin_label = @Translation("EIC Search Menu Group"),
+ *   category = @Translation("European Innovation Council"),
+ *   context_definitions = {
+ *     "group" = @ContextDefinition("entity:group", required = FALSE)
+ *   }
+ * )
+ */
+class SearchMenuGroupBlock extends BlockBase implements ContainerFactoryPluginInterface {
+
+  /** @var \Drupal\Core\Form\FormBuilderInterface $formBuilder */
```
4hardfire, line 26: Replace with:

```php
  /**
   * The form builder service.
   *
   * @var \Drupal\Core\Form\FormBuilderInterface
   */
```

4hardfire (approved):
Seems good to me! Just inform Julie that the icon next to the text input needs to be fixed in a follow-up ticket.

PR #386, opened 2021-07-14
EICNET-1044: Create separate business role for site admin separate from user 1 admin

4hardfire:
### Changes

- Create new role site_admin and keep the same permissions as content_administrator (both global and group permissions).
- Re-factor code and configurations with the new site_admin role;

### Tests

- [ ] As a site admin, check if I can manage the platform the same way as the content_administrator role.


lib/modules/eic_groups/eic_groups.module
```diff
@@ -237,7 +237,7 @@ function eic_groups_group_access(EntityInterface $entity, $operation, AccountInt
   switch ($operation) {
     case 'delete':
       // Deny access to delete group if the group is NOT in pending state and
-      // the user is not an "administator" or a "content_administrator".
+      // the user is not an "site_admin" or a "content_administrator".
```
Dromadoros, line 240: Sorry to comment just for that but ... "a" site_admin :D

4hardfire: Done :)

Dromadoros (approved):

PR #391, opened 2021-07-15
EICNET-1023: updated story data for footer and fixed styling for titles

filipvt1:


xwero:
I'm going to add the drupal code, to make the fix done.

PR #393, opened 2021-07-16
EICNET-979 added flags component to print list of flags

filipvt1:


lib/themes/eic_community/patterns/compositions/discussion/discussion.full.html.twig
```diff
@@ -45,6 +45,9 @@
     {% endblock %}
 
     {% block sidebar %}
+      {% if flags is not empty and flags.items is not empty %}
```
xwero, line 48: Most of the full pages aren't going to be used because the content in drupal is split up. So you can use a simple if like `{% if flags %}`.

xwero (approved):

PR #394, opened 2021-07-16
Feature/eicnet 1120

speeters1:


lib/themes/eic_community/patterns/compositions/featured-content-grid.html.twig
```diff
@@ -8,9 +8,8 @@
       <div class="ecl-featured-content-grid__items-wrapper">
 
         <div class="ecl-featured-content-grid__items">
-          {% for item in items %}
+          {% for key, item in items %}
```
xwero, line 11: Is the key variable used? If not, i would not change the loop.

xwero (changes requested):

lib/themes/eic_community/patterns/compositions/featured-content-grid.html.twig
speeters1: It's used on line 67 to check if the item is the first one of the for loop. Normally I would use the loop variable but that doesn't seem to work

xwero (approved):

PR #396, opened 2021-07-16
EICNET-1113: Implement the missing statistics to the group header

4hardfire:
### Features

- Instal module entity_usage;
- Create module eic_group_statistics;
- Create new custom schema for saving group statistics;
- Implement statistics counter for group members, comments and files (discussion, document, wiki page and gallery files);
- Delete group statistics when group is deleted;
- Update group statistics when installing the module.
- Index group statistics in SOLR.

### Tests

- [x] Check there is only 1 member in the group statistics
- [x] Add a discussion and insert multiple comments.
- [x] Check if the number of comments match the ones shown in the group header;
- [x] Create 2 documents with the same media;
- [x] Create another document with a different media from the previous ones.
- [x] Check if the number of documents shown in the group header is 2 (because 2 documents use the same media so we count as 1)
- [x] Change the last created document and reference the same media used in the other 2.
- [x] Check the group statistics in the group header and make sure the number of files is 1.

### Missing

- Create a cron job to update all group statistics everyday.

### DIFF
--- a/lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.install
+++ b/lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.install
@@ -73,3 +73,35 @@ function eic_group_statistics_schema() {
 
   return $schema;
 }
+
+/**
+ * Implements hook_install().
+ */
+function eic_group_statistics_install() {
+  /**
+   * Gets the Group statistics storage.
+   *
+   * @var \Drupal\group\Entity\GroupInterface[] $groups
+   */
+  $group_statistics_storage = \Drupal::service('eic_group_statistics.storage');
+  /**
+   * Loads all groups.
+   *
+   * @var \Drupal\group\Entity\GroupInterface[] $groups
+   */
+  $groups = \Drupal::entityTypeManager()->getStorage('group')->loadMultiple();
+  /**
+   * Initializes array of group statistics.
+   *
+   * @var \Drupal\eic_group_statistics\GroupStatistics[] $groups_statistics
+   */
+  $groups_statistics = [];
+
+  // Calculate statistics for each group.
+  foreach ($groups as $group) {
+    $groups_statistics[] = $group_statistics_storage->calculateGroupStatistics($group);
+  }
+
+  // Update groups statistics.
+  $group_statistics_storage->setMultipleGroupStatistics($groups_statistics);
+}
--- a/lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.install
+++ b/lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.install
@@ -5,8 +5,6 @@
  * Install, update and uninstall functions for the EIC Group Statistics module.
  */
 
-use Drupal\eic_group_statistics\GroupStatisticsHelper;
-
 /**
  * Implements hook_schema().
  */
@@ -81,6 +79,6 @@ function eic_group_statistics_schema() {
  */
 function eic_group_statistics_install() {
   // Update groups statistics for all groups.
-  \Drupal::classResolver(GroupStatisticsHelper::class)
+  \Drupal::service('eic_group_statistics.helper')
     ->updateAllGroupsStatistics();
 }
--- a/lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.install
+++ b/lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.install
@@ -5,6 +5,8 @@
  * Install, update and uninstall functions for the EIC Group Statistics module.
  */
 
+use Drupal\eic_group_statistics\GroupStatisticsHelper;
+
 /**
  * Implements hook_schema().
  */
@@ -78,30 +80,7 @@ function eic_group_statistics_schema() {
  * Implements hook_install().
  */
 function eic_group_statistics_install() {
-  /**
-   * Gets the Group statistics storage.
-   *
-   * @var \Drupal\group\Entity\GroupInterface[] $groups
-   */
-  $group_statistics_storage = \Drupal::service('eic_group_statistics.storage');
-  /**
-   * Loads all groups.
-   *
-   * @var \Drupal\group\Entity\GroupInterface[] $groups
-   */
-  $groups = \Drupal::entityTypeManager()->getStorage('group')->loadMultiple();
-  /**
-   * Initializes array of group statistics.
-   *
-   * @var \Drupal\eic_group_statistics\GroupStatistics[] $groups_statistics
-   */
-  $groups_statistics = [];
-
-  // Calculate statistics for each group.
-  foreach ($groups as $group) {
-    $groups_statistics[] = $group_statistics_storage->calculateGroupStatistics($group);
-  }
-
-  // Update groups statistics.
-  $group_statistics_storage->setMultipleGroupStatistics($groups_statistics);
+  // Update groups statistics for all groups.
+  \Drupal::classResolver(GroupStatisticsHelper::class)
+    ->updateAllGroupsStatistics();
 }
--- a/lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.install
+++ b/lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.install
@@ -1,0 +1,75 @@
+<?php
+
+/**
+ * @file
+ * Install, update and uninstall functions for the EIC Group Statistics module.
+ */
+
+/**
+ * Implements hook_schema().
+ */
+function eic_group_statistics_schema() {
+  $schema['eic_group_statistics'] = [
+    'description' => 'Stores group counters information.',
+    'fields' => [
+      'id' => [
+        'type' => 'serial',
+        'not null' => TRUE,
+        'description' => 'Primary Key: Unique record ID.',
+      ],
+      'gid' => [
+        'type' => 'int',
+        'unsigned' => TRUE,
+        'not null' => TRUE,
+        'default' => 0,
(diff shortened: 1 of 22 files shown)


xwero (approved):
Adding members and comments increments the statistics.
And the statistics are shown according to the design.

lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.install
```diff
@@ -0,0 +1,107 @@
...
+   */
+  $groups_statistics = [];
+
+  // Calculate statistics for each group.
+  foreach ($groups as $group) {
+    $groups_statistics[] = $group_statistics_storage->calculateGroupStatistics($group);
+  }
+
+  // Update groups statistics.
+  $group_statistics_storage->setMultipleGroupStatistics($groups_statistics);
```
lramarojaona, line 106: Should we put this logic as a separate function for later use (somewhere in the module)?

lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.module
```diff
@@ -0,0 +1,139 @@
...
+ */
+function eic_group_statistics_group_delete(EntityInterface $entity) {
+  \Drupal::service('eic_group_statistics.storage')->deleteGroupStatistics($entity);
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_insert().
+ */
+function eic_group_statistics_group_content_insert(EntityInterface $entity) {
+  switch ($entity->bundle()) {
```
lramarojaona, line 24: Should we put the switch directly in the EntityOperations->groupContentInsert() method instead of here?
For consistency with the other ones.

lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.module
```diff
@@ -0,0 +1,139 @@
...
+    return;
+  }
+
+  $group_content = reset($group_contents);
+
+  // Increments group comments statistic counter.
+  \Drupal::service('eic_group_statistics.storage')->increment($group_content->getGroup(), GroupStatisticsStorageInterface::STAT_TYPE_COMMENTS);
+
+  // Re-index group statistics.
+  \Drupal::service('eic_group_statistics.search_api.reindex')->reindexItem($group_content->getGroup());
```
lramarojaona, line 92: Why don't you use the EntityOperations class here?

lib/modules/eic_statistics/modules/eic_group_statistics/src/GroupStatisticsHelper.php
```diff
@@ -0,0 +1,82 @@
...
+    $this->entityTypeManager = $entity_type_manager;
+    $this->groupStatisticsStorage = $group_statistics_storage;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function loadGroupStatistics(GroupInterface $group) {
+    $group_statistics = $this->groupStatisticsStorage->load($group);
+    return $group_statistics;
```
lramarojaona, line 45: Maybe `return $this->groupStatisticsStorage->load($group);` to avoid creating a useless variable?

lib/modules/eic_statistics/modules/eic_group_statistics/src/Hooks/EntityOperations.php
```diff
@@ -0,0 +1,451 @@
...
+        /** @var \Drupal\node\NodeInterface $node */
+        $node = $entity->getEntity();
+
+        $medias = [];
+
+        $media_fields = [
+          'field_document_media',
+          'field_related_downloads',
+          'field_photos',
+        ];
```
lramarojaona, line 116: Maybe we could get this list dynamically, fields in nodes (only?) that reference media items.

https://drupal.stackexchange.com/questions/222731/how-to-get-all-field-name-of-field-type

lib/modules/eic_statistics/modules/eic_group_statistics/src/Plugin/search_api/processor/GroupStatisticsIndexer.php
```diff
@@ -0,0 +1,237 @@
...
+   *
+   * @return array
+   *   Array containing the type of group statistics group.
+   */
+  public function getGroupStatisticTypes() {
+    return [
+      GroupStatisticsStorageInterface::STAT_TYPE_MEMBERS => $this->t('Members'),
+      GroupStatisticsStorageInterface::STAT_TYPE_COMMENTS => $this->t('Comments'),
+      GroupStatisticsStorageInterface::STAT_TYPE_FILES => $this->t('Files'),
+      GroupStatisticsStorageInterface::STAT_TYPE_EVENTS => $this->t('Events'),
```
lramarojaona, line 91: Maybe we could have the labels in the same class as the constants, so we can re-use them elsewhere.
Also, probably not in the interface class.
I was looking for an example but cannot find out right now.

lramarojaona (changes requested):

lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.module
4hardfire: I wouldn't recommend that because you don't need to instantiate a class with services if the group content bundle is not those listed in that switch. At least this way we improve performance.

lib/modules/eic_statistics/modules/eic_group_statistics/src/Plugin/search_api/processor/GroupStatisticsIndexer.php
4hardfire: Usually I define constants in the interfaces. Not sure if this is a standard but I remember doing this in other projects. Maybe I should look into drupal core classes to see how they do it.

lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.install
4hardfire: Done.

lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.module
4hardfire: I think this hook is so simple that I thought there was no need to instantiate EntityOperations. Also because we will then instantiate 2 more services which we won't need and we might have a lot of comments being added and removed all the time. I don't know but maybe we get a little performance improvements. Services and classes are useful but sometimes it slows down things.

lib/modules/eic_statistics/modules/eic_group_statistics/src/GroupStatisticsHelper.php
4hardfire: Done.

lib/modules/eic_statistics/modules/eic_group_statistics/src/Hooks/EntityOperations.php
4hardfire: There is a field_related_media in Wiki pages and it can be an image or a remote video. I am not sure if this should be taken into account for the statistics. That's why I hardcoded the fields we needed. Do you have any suggestion? I guess we need to hardcode something at least.

lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.module
lramarojaona: Mmm, yeah ok, we're doing the same in other hooks, let's see

lib/modules/eic_statistics/modules/eic_group_statistics/src/Plugin/search_api/processor/GroupStatisticsIndexer.php
lramarojaona: Found the example I was looking for: https://github.com/EIC-EA/EIC-community-D8/blob/develop/lib/modules/eic_messages/src/MessageTemplateTypes.php
There are other examples accross our codebase, just search for `const ` (with a trailing space)

lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.module
lramarojaona: Ok

lib/modules/eic_statistics/modules/eic_group_statistics/src/Hooks/EntityOperations.php
lramarojaona: Ok, if we need to hardcode some of them, then maybe use a config? (no ui)

4hardfire: In the future we could improve this and probably create a configuration form to select which bundles and fields to use for the file statistics.

lramarojaona: Ok, maybe add a todo then, so we can find future potential improvements easily

lib/modules/eic_statistics/modules/eic_group_statistics/src/Plugin/search_api/processor/GroupStatisticsIndexer.php
4hardfire: Done.

lib/modules/eic_statistics/modules/eic_group_statistics/src/Hooks/EntityOperations.php
4hardfire: Done.

lramarojaona (approved):
Looks good!

PR #397, opened 2021-07-16
Feature/eicnet 1121

speeters1:


xwero (changes requested):
The statistics icons are still on the left. In the figma design they are on the right side.

speeters1:
@xwero they are aligned to the right for me, can you check again?

xwero (approved):

PR #400, opened 2021-07-16
EICNET-785 update template for group in draft

filipvt1:


xwero (changes requested):
The like and the follow links above the statistics in the header are not removed.

filipvt1:
Removed flags for groups in draft. Also noticed this error on pending state so also fixed there.

xwero (approved):

PR #407, opened 2021-07-22
EICNET-1128: Show drupal and group status messages in storybook styling.

xwero:
Tests:

- [ ] Create a group in the pending state
- [ ] *the group created message and the pending message should be visible and styled*
- [ ] Change the state of the group to draft
- [ ] *An updated group, an approved and a get the conversation started message should be visible and styled*
- [ ] Change the state of the group to published
- [ ] *An updated group, a published and a get the conversation started message should be visible and styled*


didierdemaeyer:
Tested the PR. 

The group status messages and the get the conversation started are OK, but I don't see the drupal messages (created group, updated group) at all.

didierdemaeyer (changes requested):
Tested the PR.

The group status messages and the get the conversation started are OK, but I don't see the drupal messages (created group, updated group) at all.

Dromadoros (approved):

didierdemaeyer (approved):

PR #411, opened 2021-07-23
EICNET-981: Based login redirect on current url instead of group url.

xwero:
Tests:

- Create published public group
- Create discussion
- Go as anonymous user to discussion
- *You should see that the destination in the login link is the current page*

Extra bugfix:

I saw the group header was displayed twice, so I updated the context.


4hardfire (approved):
Seems ok to me. Thanks for getting rid of the double group header :)

PR #413, opened 2021-07-23
EICNET-1126: Changed copy on about page.

xwero:
Tests:

- Create a visibility public and joining methods open group
- Create a visibility public and joining methods request group
- Create a visibility private group
- Create a visibility community members only and joining methods open group
- Create a visibility community members only and joining methods request group
- *Then you should see the correct copy*
- Create a visibility custom restrictions and joining methods open group
- Create a visibility custom restriction and joining methods request group
- *Then you should see the correct copy and the custom restrictions*

Bonus bugfix:

- Wiki group header is not displayed twice any more


lib/themes/eic_community/includes/preprocess/groups/group.inc
```diff
@@ -140,55 +140,87 @@ function _preprocess_about_page(&$variables) {
...
+      ];
+    }
 
     $sections[] = [
       'id' => 'admin',
       'title' => t('Permission and visibility'),
       'content' => [
         '#theme' => 'eic_community_extended_list',
         '#title' => t('Group Permissions'),
+        '#description' => t("Below is explained to whom this group is visible in order to understand the private or public character of the group's content."),
```
Dromadoros, line 223: Better to continue using single quote, to be consistant

Dromadoros:
This one should have a back-end label and not FE, we only have Drupal code

lib/themes/eic_community/includes/preprocess/groups/group.inc
xwero: There is a single quote in the text, that is why I needed to use double quotes.

Dromadoros: Ok but you can keep the single quotes wrapper and just put a \ in front of your quote in the text

xwero: That is true, but I not a fan of adding extra characters to text just to keep the quotes consistent. I only use breaks if it needed.

Dromadoros (approved):

PR #418, opened 2021-07-26
[EICNET-957] A a site user I'm not allowed to create group content outside of a group

korhankale:
Disable access to the node.add route for wiki_pages and discussion bundles

### To Test

- [x] Log in with an account without the 'bypass node access' permission
- [x] Go to /node/add/discussion
- [x] You should get a 403

### DIFF
--- a/config/sync/user.role.content_administrator.yml
+++ b/config/sync/user.role.content_administrator.yml
@@ -11,7 +11,6 @@ permissions:
   - 'access group overview'
   - 'add content to books'
   - 'administer comments'
-  - 'bypass node access'
   - 'close archive request'
   - 'close delete request'
   - 'create group group'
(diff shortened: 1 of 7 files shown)


4hardfire (changes requested):
I found an issue with News pages. As a trusted user I can create News pages. Can you just ask Julie why we have this permission? Probably we were using in the past while we didn't have all the user roles.
In any case... I guess we should get rid of that permission for Trusted users.

korhankale:
> I found an issue with News pages. As a trusted user I can create News pages. Can you just ask Julie why we have this permission? Probably we were using in the past while we didn't have all the user roles.
> In any case... I guess we should get rid of that permission for Trusted users.

Question asked. Don't think this is an issue, this ticket states that they should be able https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1074
(even if it was, lets keep the scope of this PR to the ticket and create a new one for it)

4hardfire (approved):
It's ok to me. Just inform Julie about the permission concerning the News for TUs.

PR #420, opened 2021-07-26
EICNET-1130: Remove the helptext shown on the LinkedIn field on a member profile

4hardfire:
### Improvements

Remove helper text from LinkedIn field and prepend string "in/" if missing from the values.

### Tests

- [ ] Go to the profile form of the user 1 -> `/user/1/member`;
- [ ] Make sure the LinkedIn field doesn't have any helper text;
- [ ] Try to add a LinkedIn profile without the "in/" like `https://www.linkedin.com/bernardo-paulino-425954ab`. Save it, go back to the form and make sure the value is shown as `in/bernardo-paulino-425954ab`

### DIFF
--- a/lib/modules/eic_user/src/Hooks/FieldWidgetOperations.php
+++ b/lib/modules/eic_user/src/Hooks/FieldWidgetOperations.php
@@ -3,7 +3,6 @@
 namespace Drupal\eic_user\Hooks;
 
 use Drupal\Core\Form\FormStateInterface;
-use Drupal\Core\StringTranslation\StringTranslationTrait;
 
 /**
  * Class FieldWidgetOperations.
@@ -12,8 +11,6 @@ use Drupal\Core\StringTranslation\StringTranslationTrait;
  */
 class FieldWidgetOperations {
 
-  use StringTranslationTrait;
-
   /**
    * Implements hook_field_widget_social_links_form_alter().
    */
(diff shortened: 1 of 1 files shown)


korhankale:
Why not keeping the 'in' part outside the value and append it during displays?
image
Another question, is there plans to do a frontend validation on these fields? If I do not follow the backend rules and add whatever I want I get strange results

E.g: add /in/1234 instead of in/1234
image

lib/modules/eic_user/src/Hooks/FieldWidgetOperations.php
```diff
@@ -56,23 +45,18 @@ public function socialLinksFieldValidate($element, FormStateInterface $form_stat
     foreach ($form_state_values as $key => $value) {
       if ($value['social'] === $social_network_name) {
         $form_state_values[$key]['link'] = str_replace($social_network_base_url, '', $value['link']);
+
+        // Exception for LinkedIn since we need to prepend "in/" if missing.
+        if ($social_network_name === 'linkedin') {
+          if (substr($form_state_values[$key]['link'], 0, 3) !== 'in/') {
```
Dromadoros, line 51: What's happening if a user does stupid thing like enter '/in/StupidUser' ? It will be 'in//in/StupidUser' ?

4hardfire: It should stay "in/StupidUser". I improved the code a little.

4hardfire:
> Why not keeping the 'in' part outside the value and append it during displays?
> image
> Another question, are we going to have a frontend validation on these fields? If I do not follow the backend rules and add whatever I want I get strange results
> 
> E.g: add /in/1234 instead of in/1234
> image

If you now add "/in/" in the beginning, it will be replaced by "in/". This is an edge case because if you do something like "/i-dont-care/hello-user" you will get an form error.

Dromadoros (approved):

PR #423, opened 2021-07-26
[EICNET-783]feature: Create the global overview with the content context

Dromadoros:
Create the overview for the global search with the context from which group the content comes from

How to test it : 

- [x] Go to block layout page (/admin/structure/block)
- [x] Add the block : **EIC Search Overview** into the content region and configure it as you wish
- [x] Be sure that you have some content (news, story, book, group, page, wiki). So you can see different renderer into the search
- [x] Clear your cache and go to the homepage
- [x] You should see the global overview with different content in it
- [x] You can search on the body and title
- [x] Bonus : check also if there is the group permission feature, normally content from group that you are not allowed to see is not shown (2 browsers, private nav and connected user)

image
image


4hardfire (changes requested):
There is an issue when we change the group visibility from public to private. Apparently the group content is still shown in the list. This seems to be an issue with the backend that is not related to your view.
I also found some issues when searching in the title of a basic page. I have a lorem ipsum text in the title and it doesn't work very well. Ex: "Proident adipisicing Lorem id do tempor non Lorem voluptate magna magna cupidatat do exercitation excepteur.". When I search for **proident** (all small letters) I get 0 results. Couldn't figure out why this is happening. Could you have a look?

4hardfire (approved):
Seems ok. Issues related to SOLR can be done in a follow-up ticket.

xwero (approved):

PR #426, opened 2021-07-26
[EICNET-580]feature: As a group member I want to view the group member overview

Dromadoros:
Integrate a new configuration to handle the group context via the current route

How to test it :

- [x] Go to context page settings (/admin/structure/context)
- [x] Go to Group - Global and add the block EIC Search Overview
- [x] On the block configuration, select the member_list or member_gallery source and check the "prefilter group" (facets/sorts as you wish)
- [x] Go into a detail group
- [x] You will have your search with prefiltered content by the current group


4hardfire (approved):
Looks good to me!

PR #427, opened 2021-07-27
EICNET-1140: Allow bypass access check for user 1 and drupal admin

4hardfire:
### Improvements

- Create method in UserHelper service to check when if an account is a power user.
- Re-factor access checkers in eic_groups module and use the UserHelper service to check if a user is a power user.

### Tests

- [ ] As a trusted user, create a new public group
- [ ] As user 1 or Drupal admin, I should always be able to join + post content + delete group + edit group + invite user.


lib/modules/eic_user/src/UserHelper.php
```diff
@@ -111,4 +118,33 @@ public function getSitePowerUsers(bool $active_only = TRUE) {
...
+  /**
+   * Returns an array of uids for all the "power users".
+   *
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The user account to check if is a power user.
+   *
+   * @return bool
+   *   TRUE if user is a power user.
+   */
+  public function isPowerUser(AccountInterface $account) {
```
korhankale, line 130: Could be a static method and be used as UserHelper::isPowerUser(AccountInterface $account), wdyt? Would avoid injecting/instantiating the service for this (unless we have/might need more than a static context)

lib/modules/eic_groups/eic_groups.module
```diff
@@ -228,7 +227,7 @@ function eic_groups_entity_field_access($operation, FieldDefinitionInterface $fi
         // Deny access if it's a new group and the user doesn't have
         // "site_admin" or "content_administrator" roles.
         if ($entity->isNew()) {
-          $access = AccessResult::forbiddenIf(!in_array(UserHelper::ROLE_SITE_ADMINISTRATOR, $account->getRoles(TRUE)) && !in_array(UserHelper::ROLE_CONTENT_ADMINISTRATOR, $account->getRoles(TRUE)))
+          $access = AccessResult::forbiddenIf(!\Drupal::service('eic_user.helper')->isPowerUSer($account))
```
Dromadoros, line 230: Your method name as one more letter in caps "isPower**US**er"

4hardfire: Nice catch. Fixed!

lib/modules/eic_user/src/UserHelper.php
4hardfire: I honestly thought about it. Glad that you brought that in, otherwise I would forget about it. Thanks :)

4hardfire: It should be done!

korhankale (approved):

Dromadoros (approved):

PR #428, opened 2021-07-27
Feature/eicnet 1143 activity stream checkbox node edit form

lramarojaona:
### Tests

- [x] Create a discussion inside a group
- [x] You should have the "Post message in the activity stream" checkbox enabled by default
- [x] Save => you should have a "stream_discussion_insert_update" message created (check the message table in the db since we don't have an activity stream yet)
- [x] Edit the discussion
- [x] You should have the "Post message in the activity stream" checkbox disabled by default
- [x] Save => no message should be created

### Remarks

- We don't have the activity stream yet, so you need to check the created messages directly in the DB
- MessageCreator classes have been turned into services
- For now only the message for discussions is created
- The content of the activity stream message is not handled yet

### DIFF
--- a/lib/modules/eic_messages/eic_messages.module
+++ b/lib/modules/eic_messages/eic_messages.module
@@ -23,7 +23,7 @@ use Drupal\node\Entity\NodeType;
  * Implements hook_ENTITY_TYPE_insert().
  */
 function eic_messages_group_insert(EntityInterface $entity) {
-  /** @var GroupMessageCreator $class */
+  /** @var Drupal\eic_messages\Hooks\GroupMessageCreator $class */
   $class = \Drupal::classResolver(GroupMessageCreator::class);
   $class->groupInsert($entity);
 }
@@ -32,7 +32,7 @@ function eic_messages_group_insert(EntityInterface $entity) {
  * Implements hook_ENTITY_TYPE_update().
  */
 function eic_messages_group_update(EntityInterface $entity) {
-  /** @var GroupMessageCreator $class */
+  /** @var Drupal\eic_messages\Hooks\GroupMessageCreator $class */
   $class = \Drupal::classResolver(GroupMessageCreator::class);
   $class->groupUpdate($entity);
 }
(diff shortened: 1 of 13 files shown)


lib/modules/eic_messages/eic_messages.module
```diff
@@ -29,7 +32,7 @@ function eic_messages_group_insert(EntityInterface $entity) {
  * Implements hook_ENTITY_TYPE_update().
  */
 function eic_messages_group_update(EntityInterface $entity) {
-  /** @var GroupMessageCreator $class */
+  /** @var Drupal\eic_messages\Hooks\GroupMessageCreator $class */
   $class = \Drupal::classResolver(GroupMessageCreator::class);
```
korhankale, line 36: We should get these using `Drupal::service($id)` since they are defined and live within the service container (even though that what the class resolver does in this case, it will spare us a few function calls [cfr](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21DependencyInjection%21ClassResolver.php/function/ClassResolver%3A%3AgetInstanceFromDefinition/8.2.x))

korhankale:
Do we allow to post an edit on the activity feed? The last 2 checkboxes are not corresponding to what I have on my local 
image
Actually it is unchecked but if I check it again and hit save, a message is created again (maybe that's the correct behavior? In this case, I guess we will display a message "Discussion X has been updated" ?)

lramarojaona:
@korhankale , yes users should be able to post an activity of an update if they choose so.
What do you mean by "The last 2 checkboxes are not corresponding to what I have on my local"?

korhankale:
> @korhankale , yes users should be able to post an activity of an update if they choose so.
> What do you mean by "The last 2 checkboxes are not corresponding to what I have on my local"?

image

From PR's description, but if we are allowing them to check it again then they are ok

lramarojaona:
Ah yes, I meant "unchecked". Disabled can be confusing indeed

lib/modules/eic_messages/eic_messages.module
lramarojaona: This should be ok now

korhankale (approved):

PR #431, opened 2021-07-28
EICNET-1151: Added search form to main menu.

xwero:
Tests:

* You should see the search form in the main menu on all the pages.
* When you add input in the search field, you should go to the search page and see your input in the search input of that page.

### DIFF
--- a/lib/themes/eic_community/templates/navigation/menu--main.html.twig
+++ b/lib/themes/eic_community/templates/navigation/menu--main.html.twig
@@ -7,9 +7,70 @@
  */
 #}
 {% if links %}
-  {{ pattern('navigation_menu', {
-    label: 'Menu'|t,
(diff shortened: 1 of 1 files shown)


lib/themes/eic_community/templates/navigation/menu--main.html.twig
```diff
@@ -7,9 +7,70 @@
...
+    },
+    searchform: {
+      extra_attributes: [
+        {
+          name: "method",
+          value: "get"
+        },
+        {
+          name: "action",
+          value: path("view.global_overviews.search")
```
Dromadoros, line 53: For now we can keep this but today or tomorrow I need to remember to update that cause we will have a new endpoint for our react overview

Dromadoros (approved):

PR #432, opened 2021-07-28
[EICNET-906]feature: Add moderation state to Wiki page so member grou…

Dromadoros:
Add moderation state to Wiki page so member group creating wiki page can moderate his content

How to test it : 

- [ ] Go into group page and add a Wiki page
- [ ] You should see a select list with different state "Draft, Published, Unpublished"


config/sync/workflows.workflow.group_wiki.yml
```diff
@@ -0,0 +1,57 @@
+uuid: 4b5eb05e-e026-4924-9267-b4b3decc2dca
+langcode: en
```
korhankale, line 2: Shouldn't we use the same workflow as news & stories? However it has "news_stories" as id which isn't accurate. If we don't, shouldn't we add the archived state also?

Dromadoros: AT first I add the content type to news & stories but the ticket mentionned to only draft, unpublished and published. Maybe something to ask to Julie

korhankale (approved):

PR #445, opened 2021-07-30
[EICNET-1175]feature: add dedicated pages for people, groups and search. disable some views feature

Dromadoros:
Add 3 endpoints for global : groups, people, search.

How to test it : 

- [x] Do not forget to index solr via search API
- [x] Go to /search, /people, /groups
- [x] Check if it's the new react block


config/sync/context.context.members_overview.yml
```diff
@@ -1,151 +1,53 @@
...
 group: 'Global overviews'
 description: 'This context is used for the Members overview'
 requireAllConditions: false
 disabled: false
 conditions:
-  view_inclusion:
-    id: view_inclusion
+  route:
+    id: route
+    routes: eic_groups.people
```
4hardfire, line 16: Shouldn't this route be called something like eic_search.people? This is the global search of EIC users.

4hardfire (changes requested):
Since we are not using the view "global_overviews", can we disable it in the backend `/admin/structure/views`?

4hardfire (approved):
Seems ok to me!

PR #461, opened 2021-08-04
EICNET-1185: Delete the recommend flag on stories

4hardfire:
### Improvements

- Remove recommend flag from CTypes News and Story;
- Fix Notice: Undefined property: Drupal\eic_content\ContentBreadcrumbBuilder::$account in Drupal\eic_content\ContentBreadcrumbBuilder when viewing a News or Story page.

### Tests

- [ ] Create a News and a Story and publish them.
- [ ] Check if the flag "Recommend" has been removed.
- [ ] Go to `/admin/reports/dblog` and make sure you have no notice: "**Notice: Undefined property: Drupal\eic_content\ContentBreadcrumbBuilder::$account in Drupal\eic_content\ContentBreadcrumbBuilder when viewing a News or Story page.**"

### DIFF
--- a/lib/modules/eic_content/src/ContentBreadcrumbBuilder.php
+++ b/lib/modules/eic_content/src/ContentBreadcrumbBuilder.php
@@ -6,6 +6,7 @@ use Drupal\Core\Breadcrumb\Breadcrumb;
 use Drupal\Core\Breadcrumb\BreadcrumbBuilderInterface;
 use Drupal\Core\Link;
 use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Session\AccountProxyInterface;
 use Drupal\Core\StringTranslation\StringTranslationTrait;
 use Drupal\node\NodeInterface;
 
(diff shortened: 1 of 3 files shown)


lib/modules/eic_content/src/ContentBreadcrumbBuilder.php
```diff
@@ -16,6 +17,23 @@ class ContentBreadcrumbBuilder implements BreadcrumbBuilderInterface {
...
+  protected $currentUser;
+
+  /**
+   * Constructs a new ContentBreadcrumbBuilder object.
+   *
+   * @param \Drupal\Core\Session\AccountProxyInterface $current_user
+   *   The current user service.
+   */
+  public function __construct(AccountProxyInterface $current_user) {
+    $this->currentUser = $current_user;
```
korhankale, line 34: I don't really see how the current user gets injected without having it added as argument in the service definition. Did I miss something?

4hardfire: I forgot to commit that. It should be ok now.

korhankale (approved):

PR #462, opened 2021-08-04
[EICNET-1175]feature: Add spellcheck on search and more filters

Dromadoros:
Implement spellcheck on all overview search

How to test : 

- [ ] Go to /search
- [ ] Write for example "grou" it will propose you a word
- [ ] Check also if you have the geo and topics filter on this global search


4hardfire (approved):
Seems ok to me!

PR #463, opened 2021-08-04
[EICNET-959]feature: Add query to prefilter private content for anonymous

Dromadoros:
How to test : 

- [x] Create/edit a news/story and set it as private
- [x] Go to global search (/search), find your news/story as a connected user
- [x] Logout and do the same thing, you should not see the content anymore


4hardfire (changes requested):
When creating a user with only "Authenticated" role, the user can see both private and public News/Stories in the global search which is not ok because only Trusted users should be able to see private News/Stories. If you try to access a private News/Story you will still get an access denied which is ok. The problem is just in the global search. Can you have a look?

lib/modules/eic_search/src/Controller/SolrSearchController.php
```diff
@@ -250,6 +250,12 @@ private function generateQueryPrivateContent(&$fq) {
       return;
     }
 
+    $roles = \Drupal::currentUser()->getRoles();
+
+    if (in_array('trusted_user', $roles)) {
```
4hardfire, line 255: You can can use `UserHelper::ROLE_TRUSTED_USER` here.

4hardfire (approved):
Looks fine to me!

PR #464, opened 2021-08-05
[EICNET-1187]feature: Create video content type

Dromadoros:
How to test :

- [x] Add new "video" content
- [x] Check fields requirements here : https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1187


4hardfire (changes requested):
I have the following error message after I click "Preview" when creating a new video node:

**Location** | http://eic.europa.eu/node/preview/e1cbb7a9-50ae-4e8f-a34f-554025f5d297/full
**Referer** | http://eic.europa.eu/node/add/video
**Message** | `Drupal\Core\Entity\EntityStorageException: Cannot load GroupContent entities for an unsaved entity. in Drupal\group\Entity\Storage\GroupContentStorage->loadByEntity() (line 92 of /var/www/html/web/modules/contrib/group/src/Entity/Storage/GroupContentStorage.php).`

Could you have a look?

4hardfire (approved):
Seems ok to me! We just need to create a follow-up ticket to fix the issue with the "Preview".

PR #387, opened 2021-07-14
Feature/eicnet 851 create overview pages

lramarojaona:
### Global overviews
- Create a global overview at `/admin/community/overview-pages`
- Go the path you defined and see if you get the page content
- Check that SA/CA can edit them but not create or delete.
- Check that anonymous or any logged in used can access it

### Group overviews
- Go to a group and try to access `group/{group}/discussions`, `group/{group}/people` and `group/{group}/search`

## Known issues and remarks
- Read the README file :)
- Global overviews:
  - For now you can only select standard block type, we'll need to change this once we have the `eic_search_overview` block type merged (see https://github.com/EIC-EA/EIC-community-D8/pull/271)
-  Group overviews:
   - The pages show the react overveiw, but not with final facets/sorts etc...
   - Some permissions are manipulated by the Group features
   - Old group menu items might mess up the group menu (better to do a clean install)

PR #467, opened 2021-08-06
EICNET-1098: Create a group feature for "Files"

4hardfire:
### Features/Improvements

- Create group permissions: "access wiki overview" and "access files overview";
- Create group feature plugin to show/hide group menu link for the files overview page (this should take into account permission "access files overview");
- Enable or disable group feature "Wiki" based on "access wiki overview" permission; Allow public roles to view Wiki overview page of a public group.

### Tests

- [x] When viewing the group page, the group menu link "Files" should be enabled by default
- [x] As anonymous user or logged in user I can see the group menu link "Files" because it's a public group
- [x] As GO disable the group feature "Files"
- [x] As anonymous or logged in user, make sure you cannot see the group menu link "Files" anymore
- [x] Test the enable and disable the group feature "Wiki" and check if it has the same behavior. Note that this group feature will never be enabled by default.


lib/modules/eic_groups/eic_groups.routing.yml
```diff
@@ -38,3 +38,12 @@ eic_groups.group.highlight_content:
     _highlight_group_content: 'TRUE'
     node: \d+
     group: \d+
+
+eic_overviews.groups.overview_page.files:
+  path: '/group/{group}/library'
+  defaults:
+    # @todo: update controller namespace once we have the class.
```
didierdemaeyer, line 45: @4hardfire Can this todo already be handled or should this be done in a different PR?

4hardfire: It should be handled after PR #387 gets merged.

didierdemaeyer (approved):

PR #469, opened 2021-08-09
[EICNET-1191] feat: Add a permission and endpoint to highlight group content

korhankale:
### To Test

- [ ] Identify yourself as a group owner or group admin using your favorite API client
- [ ] Send a POST request to `/group/{group}/highlight-content/{node}` with appropriate values
- [ ] Your content should be flagged and the `its_flag_highlight_content_count` property in solr be equal to 1

### DIFF
--- a/lib/modules/eic_groups/src/Access/HighlightGroupContent.php
+++ b/lib/modules/eic_groups/src/Access/HighlightGroupContent.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace Drupal\eic_groups\Access;
+
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Routing\Access\AccessInterface;
+use Drupal\Core\Routing\RouteMatchInterface;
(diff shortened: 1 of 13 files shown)


lib/modules/eic_groups/src/Access/HighlightGroupContent.php
```diff
@@ -0,0 +1,58 @@
...
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Routing\Access\AccessInterface;
+use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Session\AccountProxy;
+use Drupal\group\Entity\GroupInterface;
+use Drupal\node\NodeInterface;
+use Drupal\user\Entity\User;
+
+/**
+ * Class FlagGroupContentAccessCheck
```
Dromadoros, line 14: Not the correct class name

Dromadoros (approved):

PR #481, opened 2021-08-12
Feature/eicnet 1238 file list detail

filipvt1:
- Added filelist component
- Added filelist detail bundle


xwero (changes requested):
There is more whitespace between the previous page link and the rest of the content than in the figma file.

There is no comments section in figma or documentation. Just remove it.

The file list detail is a part of the group stories, I would add it there instead of a new bundle

xwero (approved):

PR #487, opened 2021-08-16
[EICNET-1179]feature: Show draft/pending groups for matching group owner

Dromadoros:
How to test it: 

- [x] Be sure you have published and draft/pending groups
- [x] Also check if you are a group owner in one of current draft/pending groups
- [x] Go to /groups and you will see published groups + draft/pending group of yours
- [x] Verify with another user or anonymous if you still can see this group

image


4hardfire (approved):
Seems ok to me!

PR #497, opened 2021-08-17
[EICNET-1248]feat: Add loadmore option on the overview

Dromadoros:
How to test:

- [x] Go to the context "Group - global" and add the block "EIC Search Overview" into the region "before content"
- [x] Configuration : datasource = activity stream  and checked -> prefiltered by group
- [x] Go to a group detail, add a discussion (check the post activity at the end), after that edit the discussion 6-7 times and always check the activity post
- [x] Return to the current group and you should see a load more link

PS : there is no front for now, only data are showing


didierdemaeyer (changes requested):
@Dromadoros Not able to test this.

The EIC Search overview block is visible in the before content if the group doesn't have activity messages, but when I add a discussion it disappears.

Error:
```
Error: Minified React error #130; visit https://reactjs.org/docs/error-decoder.html?invariant=130&args[]=undefined&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings. — 
```

didierdemaeyer (approved):
@Dromadoros OK.

Like you said, the `npm install` and `npm run react-production` commands still needed to be run.

PR #498, opened 2021-08-18
[EICNET-1083] As SA/SCM/GO/GA I can delete items from the activity stream in my group

korhankale:
### To Test

- [x] Add a comment/discussion/wiki_page
- [x] You should have an 'ss_operation_type' = 'created' for it in Solr
- [x] Delete the message with the request DELETE /group/{group_id}/item/{message_id}
- [x] The corresponding document in Solr should also be deleted


didierdemaeyer (approved):
Tested deletion of 3 activity items (comment, discussion and wiki_page).
Code looks good.

The only thing I find a bit strange is the DELETE URL. /group/{group_id}/item/{message_id}
Since we're using /groups/{group_alias}/... for other pages and /item/{message_id} is very generic and made me question "what item?", as opposed to /activity/{message_id} which clearly refers to the activity stream.

I'll approve because I assume this remark is not that important.

korhankale:
> Tested deletion of 3 activity items (comment, discussion and wiki_page).
> Code looks good.
> 
> The only thing I find a bit strange is the DELETE URL. /group/{group_id}/item/{message_id}
> Since we're using /groups/{group_alias}/... for other pages and /item/{message_id} is very generic and made me question "what item?", as opposed to /activity/{message_id} which clearly refers to the activity stream.
> 
> I'll approve because I assume this remark is not that important and perhaps logical if you have more Drupal experience.

You're right for the URL 😅 let me change it to /group/{group}/}activity-item/{message}

PR #505, opened 2021-08-19
Feature/eicnet 1159 import languages taxonomy

lramarojaona:
- See the `README.md` file in the `eic_webservices` module to set-up the webservice connection
- Run `drush mim --group=eic_smed_taxonomy`
- You should have terms in following vocabularies:
  - Topics & Subtopics (3 levels)
  - Languages (1 level)

Connection information can be found on this page (bottom): https://citnet.tech.ec.europa.eu/CITnet/confluence/display/EICNET/WS+-+Taxonomy+webservice


didierdemaeyer (approved):
@lramarojaona All items migrated correctly, but is there a SMED dashboard key in the data which can also be saved?

In D7 the taxonomies have a field `c4m_dashboard_key` which contains, what I assume, a SMED ID that we could use to map the taxonomy term references in the content migration instead of using the term name.

lib/modules/eic_webservices/src/Plugin/migrate/source/EicSmedUrl.php
```diff
@@ -0,0 +1,53 @@
...
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, MigrationInterface $migration) {
+    // Get the current migration tags.
+    $migration_tags = $migration->getMigrationTags();
+
+    // Inject the username and password values into the authentication
+    // configuration.
+    if (in_array('eic_smed_api_authentication', $migration_tags)) {
+      if (isset($configuration['authentication'])) {
+        $configuration['authentication']['username'] = Settings::get('smed_api_taxonomy_username');
```
korhankale, line 42: Maybe we should add default values for these settings in this [file](https://github.com/EIC-EA/EIC-community-D8/blob/develop/web/sites/default/default.settings.local.php) (even empty or null). WDYT?

lramarojaona: Nice idea, this is done.

lramarojaona:
@didierdemaeyer,
You're right, I created the field for all taxonomies and added it in the migration for the existing migrations (languages and thematicsTopics).

didierdemaeyer (changes requested):
@lramarojaona The `field_smed_id` is not imported. I think you still need to import the changes into config/sync.

Also I looked at the migration map tables and it seems like these IDs are not corresponding to the c4m_dashboard_key values for languages. (E.g. in the map table: sourceid1 = 16, and in c4m_dashboard_key it's `bg`.

Those for the topics seems OK, but can you make sure.

didierdemaeyer (approved):

PR #507, opened 2021-08-19
Feature/eicnet 1255: Added activity stream.

speeters1:


xwero (commented):
The only thing i don't see in storybook are the blue and red cross that is in the figma file. But it is possible this functionality is not needed? I will ask it in the ticket.

xwero (approved):

PR #508, opened 2021-08-19
[EICNET-783]fix: Use correctly the group directly instead of loading wrong group content

Dromadoros:
How to test:

- [x] Re-index Solr
- [x] Go to global search (/search) and check if it's the correct group context name/url


didierdemaeyer (approved):
@Dromadoros Everything is shown as found in the correct group, except for the page that's created by default "Proident adipisicing Lorem id do tempor non Lorem voluptate magna magna cupidatat do exercitation excepteur."

This is node 1 and already created when installing the site. I created another page content type and this doesn't have this issue.

Screenshot 2021-08-23 at 10 43 50

PR #510, opened 2021-08-19
[EICNET-1232]feat: Sort by default on creation date

Dromadoros:
To test:

- [ ] Go to global search (/search) and check if items are ordered by CREATION date in DESC


didierdemaeyer:
@Dromadoros Initial sorting is OK. But something is wrong with pagination and "Old" and "Recent" sorting seem to be switched.

didierdemaeyer (changes requested):
See comment above.

PR #518, opened 2021-08-20
EICNET-1225: Provide file download counter, node page views and entity flags (likes)

4hardfire:
#### Download counter
- [x] Create a document with 2 medias (make sure you are on private download method, this is provided in the config)
- [x] Download the media 1
- [x] Download the media 2
- [x] The node download counter should now have a total of file downloads of 2
- [x] Delete media 2
- [x] The node download counter should now have a total of file downloads of 1

#### Flag counters (like)
- [x] Like the content
- [x] All the flags and count for those flags are provided for all entities

#### Page views
- [x] Refresh the page
- [ ] The number of page views should be incremented

### Remarks

- For now the counters are not displayed and are only provided in the build array of the entities
- Page views: if you have AdBlock enabled, the page views ajax requests are blocked, please disable it for your site


lib/modules/eic_content/src/Hooks/EntityOperations.php
```diff
@@ -0,0 +1,64 @@
...
+   * @param \Drupal\Core\Entity\Display\EntityViewDisplayInterface $display
+   *   The entity view display holding the display options.
+   * @param string $view_mode
+   *   The view mode the entity is rendered in.
+   */
+  public function nodeView(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display, $view_mode) {
+    $node_views = $this->nodeStatisticsDatabaseStorage->fetchView($entity->id());
+    $build['page_views'] = [
+      '#markup' => '',
+      '#value' => $node_views->getTotalCount(),
```
4hardfire, line 60: @lramarojaona I get an error `Error: Call to a member function getTotalCount() on bool in Drupal\eic_content\Hooks\EntityOperations->nodeView() (line 60 of /var/www/html/lib/modules/eic_content/src/Hooks/EntityOperations)`.

I guess we need a condition `if (!$node_views)` because it will return FALSE when you create a new node and you're viewing it for the first time.

4hardfire:
@lramarojaona the number of page views only gets updated once we clear the cache.

lramarojaona:
> 
> 
> @lramarojaona the number of page views only gets updated once we clear the cache.

@4hardfire , this is now fixed

4hardfire:
> > @lramarojaona the number of page views only gets updated once we clear the cache.
> 
> @4hardfire , this is now fixed

@lramarojaona I haven't tested yet but it seems like it will work only for the file downloads counter and not page views counter.

lib/modules/eic_media/modules/eic_media_statistics/src/EntityFileDownloadCount.php
```diff
@@ -0,0 +1,209 @@
...
+    foreach ($entity_fields as $field) {
+      /** @var \Drupal\Core\Field\FieldDefinitionInterface $field */
+      switch ($field->getType()) {
+        // @todo Handle paragraphs.
+        case 'entity_reference':
+          $target_entity_type = $field->getConfig($entity->bundle())->getSetting('target_type');
+          if ($target_entity_type == 'media') {
+            // Load the entities.
+            foreach ($entity->get($field->getName())->referencedEntities() as $referenced_entity) {
+              $sub_entity_result = self::countFileDownloads($referenced_entity);
```
4hardfire, line 154: @lramarojaona I get a warning in my IDE -> `Non static method 'countFileDownloads' should not be called statically.`. I guess you should be calling the method using `$sub_entity_result = $this->countFileDownloads($referenced_entity)`

PR #520, opened 2021-08-20
EICNET-477: Add document language field to document media type

4hardfire:
### Changes

- Add new taxonomy term field "Language" in Media Type "Document";
- Show the media document field language when viewing the teaser version.

### Tests

- [ ] Create a new group
- [ ] Create a new Document for the group and when creating the media, select a language for the media (create a taxonomy term with a name other than "English")
- [ ] Go to the group homepage and check if it's the right language that is shown in the document teaser

### DIFF
--- a/config/sync/core.entity_form_display.media.eic_document.default.yml
+++ b/config/sync/core.entity_form_display.media.eic_document.default.yml
@@ -4,6 +4,7 @@ status: true
 dependencies:
   config:
     - field.field.media.eic_document.field_body
+    - field.field.media.eic_document.field_language
     - field.field.media.eic_document.field_media_file
     - field.field.media.eic_document.field_vocab_topics
     - media.type.eic_document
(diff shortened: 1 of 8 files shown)


didierdemaeyer (changes requested):
@4hardfire This doesn't seem to be working.

I've created a Node Document with language Dutch selected and attached one Media Document to it with language French, but neither show up on the teaser.

Also not that the Media reference field on the Node Document accepts multiple references, which isn't taken into account in code I think.

Screenshot 2021-08-24 at 09 59 54
Screenshot 2021-08-24 at 10 02 57

4hardfire:
@didierdemaeyer can you pull the latest changes and clear the cache. It should work with multiple files as well. But when there are multiple documents for now we show the first one that finds.

didierdemaeyer (approved):

PR #541, opened 2021-08-26
EICNET-1304: Show comments count in news and stories

4hardfire:
### Improvements

- Provide comments counter to the renderable array of the entity view.

### Tests

- Currently there are no visual tests to do until we have implemented the comments counter in the front-end.

### NOTES

- Only merge this branch after PR #518


lib/modules/eic_comments/eic_comments.module
```diff
@@ -33,3 +36,11 @@ function eic_comments_configure_comment_form($form) {
   $form[0]['format']['#access'] = FALSE;
   return $form;
 }
+
+/**
+ * Implements hook_entity_view().
+ */
+function eic_comments_entity_view(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display, $view_mode) {
```
lramarojaona, line 43: @4hardfire , I wonder if we should apply this to nodes only, since I don't think we'll have comments for other entity types.
So we don't need to cycle through all fields for all entity types

4hardfire: Done!

lramarojaona: @4hardfire , I would have expected a hook_node_view() :)

lramarojaona (approved):

PR #548, opened 2021-08-27
EICNET-1336: File download counter refactoring

4hardfire:
### Improvements

- Download and install module "media_entity_download";
- Create custom access check to control access to the media file download route;
- Move code from hook eic_media_statistics_file_download to the MediaFileDownloadController.

### Tests

- [ ] Create a document with 2 medias (make sure you are on private download method, this is provided in the config)
- [ ] Download the media 1
- [ ] Download the media 2
- [ ] The node download counter should now have a total of file downloads of 2
- [ ] Delete media 2
- [ ] The node download counter should now have a total of file downloads of 1

### Remarks

For now the file download counter is not displayed and it's only provided in the build array of the entities.

### Notes

- [x] We should merge after #518 get merged.


lib/modules/eic_media/modules/eic_media_statistics/src/Controller/MediaFileDownloadController.php
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function download(MediaInterface $media) {
+    try {
+      $response = parent::download($media);
+    }
+    catch (\Exception $e) {
```
korhankale, line 47: I don't really get the point of catching the exception to throw it after 🤔

4hardfire: @korhankale I am doing it in case a new module wants to override my class. But should I return nothing instead?

lib/modules/eic_media/modules/eic_media_statistics/src/Access/MediaFileDownloadAccessCheck.php
```diff
@@ -0,0 +1,89 @@
...
+
+        if ($access->isAllowed()) {
+          // We found an entity the user has access to. The user can download
+          // the media.
+          return $access;
+        }
+      }
+    }
+
+    // Allow access if the user is not a power user.
```
lramarojaona, line 85: @4hardfire , small typo here => Allow access if the user is a power user

lib/modules/eic_media/modules/eic_media_statistics/src/EntityFileDownloadCount.php
```diff
@@ -161,6 +162,12 @@ protected function countFileDownloads(EntityInterface $entity) {
           break;
 
         case 'file':
+        case 'image':
+          // We discard the thumbnail field.
```
lramarojaona, line 166: @4hardfire , is this needed?
The EntityFileDownlaodCount is quite generic, I don't see a reason to introduce custom logic here?

lib/modules/eic_media/modules/eic_media_statistics/src/EventSubscriber/MediaDownloadRouteSubscriber.php
```diff
@@ -0,0 +1,40 @@
...
+ */
+class MediaDownloadRouteSubscriber extends RouteSubscriberBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function alterRoutes(RouteCollection $collection) {
+    foreach ($collection->all() as $route) {
+      // Set custom access check for media download route.
+      if (strpos($route->getPath(), '/media/{media}/download') === 0) {
```
lramarojaona, line 20: @4hardfire , I think you don't need to loop through all route, you can just get the one you need `media_entity_download.download`
See https://www.drupal.org/docs/drupal-apis/routing-system/altering-existing-routes-and-adding-new-routes-based-on-dynamic-ones

lramarojaona (changes requested):

lib/modules/eic_media/modules/eic_media_statistics/src/Access/MediaFileDownloadAccessCheck.php
4hardfire: Fixed.

lib/modules/eic_media/modules/eic_media_statistics/src/EntityFileDownloadCount.php
4hardfire: Also, this method might fail if we have multiple file fields (not media fields) attached to an entity.

lramarojaona: @4hardfire 
Ok, I fixed it, now it will handle multiple file/image fields.

Can you improve just a bit your comment about the thumbnail as discussed?

Thanks!

lib/modules/eic_media/modules/eic_media_statistics/src/Controller/MediaFileDownloadController.php
lramarojaona: @4hardfire , @korhankale ,
What's the status on this?

korhankale: Hi
I'd say method/class overriding doesn't have to do anything here, the exception is still "forwarded" after being catched. Generally what should be done is either we stop the execution (e.g: we don't catch the exception) or we catch it, log and return/redirect something/somewhere (especially in controller classes since they are meant to respond to a route after a user request).

In this case $response won't be defined if we let the process continue (and seems like we need it). So we don't need to catch the exception (unless we want to display something custom)

Dromadoros: We can remove this try & catch, if you want to throw the error, just let it naturally done by calling the function

lib/modules/eic_media/modules/eic_media_statistics/src/EntityFileDownloadCount.php
4hardfire: Done.

korhankale (approved):

PR #551, opened 2021-08-30
EICNET-1283: As a GM I can edit every wiki page from my groups

4hardfire:
### Features

- Create custom node property to control node edit access for group members;
- Override GroupContentAccessControlHandler in order to create extra access conditions for members to be able to edit nodes;
- Edit group permissions (allow members to edit own wiki pages)

### Tests

- [ ] Create a new public group and publish it
- [ ] Create a new wiki page and check the option "Editable by members"
- [ ] As a TU, join the group and try to edit the wiki page. You should be allowed to edit.
- [ ] As GO, edit the wiki page and uncheck the option "Editable by members"
- [ ] As a TU, you should not be able to edit the wiki page anymore

### IMPORTANT

- The site needs to be re-installed in order for the new node property to be installed. Otherwise we will need to create a **hook_updateN**.
- **We have now a drush command, we need to fire it when deploying to staging: drush eg:create-field-access**


korhankale:
As a trusted user I don't really see the edit button, where it should be?
A question: Should I, as a trusted_user in a group, be able to give a value/see the field for the "editable by members" field?

Also, maybe not related to this PR but I'm getting this error when creating a wiki page 😅 
image

korhankale (approved):

PR #552, opened 2021-08-30
Feature/EICNET-1018: Sort handled requests by closed_date (desc)

korhankale:


4hardfire (changes requested):
The view is still not showing the order in DESC. I guess you also need to update that on the table display settings in the view because it is still ASC set as default.

Dromadoros (approved):

PR #556, opened 2021-08-30
[EICNET-1236]fix: Remove first spaces in domain mails field

Dromadoros:
How to test:

- [x] Edit a group and modify email domains with space after comma "test1@test.com, test2@test.com"


4hardfire (approved):
I works like a charm :ok_hand:

PR #558, opened 2021-08-30
[EICNET-1356]fix: Avoid adding array into array when value currently exists for a document

Dromadoros:
How to test it:

- [x] Re-index all elements from search API
- [x] Be sure you have content with multiple "Topics"
- [x] Go to search overview (/search) check if you have weird topic value in left column filter or in teaser


4hardfire:
@Dromadoros Everything seems to be fine, although... I think I found a new issue. Here is the test I did:

- Create a group
- Create a document in the group with 2 topics (Topic AB + Topic ABC)
- When I go to the search page I can see the facets with 2 entries (1 for each topic)
- If I select one of the topics the list gets filtered correctly
- If I select both topics I get no results. I should have had at least 1 result which is the document with 2 topics. Any idea why this is happening?

4hardfire (approved):
Seems ok. However, we will need a follow up ticket to fix the issue I mentioned in the comments.

PR #560, opened 2021-08-31
[EICNET-1104] chore: Override messages links

korhankale:
### To Test

- [ ] Click on Content > Messages in the admin menu
- [ ] You should land on /admin/content/messages


lramarojaona:
@korhankale ,
This is a silly ticket, I would say we do nothing for that and don't merge it.
See: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1104?focusedCommentId=4916398&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4916398

PR #567, opened 2021-09-02
EICNET-XXXX: Fix issue when field_language doesn't exist.

lramarojaona:
### Fixes

- Fix PHP error: `InvalidArgumentException: Field field_language is unknown. in Drupal\Core\Entity\ContentEntityBase->getTranslatedField()`

### Tests

- [ ] Create a group
- [ ] Create a document
- [ ] Go on group homepage
- [ ] You should see the group homepage and not a  PHP fatal error


4hardfire (approved):
Seems ok to me!

PR #571, opened 2021-09-03
EICNET-962: As a GO I want to choose a URL alias for my group

4hardfire:
### Improvements

- Allow SA/SCM users to edit group url alias;
- Create redirect for the old group url alias when group alias changes;
- Create redirect for the old group content and node url alias when group alias changes.

### Tests

- [x] Create a new discussion in the group
- [x] Edit the group and change the URL alias
- [x] Try to navigate to the old group URL and check if are redirected to the new URL
- [x] Run cron
- [x] Try to navigate to the old group discussion URL and check if are redirected to the new URL


korhankale:
Last point failed on my side after running the cron:
- Is it normal that the URL alias for the discussion isn't updated?
- something is strange with the redirect for the discussion, I don't see the old -> new one (should be /groups/my-group/discussions/discussion-my-group as "from path" in this case)

image

image

4hardfire:
> Last point failed on my side after running the cron:
> 
> * Is it normal that the URL alias for the discussion isn't updated?
> * something is strange with the redirect for the discussion, I don't see the old -> new one (should be /groups/my-group/discussions/discussion-my-group as "from path" in this case)
> 
> image
> 
> image

@korhankale I removed the custom code that was creating the redirects. Apparently the redirect module already has a setting to create redirects when the URL changes. Can you re-test?

korhankale:
> > Last point failed on my side after running the cron:
> > 
> > * Is it normal that the URL alias for the discussion isn't updated?
> > * something is strange with the redirect for the discussion, I don't see the old -> new one (should be /groups/my-group/discussions/discussion-my-group as "from path" in this case)
> 
> @korhankale I removed the custom code that was creating the redirects. Apparently the redirect module already has a setting to create redirects when the URL changes. Can you re-test?

Indeed it works as expected now 😄 However, don't know if it is related or not, seems like we can't delete a group anymore. Do you have the same issue?
image

Dromadoros (approved):

korhankale (approved):

PR #573, opened 2021-09-03
EICNET-1373: Change "invite member" default settings in group creation form based on joining method

4hardfire:
### Improvements

- Set "invite members" enabled by default when creating a new group with joining method set to "tu_group_membership_request".

### Tests

- [ ] Create a new public group and set joining method to "Request (for trusted users)"
- [ ] Go to the group edit for and check if the option **Invite members** is enabled
- [ ] Create a new private group and save it
- [ ] Go to the group edit for and check if the option **Invite members** is enabled

### DIFF
--- a/lib/modules/oec_group_flex/src/Entity/Form/GroupForm.php
+++ b/lib/modules/oec_group_flex/src/Entity/Form/GroupForm.php
@@ -218,6 +218,9 @@ class GroupForm extends GroupFormBase {
    * {@inheritdoc}
    */
   public function save(array $form, FormStateInterface $form_state): int {
+    $group = $form_state->getFormObject()->getEntity();
+    $is_new = $group->isNew();
+
     $return = parent::save($form, $form_state);
 
     if (!$groupFlexSettings = $this->getGroupFlexSettingsFormValues($form, $form_state)) {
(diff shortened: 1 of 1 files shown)


lib/modules/oec_group_flex/src/Entity/Form/GroupForm.php
```diff
@@ -267,6 +270,18 @@ public function save(array $form, FormStateInterface $form_state): int {
       }
     }
 
+    // If the group is new and the selected joining method is
+    // "tu_group_membership_request", we enable member invitations by default.
+    // This logic needs to be triggered during this process because the group
+    // visibility and joining methods are saved previously after the group is
+    // saved in the Database.
+    if ($is_new &&
+      ($groupFlexSettings['settings']['visibility']['plugin_id'] === 'private' || $groupFlexSettings['settings']['joining_methods'] === 'tu_group_membership_request')
```
Dromadoros, line 279: Maybe use const ?

```
GroupVisibilityType::GROUP_VISIBILITY_PRIVATE
GroupJoiningMethodType::GROUP_JOINING_METHOD_TU_MEMBERSHIP_REQUEST
```

4hardfire: We cannot do that because this module will probably be used in other community websites and if we use the constants we will create a dependency on eic_groups module and we don't want that.

Dromadoros (approved):

PR #574, opened 2021-09-06
EICNET-1359: Error when changing visibility of the group

4hardfire:
### Fixes

- Use method `loadByGroup()` from GroupPermissionsManager class to load group permissions when saving the group form. Before it was using the method `getGroupPermission()` but this method was returning an old object.

### Tests

- [x] As a TU, create a new group
- [x] As SA, change the group status to draft
- [x] As a TU, publish the group
- [x] As a TU, edit the group and change the visibilty from Private to Community members
- [x] Make sure you get no errors like `The website encountered an unexpected error. Please try again later.` after saving the group.


lramarojaona (approved):
Works fine!

PR #576, opened 2021-09-06
EICNET-1369: Adapt comment section on video detail page (group library)

4hardfire:
### Changes

- Create contributors field in CType Video and apply contributors logic from discussions;
- Move logic from eic_content_discussion to eic_content.

### Tests

- [ ] Create a new video
- [ ] Go to the video detail page and check if the title of the comments section is "Replies"
- [ ] Go to the video edit form and check if the owner is already set as contributor
- [ ] As a TU, join the group and leave a comment in the video
- [ ] As GO, go to the video edit form and check if the TU has been added to the list of contributors

### DIFF
--- a/lib/modules/eic_content/eic_content.module
+++ b/lib/modules/eic_content/eic_content.module
@@ -12,6 +12,8 @@
 use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\eic_content\Hooks\EntityOperations;
+use Drupal\eic_content\Hooks\EntityOperationsContributor;
+use Drupal\node\NodeInterface;
 
 /**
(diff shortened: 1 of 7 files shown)


lib/modules/eic_content/eic_content.module
```diff
@@ -32,4 +34,33 @@ function eic_content_theme() {
...
+
+  }
+
+}
+
+/**
+ * Implements hook_entity_presave().
+ */
+function eic_content_entity_presave(EntityInterface $entity) {
+  if (FALSE === $entity instanceof NodeInterface) {
```
Dromadoros, line 53: Usually I see `!$entity instanceof NodeInterface`  but it is also ok like that just strange :D

Dromadoros (approved):

PR #582, opened 2021-09-07
EICNET-1367: Add social share block to Video CT.

lramarojaona:
### Tests

- [ ] Create a Video content
- [ ] Check that the Social Share block is ont the right column (like in Stories)


lib/themes/eic_community/includes/preprocess/common.inc
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Returns the build of the social share block.
+ *
+ * @return array|bool
+ *   The build array of the block or FALSE if not found.
+ */
+function _eic_community_get_social_share_block() {
+  $share_block = \Drupal::service('plugin.manager.block')
+    ->createInstance('social_share', []);
+  if ($share_block instanceof BlockPluginInterface) {
```
Dromadoros, line 19: Isn't it cleaner to put the positive return as the final one ? WDYT ?

Something like this if (!$share_block instanceof BlockPluginInterface) { return FALSE; } then the build

lramarojaona: Yes this is done

PR #587, opened 2021-09-08
[EICNET-1385] feat: Add an apache/tika container

korhankale:
### To test

- [ ] Start the project
- [ ] Go to /admin/config/search/search_api_attachments
- [ ] Be sure the following configuration is applied and save
image
- [ ] Tika integration should be validated


lramarojaona (approved):
Works as expecetd, thanks! :)

PR #594, opened 2021-09-10
EICNET-1397: Allow members to create videos in groups

4hardfire:
### Features

- Add permissions to create videos in groups.

### Tests

- [x] Create a group, enable the Files feature and publish it
- [x] As a TU, join the group and check if you are able to create Videos via group header


lramarojaona (approved):
It works, but only for new groups.
I guess always the same issue with old groups, so this is not an issue right now

PR #596, opened 2021-09-13
EICNET-1401: Fix download button URL in CT Document

4hardfire:
### Improvements

- CT Document: update download button URL in order to trigger file counter and show updated file statistics.

### Tests

- [x] Create a group
- [x] Create a new document in the group and upload multiple documents
- [x] Go to the document detail page and click on the download buttons
- [x] Check if the download counters are updated


lramarojaona (changes requested):
@4hardfire ,
The download counter doesn't get updated, I need to do a cache rebuild in order to have the updated counters.

4hardfire:
> @4hardfire ,
> The download counter doesn't get updated, I need to do a cache rebuild in order to have the updated counters.

Can you pull the latest developments and re-test?

lramarojaona (approved):
It works, but we'll still need to think of node caching.
Can be merged for now.

PR #602, opened 2021-09-14
[EICNET-798]fix: site_admin should also see content with community restriction

Dromadoros:
How to test:

- [ ] Check if you are site_admin role (not trusted user just)
- [ ] In search for content locked for community, a site_admin should be able to see it

### DIFF
--- a/lib/modules/eic_search/src/Plugin/search_api/processor/GroupAccessContent.php
+++ b/lib/modules/eic_search/src/Plugin/search_api/processor/GroupAccessContent.php
@@ -142,7 +142,7 @@ class GroupAccessContent extends ProcessorPluginBase {
     ';
 
     // Restricted community group, only trusted_user role can view.
-    if (!$user->isAnonymous() && ($user->hasRole('trusted_user') || $user->hasRole('site_admin'))) {
+    if (!$user->isAnonymous() && ($user->hasRole(UserHelper::ROLE_TRUSTED_USER) || $user->hasRole(UserHelper::ROLE_SITE_ADMINISTRATOR) || UserHelper::isPowerUser($this->getCurrentUser()))) {
       $query .= ' OR (ss_group_visibility:' . GroupVisibilityType::GROUP_VISIBILITY_COMMUNITY . ')';
     }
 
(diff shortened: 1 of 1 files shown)


korhankale (approved):

lib/modules/eic_search/src/Plugin/search_api/processor/GroupAccessContent.php
```diff
@@ -142,7 +142,7 @@ private function buildGroupVisibilityQuery(): string {
     ';
 
     // Restricted community group, only trusted_user role can view.
-    if (!$user->isAnonymous() && $user->hasRole('trusted_user')) {
+    if (!$user->isAnonymous() && ($user->hasRole('trusted_user') || $user->hasRole('site_admin'))) {
```
lramarojaona, line 145: @Dromadoros ,
You should always use the defined constants for the roles:
https://github.com/EIC-EA/EIC-community-D8/blob/develop/lib/modules/eic_user/src/UserHelper.php#L48

Also in this case, I guess SA/SCM should be able to see restricted groups, I suggest that you use this method:
https://github.com/EIC-EA/EIC-community-D8/blob/develop/lib/modules/eic_user/src/UserHelper.php#L130

lramarojaona (changes requested):

lib/modules/eic_search/src/Plugin/search_api/processor/GroupAccessContent.php
Dromadoros: Forgot to use constants, you're also right with the function, bedankt

lramarojaona: Then I guess you can just do
```
if (!$user->isAnonymous() && ($user->hasRole(UserHelper::ROLE_TRUSTED_USER) || UserHelper::isPowerUser($this->getCurrentUser()))) {
```
cause `UserHelper::ROLE_SITE_ADMINISTRATOR` is already included in `UserHelper::isPowerUser()` ;)

PR #605, opened 2021-09-14
EICNET-1345: SA/SCM should be able to edit all group content of any group (Part 2)

4hardfire:
### Fixes

- Override GroupNode class in order to fix issue where group operations were empty for SA/SCM users that were not members of the group;
- Allow SA/SCM to create group content (override permission system).

### Tests

- [x] As a TU, create a new group
- [x] Make sure you can't add new group content until the group status is set to draft
- [x] As SA/SCM, go to the group homepage and check if you have the **Post content** dropdown in the group header
- [x] Edit the group status to draft
- [x] As TU, go to the group homepage and check if you are able to create content

### DIFF
--- a/lib/modules/eic_groups/src/Access/GroupContentCreateEntityAccessCheck.php
+++ b/lib/modules/eic_groups/src/Access/GroupContentCreateEntityAccessCheck.php
@@ -64,26 +64,25 @@ class GroupContentCreateEntityAccessCheck extends GroupContentCreateEntityAccess
   public function access(Route $route, AccountInterface $account, GroupInterface $group, $plugin_id) {
     $access = $this->groupContentCreateEntityAccessCheck->access($route, $account, $group, $plugin_id);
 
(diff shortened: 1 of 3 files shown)


lib/modules/eic_groups/src/Access/GroupContentCreateEntityAccessCheck.php
```diff
@@ -64,26 +64,25 @@ public function __construct(
   public function access(Route $route, AccountInterface $account, GroupInterface $group, $plugin_id) {
     $access = $this->groupContentCreateEntityAccessCheck->access($route, $account, $group, $plugin_id);
 
+    // Edge case where we make sure that no user (even user 1 or a user with
+    // Drupal administrator role) can create new book pages inside groups.
+    // There should only be 1 book page per group and it's automatically
+    // created after creating the group.
```
lramarojaona, line 70: Good to have such comments, so we know the intention of the code! :)

lramarojaona (approved):
It works!

PR #609, opened 2021-09-15
[EICNET-1417]feat: Create comment system in react

Dromadoros:
Implement a new block on discussion detail that inject a react component to display all comments with async actions. New controller needed for all endpoints and also the permissions check for each action

How to test it:

- [x] Go to a discussion detail from a group
- [x] Try to add comment, edit, delete, ... Also reply to it, you can have 3 level (parent, level 2, level 3)
- [x] You also have a load more after 3 items shown (even in answers)


4hardfire (changes requested):
### Tests

**As SA/SCM** - OK

- Create new public group and a discussion
- I can post comments, edit, delete, request deletion, request archival, reply and like

**As TU** - NOT OK

- Go to the discussion page created by SA/SCM
- I **can't** ~~post comments,~~ request deletion, request archival ~~and reply~~ (**NOT OK**). We need to check with PM if the user can request deletion and archival
- I can like (OK)
- ~~I joined the group but I still **can't** post comments, request deletion, request archival and reply (**NOT OK**)~~
- I created a new public group and a discussion
- I go to the discussion detail page and I see the list of comments from the other discussion which belongs to the other group (**NOT OK**)
- ~~I still **can't** post comments, request deletion, request archival and reply even if I am the owner of the discussion and group (**NOT OK**)~~

4hardfire (approved):
Everything seems to work: post comments, reply, like, request deletion, request archival and it checks accesses properly.

PR #614, opened 2021-09-16
EICNET-962: As a GO I want to choose a URL alias for my group (rework)

4hardfire:
### Improvements

- Queue group content URL aliases to be updated when a group (with manual URL alias) has changed URL.

### Tests

- [ ] Create a discussion in the group
- [ ] Edit the group, enable manual URL alias and change the URL to a different one
- [ ] Save the group and check if you get redirected to the new URL when navigating to the old one
- [ ] Run cron
- [ ] Check if the discussion URL has been updated according to the new group URL and make sure you get redirected to this new URL when navigating to the old one


lib/modules/eic_groups/eic_groups.module
```diff
@@ -427,3 +427,27 @@ function eic_groups_entity_base_field_info(EntityTypeInterface $entity_type) {
...
+  }
+
+  if ($entity->isNew()) {
+    return;
+  }
+
+  // If group alias is set to manual and has changed, we add the group into a
+  // queue so that all group content url aliases can be updated at a later
+  // stage with cron.
+  if (!$entity->get('path')->pathauto) {
```
Dromadoros, line 446: we can just put in the same "if" statement

lramarojaona: @4hardfire , We don't need this check.
When you revert back to pathauto, it will also change the URL alias, so we need this code to run in any case (in case alias has changed)

lramarojaona (changes requested):

lib/modules/eic_groups/eic_groups.module
4hardfire: If the URL alias is set to manual it won't trigger pathauto alias alter hook and for that reason we need this condition otherwise group content aliases won't get updated.

lramarojaona (approved):

PR #619, opened 2021-09-17
EICNET-1378: Automatically subscribe (flag) user on certain user actions

4hardfire:
### New Features

- Implement logic to subscribe user to a group after becoming a member; 
- Implement partial logic to subscribe user to a topic of interest after user profile has been updated.

### Fixes

- Coding standards.

### Tests

- [x] Check if you get automatically subscribed (**Follow flag**) to the group
- [x] As TU, join the group and check if you also get automatically subscribed (**Follow flag**) to the group

### Todo

- [x] Implement missing logic to subscribe users to nodes and topics of interest. NOTE: this can only be done after implementing the missing Follow flag for Nodes and Taxonomy terms.

### Notes

- [x] This PR should be merged after #621


lib/modules/eic_flags/eic_flags.module
```diff
@@ -232,3 +231,31 @@ function eic_flags_entity_view(array &$build, EntityInterface $entity, EntityVie
...
+    case 'node':
+      \Drupal::classResolver(EntityOperations::class)
+        ->entityInsert($entity);
+      break;
+
+  }
+}
+
+/**
+ * Implements hook_entity_insert().
```
lramarojaona, line 251: Wrong description ;)

lib/modules/eic_flags/eic_flags.module
```diff
@@ -232,3 +231,31 @@ function eic_flags_entity_view(array &$build, EntityInterface $entity, EntityVie
...
+ */
+function eic_flags_entity_update(EntityInterface $entity) {
+  switch ($entity->getEntityTypeId()) {
+    case 'profile':
+      \Drupal::classResolver(EntityOperations::class)
+        ->entityInsert($entity);
```
lramarojaona, line 257: I see

```
function eic_flags_entity_update() {
  ->entityInsert($entity);
}
```

Is that intentional?

lib/modules/eic_flags/src/Hooks/EntityOperations.php
```diff
@@ -190,4 +234,121 @@ public function entityView(array &$build, EntityInterface $entity, EntityViewDis
...
+  public function followEntityOnCreation(EntityInterface $entity) {
+    $flag_entity = FALSE;
+    $flag_type = FALSE;
+
+    switch ($entity->getEntityTypeId()) {
+      case 'comment':
+        $flag_entity = $entity->getCommentedEntity();
+
+        // Array of node types that don't implement follow flag.
+        $excluded_node_types = [];
```
lramarojaona, line 277: Here it would be better to get this from the follow flag itself, so if make a change in the flag config, it's directly reflected here

lramarojaona (changes requested):

lib/modules/eic_flags/eic_flags.module
4hardfire: This has been fixed in the referenced PR :)

lib/modules/eic_flags/eic_flags.module
4hardfire: This has been fixed in the referenced PR :)

lib/modules/eic_flags/src/Hooks/EntityOperations.php
4hardfire: This has been fixed in the referenced PR :)

lramarojaona (approved):

PR #621, opened 2021-09-20
EICNET-1375: Create all the follow flags

4hardfire:
### New features

- Create new flag to follow content (follow_content);
- Automatically follow content after creating it;
- Create new flag to follow taxonomy terms (follow_taxonomy_term);
- Automatically follow topics of interest after user profile create/update.

### Improvements

- Create new class to define constants for EIC flags;
- Fix coding standards.

### Tests

**Follow content:**

- [x] As a TU, create a News/Story;
- [x] From the detail page, check if you are automatically following the News/Story;
- [x] Go to a News/Story page which you are not following;
- [x] Leave a comment, refresh the page and check if are now following that News/Story;

**Follow group:**

- [x] Join a group as a member and check if you are automatically following that group;

**Follow topics of interest:**

- [x] Edit your profile in `/user/<user-id>/member` and reference some topics of interest;
- [x] Go to one of the topics page `/taxonomy/term/<term-id>` that you referenced and check if you are automatically following that term.


lib/modules/eic_flags/src/FlagType.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Drupal\eic_flags;
+
+/**
+ * EIC flag types.
+ *
+ * @package Drupal\eic_flags
+ */
+final class FlagType {
```
lramarojaona, line 10: Since you are just defining follow flags, I would rename this class to _FollowFlagType_

lramarojaona: Unless this class is meant for toher flag types as well

lramarojaona (changes requested):

lramarojaona:
Code works, but 2 (unrelated) remarks:

- TU can't save a Story or News (workflow transition issue)
- TU can't edit his member profile

lib/modules/eic_flags/src/FlagType.php
4hardfire: I would leave it like this just in case we need to create other constants. At least I would prefer to have everything in one place. But if you think it's better to separate each flag types I can create a new class.

lib/modules/eic_flags/src/Hooks/EntityOperations.php
```diff
@@ -332,23 +339,50 @@ public function followTopicsOnUserProfileUpdate(ProfileInterface $profile) {
...
       $topics[$topic->id()] = $topic;
     }
 
-    $old_topics = $profile->original->get('field_vocab_topic_interest')->referencedEntities();
-    foreach ($old_topics as $topic) {
-      if (!isset($topics[$topic->id()])) {
-        // @todo Unflag topic.
-        continue;
-      }
+    // If node is not new then we need to unfollow old topics.
```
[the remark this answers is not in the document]
4hardfire: That was a mistake. Fixed!

4hardfire:
> Code works, but 2 (unrelated) remarks:
> 
> * TU can't save a Story or News (workflow transition issue)
> * TU can't edit his member profile

We should create a follow-up ticket.

lramarojaona (approved):

PR #626, opened 2021-09-21
EICNET-1355: Do not make avatar clickable for anonymous users

4hardfire:
### Improvements

- Add author profile URL in user teaser if the current user has access to view profile pages.

### Tests

- [ ] Go to the homepage and check if the authors avatars of the Latest News & Stories are not clickable for anonymous users

### Todo

- [ ] Fix storybook component `/patterns/components/author.html.twig` so that it does not print an empty `<a>` tag around the avatar when the path variable is empty. Currently it only checks if the name variable is presented but not the path.

### DIFF
--- a/lib/themes/eic_community/includes/preprocess/users/user.inc
+++ b/lib/themes/eic_community/includes/preprocess/users/user.inc
@@ -42,8 +42,12 @@ function _get_profile_image_array(EntityInterface $user): ?array {
 function eic_community_get_teaser_user_display(User $user) {
   $author = [
     'name' => $user->getDisplayName(),
-    'path' => $user->toUrl()->toString(),
   ];
+
+  // We add the user profile page URL if the current user has access to it.
+  if ($user->toUrl()->access()) {
+    $author['path'] = $user->toUrl()->toString();
+  }
 
   if (!$user->get('field_media')->isEmpty()) {
     /** @var \Drupal\media\Entity\Media $media */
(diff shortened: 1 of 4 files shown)


lib/themes/eic_community/includes/preprocess/users/user.inc
```diff
@@ -42,9 +42,13 @@ function _get_profile_image_array(EntityInterface $user): ?array {
 function eic_community_get_teaser_user_display(User $user) {
   $author = [
     'name' => $user->getDisplayName(),
-    'path' => $user->toUrl()->toString(),
   ];
 
+  // We add the user profile page URL if the current user has access to it.
+  if ($user->toUrl()->access()) {
```
xwero, line 48: The user that is passed to the function is not the current user most of the time, so this would not work.
This is the reason I use the `$variables['logged_in']` check in the hook function where the preprocess variables are added.  

Another reason why i don't add a check to display the path there is because the permissions could get more complicated than just being logged in or not. Then function parameters need to accommodate for every situation possible. 
Just transforming the User object and figure things out further down the line makes more sense to me. 
To make it more generic there could be a function or functions that checks for the permissions to display the author data, which could look something like this, `$variables['author'] = eic_community_author_path_check(eic_community_get_teaser_user_display($node_translation->getOwner()))`

xwero (changes requested):

lib/themes/eic_community/includes/preprocess/users/user.inc
4hardfire: The `$user->toUrl()->access()` will assume the current user (if you don't provide any parameter) and not the user that is passed as an argument in `eic_community_get_teaser_user_display`.

xwero (approved):
I tested the user link with and without a user image because I needed to change both in the author.html.twig file.

It is good for me.

PR #627, opened 2021-09-22
EICNET-1370: Switch full group header for previous page link on group video page

xwero:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1370?focusedCommentId=4950686&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4950686

### DIFF
--- a/lib/themes/eic_community/includes/preprocess/content/node--video.inc
+++ b/lib/themes/eic_community/includes/preprocess/content/node--video.inc
@@ -11,6 +11,14 @@
 function eic_community_preprocess_node__video(array &$variables) {
   switch ($variables['view_mode']) {
     case 'full':
+      $request = Drupal::request();
+
+      $variables['parent']['link'] = [
+        'label' => t('Previous page'),
+        'path' => $request->request->get('referer', 'javascript:history.go(-1)'),
+      ];
+      $variables['icon_file_path'] = $variables['eic_icon_path'];
+
       _eic_community_display_flags($variables);
       _eic_community_display_topics($variables);
       _eic_community_preprocess_video_contributors($variables);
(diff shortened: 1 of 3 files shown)


lib/themes/eic_community/includes/preprocess/content/node--video.inc
```diff
@@ -11,6 +11,14 @@
 function eic_community_preprocess_node__video(array &$variables) {
   switch ($variables['view_mode']) {
     case 'full':
+      $request = Drupal::request();
+
+      $variables['parent']['link'] = [
+        'label' => t('Previous page'),
+        'path' => $request->request->get('referer', 'javascript:history.go(-1)'),
```
Dromadoros, line 18: instead of instantiate the drupal request just to get the previous page, we can just put "path" as: 'javascript:history.go(-1)'

Behind, in the ecl-link, it will just put the value into an href:

```
<a
  href="{{ _link.path }}"
  class="{{ _css_class }}"
  {{ _extra_attributes|raw }}
>
```

So it should work like that

xwero: It is preferred to use an url instead of the javascript code for SEO, that is why I used the request. 
But in this project SEO is not the main goal, so the javascript link is ok.

filipvt1 (approved):

lib/themes/eic_community/includes/preprocess/content/node--video.inc
Dromadoros: Do we keep the history.go(-1) solution ? Even if it can returns to google for example ?

PR #651, opened 2021-09-23
EICNET-1204: Added video detail page with filelist.

xwero:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1204?focusedCommentId=4954178&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4954178

I had to create the full detail page too, the display of the slides is not done.


lramarojaona (approved):
@xwero ,
The download links were wrong.
We should use the `media/<id>/download` URL to track downloads.
I created a helper class/function to do that and refactored the preprocess functions to use this helper function.

PR #652, opened 2021-09-23
EICNET-1214: Overview empty state

speeters1:


xwero (commented):
Maybe add an example of the empty state and no navigation in storybook?

Dromadoros:
> Maybe add an example of the empty state and no navigation in storybook?

Yes I agree, would be nice

Askerx9:
is done @xwero

xwero (commented):
I see the added group empty story, but I don't see the empty state text?

Askerx9:
sorry small problem of understanding. I think it's ok now

filipvt1:
@Askerx9 can you handle resolving the conflicts on this branch? It looks to me like there are 2 implementations of the empty state at the moment. One uses the fields no_results and no_results_text while the other uses no_results_title, no_results_body.

To check which version you keep.

filipvt1 (changes requested):
Inside the file lib/themes/eic_community/react/components/Block/ActivityStream/index.js
There is still a reference to the old values: 
```
           <React.Fragment>
              <h3 className="ecl-activity-stream__no-content-header">{this.props.translations.no_results}</h3>
              <p className="ecl-teaser-overview__no-content-message">{this.props.translations.no_results_text}</p>
            </React.Fragment>
```

Askerx9:
Yeeeaaahhhh it's done @filipvt1

filipvt1 (approved):

PR #653, opened 2021-09-27
EICNET-1379: Send message upon user actions and subscribe users to it using the correct flags

4hardfire:
### New features

- Send subscription messages on comment creation (MT16, MT19);
- Send subscription messages on group content published/updated (MT18, MT20);
- Send subscription messages of Content published about Terms of Interest (MT23).

### Tests

- [x] As TU, follow the group
- [x] As GO/GA, add new discussion and check the option to "Send notification"
- [x] Check if the TU received an email notification (SUBSCRIPTION :: MT18)
- [x] As GO/GA, edit the discussion and check the option to "Send notification"
- [x] As TU, follow the discussion node
- [x] Check if the TU received an email notification (SUBSCRIPTION :: MT20)
- [x] As GO/GA, add a new comment in the discussion
- [x] Make sure the TU didn't receive any notification
- [x] As TU, follow the discussion node
- [x] As GO/GA, add a new comment in the discussion
- [x] Check if the TU received an email notification (SUBSCRIPTION :: MT19)
- [x] As GO/GA, add a new comment **reply** in the discussion
- [x] Check if the TU received an email notification (SUBSCRIPTION :: MT16)

### Todo

- Define the notifications workflow for every message subscription type.

### Needs discussion

- Currently it is using a custom cache to save the group content nodes that will need to trigger a subscription after the node has been created. We could probably use state API instead. **Update**: this has been tackled in #668.

### DIFF
--- a/lib/modules/eic_messages/modules/eic_message_subscriptions/src/MessageSubscriptionTypes.php
+++ b/lib/modules/eic_messages/modules/eic_message_subscriptions/src/MessageSubscriptionTypes.php
@@ -17,4 +17,8 @@ final class MessageSubscriptionTypes {
 
   const GROUP_CONTENT_UPDATED = 'sub_group_content_updated';
 
+  const NODE_PUBLISHED = 'sub_content_interest_published';
+
+  const NODE_UPDATED = NULL;
+
 }
--- a/lib/modules/eic_messages/modules/eic_message_subscriptions/src/MessageSubscriptionTypes.php
+++ b/lib/modules/eic_messages/modules/eic_message_subscriptions/src/MessageSubscriptionTypes.php
@@ -7,7 +7,7 @@ namespace Drupal\eic_message_subscriptions;
  *
  * @package Drupal\eic_message_subscriptions
  */
-final class MessageSubscriptionType {
+final class MessageSubscriptionTypes {
 
   const NEW_COMMENT_REPLY = 'sub_new_comment_reply';
 
--- a/config/sync/core.entity_form_display.message.sub_group_content_updated.default.yml
+++ b/config/sync/core.entity_form_display.message.sub_group_content_updated.default.yml
@@ -4,6 +4,7 @@ status: true
 dependencies:
   config:
     - field.field.message.sub_group_content_updated.field_event_executing_user
+    - field.field.message.sub_group_content_updated.field_referenced_node
     - message.template.sub_group_content_updated
 id: message.sub_group_content_updated.default
 targetEntityType: message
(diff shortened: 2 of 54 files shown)


lib/modules/eic_messages/modules/eic_message_subscriptions/src/MessageSubscriptionTypes.php
```diff
@@ -0,0 +1,24 @@
...
+
+  const NEW_COMMENT = 'sub_new_comment_on_content';
+
+  const NEW_GROUP_CONTENT_PUBLISHED = 'sub_new_group_content_published';
+
+  const GROUP_CONTENT_UPDATED = 'sub_group_content_updated';
+
+  const NODE_PUBLISHED = 'sub_content_interest_published';
+
+  const NODE_UPDATED = NULL;
```
lramarojaona, line 22: @4hardfire , should we add @todo here?

lramarojaona (changes requested):
@4hardfire , why don't we use the https://www.drupal.org/project/message_subscribe module?
It is completely part of the message stack, I think we should use it?

lramarojaona:
The following step doesn't work on my side

- [ ] Check if the TU received an email notification (SUBSCRIPTION :: MT20)

4hardfire:
> The following step doesn't work on my side
> 
> _Check if the TU received an email notification (SUBSCRIPTION :: MT20)_

@lramarojaona, I forgot to add a step: **As TU, follow the discussion node**. Can you re-test again?

- [ ] As GO/GA, edit the discussion and check the option to "Send notification"
- [ ] As TU, follow the discussion node
- [ ] Check if the TU received an email notification (SUBSCRIPTION :: MT20)

lramarojaona:
> 
> 
> > The following step doesn't work on my side
> > _Check if the TU received an email notification (SUBSCRIPTION :: MT20)_
> 
> @lramarojaona, I forgot to add a step: **As TU, follow the discussion node**. Can you re-test again?
> 
>     * [ ]  As GO/GA, edit the discussion and check the option to "Send notification"
> 
>     * [ ]  As TU, follow the discussion node
> 
>     * [ ]  Check if the TU received an email notification (SUBSCRIPTION :: MT20)

Yes, then it works

lramarojaona (approved):

4hardfire:
> @4hardfire , why don't we use the https://www.drupal.org/project/message_subscribe module? It is completely part of the message stack, I think we should use it?

@lramarojaona, The message_subscribe module uses their own flags provided by the module and in this case we have our custom ones. Also, it is not compatible with groups (only Organic groups) which would need some re-work on that part. Maybe it will still worth to have a look for the PART 2.

lib/modules/eic_messages/modules/eic_message_subscriptions/src/MessageSubscriptionTypes.php
4hardfire: the NODE_UPDATED is not used for now and we can remove it. This can be done in the PART 2 #670

PR #654, opened 2021-09-27
Feature/eicnet 1199 add slider to gallery detail page

xwero:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1199?focusedCommentId=4960143&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4960143


filipvt1 (changes requested):
- The gallery doesn't handle images with different aspect ratios well. If you upload 1 file with a 16:9 ratio and a file with a 9:16 ratio you get to see all the tiles in the 9:16 ratio.
- The original images are used in the carousel. This is not ideal. If the user uploads some high-res images this will increase the page size by a great margin
- I think an image style should be added that downsizes the thumbnails and crops them to the correct ratio for the gallery

PR #655, opened 2021-09-27
Feature/eicnet 1331 last flagged items

lramarojaona:
### Tests

**Highlight**

- [x] Highlight an item
- [x] Select the "Last highlighted" sort option on the overview
- [x] The item you last highlighted should appear first

**Bookmark**

- [x] Bookmark an item
- [x] Select the "Last bookmarked" sort option on the overview
- [x] The item you last bookmarked should appear first

Remark:

- Bookmark flag might not be available for now, this needs to be added in separate ticket
- For now, you will see "undefined" in the sort options dropdown because this sort doesn't implement an ASC direction, only DESC. This will be fixed in React later.


4hardfire (changes requested):
There is a missing dependency in eic_search module.
`The service "eic_search.solr_document_processor" has a dependency on a non-existent service "eic_statistics.statistics.storage.node.decorator".`. We just need to add `eic_statistics:eic_statistics` module in `eic_search.info.yml`.

4hardfire (changes requested):
The bookmark flag is not visibile in the detail page of a node because of the selected "Link type" in flag administration `/admin/structure/flags/manage/bookmark_content`. We should select the option "EIC Count link" as Link type.

Dromadoros (approved):

PR #659, opened 2021-09-28
EICNET-1347: Removed contributors / fixed overlap with footer.

filipvt1:
- Removed contributors 
- Fixed overlap issuewith footer  when too little content is present


Askerx9 (approved):
it's ok in storybook

PR #660, opened 2021-09-28
[EICNET-1205]feat: Adapt also comments for gallery and video

Dromadoros:
How to test:

- [x] Create a gallery or video in a group
- [x] Go to gallery and video detail
- [x] Check if you have comments block and try to comment it


lramarojaona:
@Dromadoros ,

- I see 2 comment blocks one Drupal and one React
- When I comment on the react one, then the block disappears with JS errors
- Then I reload the page, the block is there again with my comment
- If I click on reply, I4m being redirected to /comment/reply/node/22/field_comments/2

Maybe I'm missing something?

PR #664, opened 2021-09-29
Feat: scroll to top when page change

Askerx9:


Dromadoros:
Très gros changement, gg

lib/themes/eic_community/react/components/Block/Overview/index.js
```diff
@@ -42,6 +42,7 @@ class Overview extends React.Component {
   }
 
   changePage(page) {
+    window.scrollTo(0,0)
```
Dromadoros, line 45: For now it's working because we have no banner, etc. But for example if you have a big banner, that takes half page and you go always on the top page, it's not really great.

Maybe just a create a ref in the react component and put it on the first html element. Then this.myRef.current.scrollIntoView()

Askerx9: ha yes you right so it's done now

PR #674, opened 2021-10-01
EICNET-1283: As a GM I can edit every wiki page from my groups

4hardfire:
### Improvements

- Create static method in `EICGroupsHelper `to check if a user is a GO/GA of a given group.

### Tests

- [ ] As GO, create a new group
- [ ] As SA/SCM, publish the group
- [ ] As TU, join the group
- [ ] As TU, create a new wiki page and **uncheck** the option "Editable by members"
- [ ] As GO/GA, make sure you can edit the wiki page and be able change the option "Editable by members"
- [ ] As GO/GA, create another wiki page and **check** the option "Editable by members"
- [ ] As TU, make sure you can edit the wiki page but you **cannot** change the option "Editable by members" (the option should not be visible at all)
- [ ] As GO/GA, edit the wiki page and **uncheck** the option "Editable by members"
- [ ] As TU, make sure you **cannot** edit the wiki page anymore

### DIFF
--- a/lib/modules/eic_groups/src/EICGroupsHelper.php
+++ b/lib/modules/eic_groups/src/EICGroupsHelper.php
@@ -9,6 +9,7 @@ use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityStorageException;
 use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Session\AccountProxyInterface;
 use Drupal\Core\StringTranslation\StringTranslationTrait;
 use Drupal\eic_groups\Constants\GroupJoiningMethodType;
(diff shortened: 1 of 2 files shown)


lib/modules/eic_groups/src/EICGroupsHelper.php
```diff
@@ -386,4 +387,39 @@ public static function groupIsFlaggable(GroupInterface $group) {
...
+   * @return bool
+   *   TRUE if user is a group admin.
+   */
+  public static function userIsGroupAdmin(GroupInterface $group, AccountInterface $account) {
+    $membership = $group->getMember($account);
+    $membership_roles = $membership->getRoles();
+    $is_admin = FALSE;
+
+    foreach ($membership_roles as $role) {
+      $is_admin = in_array(
```
Dromadoros, line 407: Isn't it cleaner to do something like this ?

`$membership_roles = array_filter($membership_roles, function(GroupMembership $membership_role) {
   return in_array($membership_role->id(), [
          self::GROUP_ADMINISTRATOR_ROLE,
          self::GROUP_OWNER_ROLE,
        ]);
}

return !empty($membership_roles);
`

Well in fact in your case, what's nice is that we will not foreach all membership roles if first one is ok (but at least we will have max an array of 4 elements...)

4hardfire: Indeed... I exit the loop once I find one the admin roles. For me, we can use my solution but if you think your changes improves code readability I can do the changes.

lib/modules/eic_groups/src/Hooks/EntityOperations.php
```diff
@@ -432,13 +433,27 @@ public function entityFieldAccess($operation, FieldDefinitionInterface $field_de
             return AccessResult::forbidden();
           }
 
-          // We return access denied if the user is not the owner of the wiki
-          // page neither administrator.
-          if ($entity->getOwnerId() !== $account->id()) {
-            if (!UserHelper::isPowerUser($account)) {
-              $access = AccessResult::forbidden();
-            }
+          // If user is a group admin, we allow access.
```
lramarojaona, line 436: Description seems wrong here

lramarojaona (changes requested):

lib/modules/eic_groups/src/Hooks/EntityOperations.php
4hardfire: Fixed!

PR #679, opened 2021-10-01
[EICNET-745]feature: Do not force delete, use soft delete and replace by message

Dromadoros:
How to test:

- [ ] Go to a discussion detail
- [ ] Add multiple comments
- [ ] Try to remove one, you should have a message instead of the comment

PS: for mvp we don't force remove, always show messages at any level


lib/modules/eic_groups/src/Controller/DiscussionController.php
```diff
@@ -305,26 +319,50 @@ public function editComment(Request $request, int $discussion_id, $comment_id) {
...
-    if ($user->id() !== $comment->getOwnerId()) {
+    $user_group_roles = array_merge(
+      $user_group_roles,
+      $this->currentUser()->getRoles(TRUE)
+    );
+
+    if (
+      !in_array(UserHelper::ROLE_SITE_ADMINISTRATOR, $user_group_roles) &&
+      !in_array(UserHelper::ROLE_CONTENT_ADMINISTRATOR, $user_group_roles) &&
+      !in_array(UserHelper::ROLE_DRUPAL_ADMINISTRATOR, $user_group_roles)
```
lramarojaona, line 354: Here it would be better to use `UserHelper::isPowerUser()` instead of harcoding the roles.

Even better, why not using a custom access checker for this route instead of using 'access content' permission?

```yml
  requirements:
    _custom_access: '\Drupal\example\Controller\ExampleController::access'
```

lib/modules/eic_groups/src/Plugin/Block/CommentsFromDiscussionBlock.php
```diff
@@ -172,11 +181,9 @@ public function build() {
...
           $group_contents
         )->isAllowed(),
-        'delete_all_comments' => $this->groupPermissionChecker->getPermissionInGroups(
-          'delete any page content',
-          $current_user,
-          $group_contents
-        )->isAllowed(),
+        'delete_all_comments' => in_array(UserHelper::ROLE_SITE_ADMINISTRATOR, $user_group_roles) ||
+          in_array(UserHelper::ROLE_CONTENT_ADMINISTRATOR, $user_group_roles) ||
+          in_array(UserHelper::ROLE_DRUPAL_ADMINISTRATOR, $user_group_roles),
```
lramarojaona, line 186: Same here, please use `UserHelper::isPowerUser()`

lib/modules/eic_groups/eic_groups.routing.yml
```diff
@@ -38,12 +38,13 @@ eic_groups.discussion_update_comment:
     _permission: 'access content'
 
 eic_groups.discussion_delete_comment:
-  path: '/api/discussion/{discussion_id}/delete/{comment_id}'
+  path: '/api/discussion/{group_id}/{discussion_id}/delete/{comment_id}'
   methods: [ DELETE ]
   defaults:
     _controller: '\Drupal\eic_groups\Controller\DiscussionController::deleteComment'
   requirements:
     _permission: 'access content'
```
lramarojaona, line 46: This should be removed if we have a custom access check:
`_permission: 'access content'`

lramarojaona (approved):

PR #681, opened 2021-10-04
[EICNET-1406] chore: switch to case insensitive search

korhankale:
This PR introduces a case insensitive search using entity titles
We now define the tm_* Solr dynamic field's type as 'text_lower_case' which has different filters & tokenizer properties to make the search case insensitive

## To Test

- [ ] Update your local & re-index
- [ ] Go to /solr/#/eic/schema?field=tm_global_title and make sure you have the field definition with type = 'text_lower_case'
- [ ] Go to /groups (or any other search block)
- [ ] Make uppercase and lowercase searches (e.g: GROUP, group, GrOUp, etc.)
- [ ] You should have matching results

### DIFF
--- a/config/sync/context.context.group_discussions_overviews.yml
+++ b/config/sync/context.context.group_discussions_overviews.yml
@@ -50,7 +50,7 @@ reactions:
         sort_options:
           its_flag_highlight_content: its_flag_highlight_content
           timestamp: timestamp
-          ss_global_title: ss_global_title
+          tm_global_title: tm_global_title
           its_last_flagged_bookmark_content: its_last_flagged_bookmark_content
           its_last_flagged_highlight_content: its_last_flagged_highlight_content
           dm_aggregated_changed: dm_aggregated_changed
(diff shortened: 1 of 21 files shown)


4hardfire (commented):
Looks good to me! However, if one of the searched terms are in the results it will always return results even if it does not contain the other terms. I don't know if this is the normal behaviour. @Dromadoros can you have a look?

korhankale:
> Looks good to me! However, if one of the searched terms are in the results it will always return results even if it does not contain the other terms. I don't know if this is the normal behaviour. @Dromadoros can you have a look?

Seems like in the D7 the search phrase is tokenized into words ("Group hello Brussels" -> "Group", "Hello", "Brussels"), any document matching to at least one word will be returned. I guess we should keep it
Just in case, if we want to change that behaviour and keep the search phrase as whole, we should change 
`<tokenizer class="solr.StandardTokenizerFactory"/>` to `<tokenizer class="solr.KeywordTokenizerFactory"/>` in field type `text_lower_case`

PR #688, opened 2021-10-06
EICNET-1511: Group overview in Drupal admin

4hardfire:
### Features

- Create groups administration view for SA/SCM.

### Tests

- [ ] Go to `/admin/community/groups`
- [ ] Make sure you can filter groups by: ID, group name, status and owner


korhankale:
> > image Seems like there is an issue with the owner column
> 
> @korhankale can you check if the owner (I think in your case might be the drupal admin) has a first and last name?

Indeed works better now

korhankale (approved):

PR #692, opened 2021-10-08
EICNET-1493: Groups - Group owner in group teaser links to groups overview

4hardfire:
### Fixes

- Fix group owner url in groups overview.

### Tests

- [x] Re-index solr
- [x] As TU, go to the groups overview `/groups`
- [x] Click in the avatar or name of a group owner and make sure you get redirected to the user detail page


4hardfire:
> In your description it is mentioned that the avatar should be clickable but it doesn't seems to be. Should it?

That was not part of the ticket but indeed... I mentioned in the tests. In any case I changed the react template and the avatar should now be a link.

PR #694, opened 2021-10-11
EICNET-1490: As TU, profile avatars are not shown

4hardfire:
### Improvements

- Allow users with permission "access user profiles" to view other user avatars.

### Tests

- [ ] As TU, go to groups overview `/groups` and make sure you can view user avatars for each group

### DIFF
--- a/lib/modules/eic_media/eic_media.module
+++ b/lib/modules/eic_media/eic_media.module
@@ -43,11 +43,9 @@ function eic_media_media_access(EntityInterface $entity, $operation, AccountInte
       $entity = \Drupal::entityTypeManager()->getStorage($entity_type)->load($entity_id);
 
       // Allow access to user pictures if user has permission.
-      if ($entity_type === 'user') {
-        if ($account->hasPermission('access user profiles')) {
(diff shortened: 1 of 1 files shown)


lib/modules/eic_media/eic_media.module
```diff
@@ -41,6 +41,15 @@ function eic_media_media_access(EntityInterface $entity, $operation, AccountInte
   foreach (array_keys($media_usage) as $entity_type) {
     foreach (array_keys($media_usage[$entity_type]) as $entity_id) {
       $entity = \Drupal::entityTypeManager()->getStorage($entity_type)->load($entity_id);
+
+      // Allow access to user pictures if user has permission.
+      if ($entity_type === 'user') {
```
Dromadoros, line 46: why not just 

`if ($entity_type === 'user' &&  $account->hasPermission('access user profiles'))`

instead of 2 if

4hardfire: Because my brain stopped working for a sec 😄. It's done!

PR #696, opened 2021-10-11
Feature/eicnet 1255 - activity stream fixe size and padding

Askerx9:


filipvt1 (changes requested):
The padding on the ".ecl-activity-stream__item" item is too big now. It does not match the design. The increased padding also makes the block a lot higher. I would revert back to the original padding and add a right padding (not a margin because that would mess with the alignment of the border) on ".ecl-activity-stream__item__main" to push it away from the closing icon.

PR #698, opened 2021-10-11
[EICNET-1083]fix: Set correct permission check for activitystream deletion

Dromadoros:
How to test it:

- [ ] Go to a group detail and click to latest activity 
- [ ] Add a content via a group (for eg: discussion)
- [ ] Return to the activity stream overview
- [ ] Check if you can delete an activity with these roles: SA, SCM, GO and GA


lib/modules/eic_groups/src/Plugin/Block/LastGroupActivitiesBlock.php
```diff
@@ -159,16 +160,27 @@ public function build() {
     }, $members);
 
     $current_group_route = $this->groupsHelper->getGroupFromRoute();
+    $account = \Drupal::currentUser();
```
korhankale, line 163: might be better to use dependency injection instead of calling Drupal::

PR #701, opened 2021-10-12
EICNET-1256 | Fix: add border on button cancel in activity stream modal

Askerx9:


filipvt1:
Overwriting the ghost-button styling isn't the right way I think. The ghost style is used for different items as well (for example the close link). Those now also receive a border. There is a variant called "ecl-link--button-secondary" which does have a border.

PR #702, opened 2021-10-12
[EICNET-1474]feature: adapt tree widget for users

Dromadoros:
How to test:

- [x] Edit your accound
- [x] For now, you should have the widget for the field "deleted by"
- [x] Test searching with first name, last name and email
- [x] try to check, uncheck values and save


4hardfire:
> @4hardfire I added a new config to the widget to ignore or not the current user

@Dromadoros, I created a user "Bernardo Paulino (bernardopaulino@blue4you.be)" and if I search for "Bernardo Paulino" I get no results. But if I search for the first name or last name only, then it works. I am also still able to select the current user if I search for it (not using the dropdown).

4hardfire (changes requested):
@Dromadoros, I created a user "Bernardo Paulino (bernardopaulino@blue4you.be)" and if I search for "Bernardo Paulino" I get no results. But if I search for the first name or last name only, then it works. I am also still able to select the current user if I search for it (not using the dropdown).
Also, I still see a required field message when saving the "Is deleted by" as empty. The user account gets saved anyway.

PR #709, opened 2021-10-12
EICNET-1369 Fix: Add margin after comment overview above an video

Askerx9:


filipvt1 (changes requested):
I would remove the margin-bottom and display inline-block declaration and turn the comment overview title into an h3, at the moment this is a span and that is not semantically correct. By turning this into an h3 it automatically becomes a block item and it gets a bottom margin

PR #719, opened 2021-10-14
EICNET-1520: Groups - TU does not have access to comment settings, url alias

4hardfire:
### Improvements

- Allow users to manage comment settings of their own nodes.

### Tests

- [x] As TU, create a content in a group and make sure you can manage comment settings (Open or close comments)
- [x] As GO/GA, edit a content which you're are not the owner and make sure you can manage comment settings.

### Todo:

- [ ] Enable/Disable comments in the frontend based on the comment settings

### DIFF
--- a/lib/modules/eic_comments/eic_comments.module
+++ b/lib/modules/eic_comments/eic_comments.module
@@ -82,11 +82,9 @@ function eic_comments_form_alter(&$form, FormStateInterface $form_state, $form_i
     $groupType = $group->getGroupType();
 
     $installedContentPlugins = $groupType->getInstalledContentPlugins()->getInstanceIds();
-    if (!in_array("group_node:{$entity->bundle()}", $installedContentPlugins)) {
-      return;
+    if (in_array("group_node:{$entity->bundle()}", $installedContentPlugins)) {
+      $can_edit_comment_settings = EICGroupsHelper::userIsGroupAdmin($group, $current_user);
     }
-
-    $can_edit_comment_settings = EICGroupsHelper::userIsGroupAdmin($group, $current_user);
   }
 
   /** @var \Drupal\Core\Field\FieldDefinitionInterface[] $field_definitions */


lib/modules/eic_comments/eic_comments.module
```diff
@@ -49,3 +51,54 @@ function eic_comments_entity_view(array &$build, EntityInterface $entity, Entity
 
   }
 }
+
+/**
+ * Implements hook_entity_view().
```
lramarojaona, line 56: Typo here

lib/modules/eic_groups/src/Hooks/FormOperations.php
```diff
@@ -263,4 +304,76 @@ protected function enableDefaultFeatures(Group $group) {
...
+   * Alters comment settings field during form_alter phase.
+   *
+   * @param array $form
+   *   An associative array containing the structure of the form.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
+   * @param string $form_id
+   *   The form id.
+   */
+  public function commentSettingsFormFieldAlter(&$form, FormStateInterface $form_state, $form_id) {
```
lramarojaona, line 317: I don't get why we have the pretty same logic in 2 places?
`eic_groups_form_alter()` and `eic_comments_form_alter()` ?

4hardfire: @lramarojaona,
I wanted to make the eic_comments as much generic as possible without taking into account the group context. Since we have logic specific for the comments in group content I thought this would be better to do it in eic_groups module.
To be honest, the only logic that is a bit redundant is this:
```php
<?php

$form_object = $form_state->getFormObject();

// If we are not in a node form, we do nothing.
if (!$form_object instanceof NodeForm) {
  return;
}

$entity = $form_object->getEntity();

if (!$entity instanceof NodeInterface) {
  return;
}

// If entity is new, we do nothing.
if ($entity->isNew()) {
  return;
 }
```

Do you think we should do everything in the eic_comments module?

lib/modules/eic_comments/eic_comments.module
4hardfire: Fixed.

lib/modules/eic_groups/src/Hooks/FormOperations.php
lramarojaona: Personnaly I would group all this stuff in eic_comments indeed

lib/modules/eic_groups/src/Hooks/FormOperations.php
```diff
@@ -263,4 +304,76 @@ protected function enableDefaultFeatures(Group $group) {
...
+    $groupType = $group->getGroupType();
+
+    $installedContentPlugins = $groupType->getInstalledContentPlugins()->getInstanceIds();
+    if (!in_array("group_node:{$entity->bundle()}", $installedContentPlugins)) {
+      return;
+    }
+
+    $is_admin = EICGroupsHelper::userIsGroupAdmin($group, $this->currentUser->getAccount());
+
+    if (!$is_admin) {
```
lramarojaona, line 353: Are the comment settings allowed for GO/GA only?

lib/modules/eic_groups/src/Hooks/FormOperations.php
4hardfire: Done!

lib/modules/eic_comments/eic_comments.module
```diff
@@ -69,6 +69,33 @@ function eic_comments_form_alter(&$form, FormStateInterface $form_state, $form_i
...
 
+  $current_user = \Drupal::currentUser();
+
+  $is_admin = FALSE;
+
+  $moduleHandler = \Drupal::service('module_handler');
+
+  // If eic_group module is enabled, we need to make sure the user can manage
+  // comment settings.
+  if ($moduleHandler->moduleExists('eic_group')) {
```
lramarojaona, line 80: I believe the correct module name is `eic_groups`

lib/modules/eic_comments/eic_comments.module
```diff
@@ -69,6 +69,33 @@ function eic_comments_form_alter(&$form, FormStateInterface $form_state, $form_i
...
+    $group_content = reset($group_contents);
+    $group = $group_content->getGroup();
+    $groupType = $group->getGroupType();
+
+    $installedContentPlugins = $groupType->getInstalledContentPlugins()->getInstanceIds();
+    if (!in_array("group_node:{$entity->bundle()}", $installedContentPlugins)) {
+      return;
+    }
+
+    $is_admin = \Drupal::service('eic_groups.helper')->userIsGroupAdmin($group, $current_user);
```
lramarojaona, line 96: Why not `EICGroupsHelper::userIsGroupAdmin($group, $current_user)` as this is a static method?

So no need to instantiate the service.

lib/modules/eic_groups/src/Hooks/FormOperations.php
lramarojaona: Ok

lib/modules/eic_comments/eic_comments.module
4hardfire: If we do it that way, we have to make usage of the namespace `Drupal\eic_groups\EICGroupsHelper` and therefore we are creating dependency on the eic_groups module. That's the reason why I did it this way.

PR #747, opened 2021-10-21
[EICNET]fix: Fix community path on different env

Dromadoros:
How to test:

- [ ] Go to a discussion detail and be sure it's working

### DIFF
--- a/lib/themes/eic_community/react/components/Block/CommentsDiscussion/Services/UrlHelper.js
+++ b/lib/themes/eic_community/react/components/Block/CommentsDiscussion/Services/UrlHelper.js
@@ -1,0 +1,9 @@
+function getPrefixDiscussionEndpoint() {
+  if (window.location.pathname.indexOf("/community/") === 0) {
+    return '/community/api/discussion';
+  }
+
+  return '/api/discussion';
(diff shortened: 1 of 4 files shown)


lib/themes/eic_community/react/components/Block/CommentsDiscussion/Services/UrlHelper.js
```diff
@@ -0,0 +1,9 @@
+function getPrefixDiscussionEndpoint() {
+  if (window.location.pathname.indexOf("/community/") === 0) {
```
lramarojaona, line 2: Can you access Drupal settings?

In this case you should use `drupalSettings.path.baseUrl`

lramarojaona (changes requested):

lib/themes/eic_community/react/components/Block/CommentsDiscussion/Services/UrlHelper.js
Dromadoros: Yeah forgott that one with the deploy rush... done, thx

lramarojaona (approved):

PR #766, opened 2021-10-26
Feature/forms | make a correct layout for add group form

Askerx9:
### How to test
As a CM, go to http://localhost:8080/group/add/group

---------

### What I do

- [x] add form components in storybook
- [x] add form layout in storybook
- [x] make a better layout for add group form


filipvt1 (changes requested):
- The form doesn't have any margin which causes the submit button to stick to the footer. In storybook it's OK, but that is because in Storybook there is a section-wrapper around the form.
- The styling of a fieldset seems off. Now it just looks like the same label is repeated twice. I think the legend of the fieldset should be bigger and have some kind of separator to highlight that it is a separate section. But maybe this should be designed first?
fieldset

Askerx9:
I work on this ticket: https://citnet.tech.ec.europa.eu/CITnet/jira/secure/RapidBoard.jspa?rapidView=6628&view=detail&selectedIssue=EICNET-1586&quickFilter=26820&quickFilter=26822

topic selection need an improvement so it's done

Dromadoros (approved):

PR #771, opened 2021-10-26
Feature/eicnet 1534 group event overview

lramarojaona:
### Tests

- [ ] Create an event inside a group
- [ ] Go to `group/<group-id>/events`
- [ ] You should see the event with 3 facets: Type, Topic and Region/Country


lib/modules/eic_search/src/Search/Sources/GroupEventSourceType.php
```diff
@@ -0,0 +1,147 @@
...
+        'label' => $this->t('Highlight', [], ['context' => 'eic_search']),
+        'DESC' => $this->t('Highlighted first', [], ['context' => 'eic_search']),
+        'ASC' => $this->t('Highlighted last', [], ['context' => 'eic_search']),
+      ],
+      'ss_drupal_timestamp' => [
+        'label' => $this->t('Timestamp', [], ['context' => 'eic_search']),
+        'ASC' => $this->t('Old', [], ['context' => 'eic_search']),
+        'DESC' => $this->t('Recent', [], ['context' => 'eic_search']),
+      ],
+      'tm_global_title' => [
```
Dromadoros, line 65: sort will not work on tm

lramarojaona: This is fixed now

Dromadoros (approved):

PR #775, opened 2021-10-27
FIX: layout for group members overview

Askerx9:
### How to test

go to group member overview and check you have a correct layout like 
http://localhost:9000/?path=/story/bundles-group--group-member-private

- [ ] Groups members overview is ok on Storybook
- [ ] Groups members overview is ok on Localhost


filipvt1 (changes requested):
On localhost the "ecl-teaser-overview--has-3-columns" isn't set (at least on my localhost it isn't). To check.

filipvt1 (approved):

PR #780, opened 2021-10-28
EICNET-1581: Issue when generating group content URLs

4hardfire:
### Fixes

- Fix issue with duplicated base URLs when pathauto generates a new URL containing the "node_group_url" token.

### Improvements
- Create new custom token "node_book_parent_url" to grab the parent book URL of a node without base url. This will also fix the duplicated base URLs when pathauto generates a new URL for the wiki pages.

### Tests

- [ ] Create a new content in the group and check if the URL is `/groups/<group-title>/<group-content-overview-page>/<group-content-title>`
- [ ] Create a new wiki page in the group and check if the URL is `/groups/<group-title>/wiki/<wiki-page-title>`

**Notes:** when testing this in QA, make sure the URLs are prefixed with `/community` and not `/community/community`


4hardfire:
> Don't the other nodes than book/wiki pages need this?
> 
> Also, it would good to try that patch instead of having custom code (hopefully this patch will be included in a future Group release): https://www.drupal.org/project/group/issues/2774827#comment-13905685

Previously we were using `[node:book:parent:url:relative]` but that didn't take into account having a base URL like we have in QA (`/community`). So I created a custom token that fixes that.

Concerning the group token, the patch couldn't be applied and after looking at the code it doesn't seem the issue will fixed anyway. So this is why I kept the custom token `node_group_url` and fixed the issue with base URL for this one as well.

lib/modules/eic_content/modules/eic_content_book/src/Hooks/BookTokens.php
```diff
@@ -0,0 +1,124 @@
...
+                 *
+                 * @var \Drupal\node\NodeInterface $parent_book_node
+                 */
+                $parent_book_node = $this->entityTypeManager->getStorage('node')
+                  ->load($node->book['pid']);
+                $parent_book_node_url = $parent_book_node->toUrl()->toString();
+
+                // If base path is presented in the URL, we need to remove it
+                // in order to avoid duplicated base paths.
+                if (substr($parent_book_node_url, 0, strlen($base_url)) === $base_url) {
```
Dromadoros, line 109:
```
$has_base_path = substr($parent_book_node_url, 0, strlen($base_url)) === $base_url;
$replacements[$original] = $has_base_path ? substr_replace($parent_book_node_url, '', 0, strlen($base_url)) : $parent_book_node_url;
```

WDYT?

lib/modules/eic_groups/src/Hooks/GroupTokens.php
```diff
@@ -115,14 +133,22 @@ public function tokens($type, $tokens, array $data, array $options, BubbleableMe
                     $group = $group_content->getGroup();
                   }
 
-                  $replacements[$original] = $group->toUrl()->toString();
+                  $group_url = $group->toUrl()->toString();
+
+                  // If base path is presented in group URL, we need to remove
+                  // it in order to avoid duplicated base paths.
+                  if (substr($group_url, 0, strlen($base_url)) === $base_url) {
```
Dromadoros, line 140: same for ternary maybe ?

lib/modules/eic_content/modules/eic_content_book/src/Hooks/BookTokens.php
```diff
@@ -0,0 +1,124 @@
...
+  public function tokens($type, $tokens, array $data, array $options, BubbleableMetadata $bubbleable_metadata) {
+    $replacements = [];
+    $base_url = $this->requestStack->getCurrentRequest()->getBaseUrl();
+
+    // Node tokens.
+    if ($type === 'node' && !empty($data['node'])) {
+      foreach ($tokens as $name => $original) {
+        switch ($name) {
+          case 'node_book_parent_url':
+            if (isset($data['node'])) {
```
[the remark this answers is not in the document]
4hardfire: Code improved! Looks much better now.

lib/modules/eic_content/modules/eic_content_book/src/Hooks/BookTokens.php
4hardfire: I agree!

lib/modules/eic_groups/src/Hooks/GroupTokens.php
4hardfire: Done!

lib/modules/eic_content/modules/eic_content_book/src/Hooks/BookTokens.php
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * Implements hook_tokens().
+   */
+  public function tokens($type, $tokens, array $data, array $options, BubbleableMetadata $bubbleable_metadata) {
+    $replacements = [];
+    $base_url = $this->requestStack->getCurrentRequest()->getBaseUrl();
+
+    // Node tokens.
+    if ($type === 'node' && !empty($data['node'])) {
```
Dromadoros, line 84: I do like to avoid at maximum condition/loop wrapper. What we can do is just 

```
if ($type === 'node' && !empty($data['node'])) {
  return $replacements;
}
```

Dromadoros:
Added a small feedback otherwise it's working pretty good

lib/modules/eic_content/modules/eic_content_book/src/Hooks/BookTokens.php
4hardfire: Fixed!

Dromadoros (approved):

PR #783, opened 2021-10-28
Feature/eicnet 1532 group content event detailed page

Askerx9:
### How to test

- Go to storybook bundles/group/group event detailed
- check if it's ok with the [figma](https://www.figma.com/file/WZH3qhNynnmm0H6FKmD3hi/EASME_v1?node-id=9118%3A40870)


filipvt1 (changes requested):
- The design of the signup block looks different, to check what has changed.
- I would add a wrapper div around the h3 and ul which contains the event details. It is cleaner to have a single div for that component. All other items in the sidebar also have that wrapper.
- Links in event detail are black and have an underline in the design. In the design Facebook and twitter page doesn't have an underline, but because they are links I would also put an underline there

Askerx9:
@filipvt1 it's done :)

filipvt1 (changes requested):
It's best to apply a unique class to this div (for example "ecl-infos", to make it more in line with the item below it that has the class "ecl-additional-infos".

Askerx9:
Right @filipvt1

filipvt1 (approved):

PR #785, opened 2021-10-29
EICNET-1223: Also search on gallery media names.

lramarojaona:
### Tests

- [ ] Create a gallery
- [ ] Search for media name (by default it's the filename) in group's Files overview
- [ ] Search for media name in global search


Dromadoros (approved):
cool

PR #787, opened 2021-10-29
EICNET-1601: Increment event files in group file statistics

4hardfire:
### Features

- Increment/decrement group file statistics when uploading documents in CT Event.

### Tests

- [ ] As SA/SCM, go to a group and create a new Event
- [ ] Make sure you upload new documents in the Event (do not use existing ones)
- [ ] Save the Event
- [ ] Go to the group detail page and make sure the file statistics are updated
- [ ] Try to remove the documents from the event and also make sure the group file statistics are updated

### DIFF
--- a/lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.module
+++ b/lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.module
@@ -26,6 +26,7 @@ function eic_group_statistics_group_content_insert(EntityInterface $entity) {
     case 'group-group_membership':
     case 'group-group_node-discussion':
     case 'group-group_node-document':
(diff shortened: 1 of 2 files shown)


Dromadoros:
It's working but can we use a better widget inside event content type ? Actually you can only link existing media, not uploading one

lramarojaona:
> 
> 
> It's working but can we use a better widget inside event content type ? Actually you can only link existing media, not uploading one

The widget allows you to create new ones:
 _Create new media
Create your media on the media add page (opens a new window), then add it by name to the field below._ 

But indeed, there's still work to do for Forms UX, this will be another track, not in the scope of this ticket

Dromadoros:
ok so should be good for me

Dromadoros (approved):

PR #790, opened 2021-11-02
[EICNET-1609]feat: implement contact user form + allow config by profile + implment into overview link

Dromadoros:
How to test it:

- [x] Be sure you have some user's profiles that have enabled _"Members are allowed to contact you through your personal contact form"_
- [x] Go to /people
- [x] Users that doesn't have mail icon mean the field is not enabled
- [x] Try to contact someone by clicking on the icon
- [x] Fill the form check if you see the mail sent
- [x] Try and check the "send me a copy" checkbox and verify you see 2 emails, the original and the one prefix with "self copy: ..."
- [x] Try to hardcode url with user that does not allow contact and check that form is correctly hidden

ps: reindex items in search api + compile react


config/sync/core.entity_form_display.profile.member.default.yml
```diff
@@ -6,6 +6,7 @@ dependencies:
     - field.field.profile.member.field_body
     - field.field.profile.member.field_cv_link
     - field.field.profile.member.field_location_address
+    - field.field.profile.member.field_profile_can_contact
```
lramarojaona, line 9: I would definitely move this field to the user object and not the profile

lib/modules/eic_private_message/eic_private_message.routing.yml
```diff
@@ -0,0 +1,7 @@
+eic_private_message.user_private_message:
+  path: '/user/{user}/contact'
+  defaults:
+    _title: 'User contact form'
+    _form: '\Drupal\eic_private_message\Form\PrivateMessageForm'
+  requirements:
+    _role: 'trusted_user'
```
lramarojaona, line 7: We should create a specific permission which allows better flexibility

lib/modules/eic_private_message/src/Form/PrivateMessageForm.php
```diff
@@ -0,0 +1,242 @@
...
+
+    if (empty($recipients)) {
+      $recipients = [$user_id];
+    }
+
+    $form['from'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('From', [], ['context' => 'eic_private_message']),
+      '#value' => $this->t(
+        '"@fullname at European Innovation Council" @site_email',
```
lramarojaona, line 148: We shoud use the site_name variable here instead of hard-coded name

lib/modules/eic_user/src/UserHelper.php
```diff
@@ -147,4 +148,23 @@ public static function isPowerUser(AccountInterface $account) {
     return FALSE;
   }
 
+  /**
+   * @param \Drupal\user\UserInterface|NULL $user
+   *
+   * @return string
+   */
+  public function getFullName(UserInterface $user = NULL): string {
```
lramarojaona, line 156: There should be a helper function for this, check the realname module

lramarojaona (changes requested):

config/sync/core.entity_form_display.profile.member.default.yml
Dromadoros: Ok just to be sure, trusted users will be able to modify User and Profile ?

lramarojaona: Yes, that's the goal, users hould be able to modify:

- user settings
- their own member profile

lramarojaona:
@Dromadoros , FYI as SA I see the envelop but nothing happens when I click on it (it seems there's no URL included in that link, no href)

lramarojaona:
Maybe it's the right time to switch to the core's Contact module instead and build around this functionnality

Dromadoros:
I really would not prefer to go on contrib for simple stuff like that, in 5hours the feature is ready with small feedbacks on it. In long term we will have a full control of what we did and not installing patch, investigating, etc. I think for MVP we can keep it like that but still can debate about that feature after, what do you think ?

PS: I did all changes

Dromadoros:
> @Dromadoros , FYI as SA I see the envelop but nothing happens when I click on it (it seems there's no URL included in that link, no href)

It was not working due to the missing profile link to your user since the field was in the profile. I did the update with User entity, now it should work! :)

lib/themes/eic_community/react/components/Block/Overview/Search/ResultItem/UserGalleryResultItem.js
```diff
@@ -61,9 +61,9 @@ class UserGalleryResultItem extends React.Component {
 
             <div className="ecl-teaser__meta-footer">
               <div className="ecl-teaser__actions">
-                {this.props.result.ss_user_mail &&
+                {0 !== parseInt(this.props.result.ss_user_allow_contact) &&
                 <div className="ecl-teaser__action">
-                  <a href={`mailto:${this.props.result.ss_user_mail}`}
+                  <a href={this.props.result.ss_user_link_contact}
```
lramarojaona, line 66: I would definetely add a `destination` parameter to that link so that we are redirected back to the previous page once the form is subimtted

lramarojaona (changes requested):

lramarojaona:
This works, but we probably need a followup ticket for the test copy (messages and so...)

lramarojaona (approved):

PR #795, opened 2021-11-03
[EICNET-1611]feat: Add filter on date interval

Dromadoros:


filipvt1 (changes requested):
A multi-day event will not show up if you select a range that doesn't include the entire range of the event. For example an event going from the 10th to the 15th will not show up if you search on dates from 9th to 13th. It does show up if you select 9th till the 16th for example.

PR #806, opened 2021-11-05
[EICNET-1591] fix: An endpoint archival/delete requests

korhankale:
## To Test

- [ ] Make an archival/delete request on a comment in a discussion
- [ ] Run cron (be sure to have at least one site admin/content admin user)
- [ ] You should get an notification e-mail and see the request in the admin list


4hardfire (approved):
Looks ok to me!

PR #812, opened 2021-11-08
[EICNET-1647] I want an option on group invitation so that I can resend an invitation to a user

korhankale:
## To Test
- [x] Invite a user to a group
- [x] Go to [GROUP_URL]/invitations
- [x] Click to "Resend invitation"
- [x] Pay attention, the e-mail should be the same as the first one (it must also be different depending on whether the user exists or not)

### DIFF
--- a/lib/modules/eic_groups/src/Controller/GroupMembershipInvitationController.php
+++ b/lib/modules/eic_groups/src/Controller/GroupMembershipInvitationController.php
@@ -7,6 +7,7 @@ use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Mail\MailManagerInterface;
 use Drupal\Core\Messenger\MessengerInterface;
 use Drupal\Core\Routing\RedirectDestinationInterface;
+use Drupal\eic_groups\EICGroupsHelper;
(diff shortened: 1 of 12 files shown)


lib/modules/eic_groups/src/Controller/GroupMembershipInvitationController.php
```diff
@@ -0,0 +1,150 @@
...
+   */
+  private $groupContentEnablerManager;
+
+  /**
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   * @param \Symfony\Component\HttpFoundation\RequestStack $request_stack
+   * @param \Drupal\Core\Routing\RedirectDestinationInterface $redirect_destionation
+   * @param \Drupal\Core\Messenger\MessengerInterface $messenger
+   * @param \Drupal\Core\Mail\MailManagerInterface $mail_manager
+   * @param \Drupal\group\Plugin\GroupContentEnablerManagerInterface $group_content_enabler_manager
```
4hardfire, line 46: Missing parameter descriptions.

4hardfire (changes requested):
Just some missing descriptions needed to pass phpcs. The rest seems good to me!

PR #851, opened 2021-11-15
EICNET-1689/EICNET-296: Homepage groups/events

4hardfire:
### Features

- Create new view block to show last updated groups in the homepage;
- Create new view block to show list of upcoming events in the homepage;
- Add group moderation state and date changed to group solr index.

### Tests

- [x] Go to the homepage
- [x] Check if the list of groups is ordered by "Date changed". Make sure only published groups are shown.
- [x] Check if the list of upcoming events are ordered by Event date ASC. Also make sure past events are not shown.

### Todo:

- [x] We need a new follow-up ticket to create the teaser view for the event item in the list

### DIFF
--- a/lib/themes/eic_community/includes/preprocess/views.inc
+++ b/lib/themes/eic_community/includes/preprocess/views.inc
@@ -21,3 +21,19 @@ function eic_community_preprocess_views_view_field(&$variables) {
 
   }
 }
+
+/**
+ * Implements template_preprocess_views_view_field().
+ */
+function eic_community_preprocess_views_view(&$variables) {
+  /** @var \Drupal\views\ViewExecutable $view */
+  $view = $variables['view'];
+  $variables['is_compact'] = FALSE;
+
+  switch ($view->id()) {
+    case 'groups_homepage':
+      $variables['is_compact'] = TRUE;
+      break;
+
+  }
+}
--- a/lib/themes/eic_community/includes/preprocess/views.inc
+++ b/lib/themes/eic_community/includes/preprocess/views.inc
@@ -30,32 +30,29 @@ function eic_community_preprocess_views_view_field(&$variables) {
 function eic_community_preprocess_views_view(&$variables) {
   /** @var \Drupal\views\ViewExecutable $view */
   $view = $variables['view'];
-  $variables['is_compact'] = FALSE;
 
-  switch ($view->id()) {
-    case 'groups_homepage':
-      // Adds call to actions.
-      if ($view->current_display === 'block_groups_homepage') {
-        if ($call_to_action_url = GlobalOverviewPages::getGlobalOverviewPageUrl('groups')) {
-          $variables['call_to_action'] = [
-            'link' => [
-              'label' => 'View all groups',
-              'path' => $call_to_action_url,
-            ],
-          ];
-        }
+  $key = $view->id() . '-' . $view->current_display;
+  switch ($key) {
+    case 'groups_homepage-block_groups_homepage':
+      if ($call_to_action_url = GlobalOverviewPages::getGlobalOverviewPageUrl('groups')) {
+        $variables['call_to_action'] = [
+          'link' => [
+            'label' => 'View all groups',
+            'path' => $call_to_action_url,
+          ],
+        ];
       }
-      elseif ($view->current_display === 'block_events_homepage') {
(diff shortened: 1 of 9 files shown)


lib/themes/eic_community/includes/preprocess/views.inc
```diff
@@ -21,3 +23,40 @@ function eic_community_preprocess_views_view_field(&$variables) {
...
   }
 }
+
+/**
+ * Implements template_preprocess_views_view_field().
+ */
+function eic_community_preprocess_views_view(&$variables) {
+  /** @var \Drupal\views\ViewExecutable $view */
+  $view = $variables['view'];
+  $variables['is_compact'] = FALSE;
```
lramarojaona, line 33: @4hardfire , trying to understand what this? `is_compact`
Why do we set this for all views?

lib/themes/eic_community/includes/preprocess/views.inc
```diff
@@ -21,3 +23,40 @@ function eic_community_preprocess_views_view_field(&$variables) {
...
+
+  switch ($view->id()) {
+    case 'groups_homepage':
+      // Adds call to actions.
+      if ($view->current_display === 'block_groups_homepage') {
```
lramarojaona, line 38: You can probably make this code more readable. What I usually do is:
```php
$key = $view->id() . '-' . $view->current_display;
switch ($key) {
  case 'groups_homepage-block_groups_homepage':
    // Do something.
    break;
}
```

lib/themes/eic_community/includes/preprocess/views.inc
4hardfire: This variable is now used in **views-view--featured-content-collection.html.twig**. But indeed, maybe I shouldn't set it for all views. I will improve this.

lramarojaona:
Followup ticket for the teaser view mode is here: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1680

lramarojaona:
Followup ticket for the teaser view mode is here: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1680

lib/themes/eic_community/includes/preprocess/views.inc
4hardfire: I re-worked on this a little bit and instead of adding that variable, I created a new template for the compact version of `views-view--featured-content-collection`.

lib/themes/eic_community/includes/preprocess/views.inc
4hardfire: Improved :)

lib/themes/eic_community/includes/preprocess/views.inc
```diff
@@ -30,32 +30,29 @@ function eic_community_preprocess_views_view_field(&$variables) {
...
-          ];
-        }
+
+  $key = $view->id() . '-' . $view->current_display;
+  switch ($key) {
+    case 'groups_homepage-block_groups_homepage':
+      if ($call_to_action_url = GlobalOverviewPages::getGlobalOverviewPageUrl('groups')) {
+        $variables['call_to_action'] = [
+          'link' => [
+            'label' => 'View all groups',
```
lramarojaona, line 40: This string will sad without a `t()` function :)

lib/themes/eic_community/includes/preprocess/views.inc
```diff
@@ -30,32 +30,29 @@ function eic_community_preprocess_views_view_field(&$variables) {
...
-            ],
-          ];
-        }
+      break;
+
+    case 'groups_homepage-block_events_homepage':
+      if ($call_to_action_url = GlobalOverviewPages::getGlobalOverviewPageUrl('events')) {
+        $variables['call_to_action'] = [
+          'link' => [
+            'label' => 'View all events',
```
lramarojaona, line 51: Same here ;)

lib/themes/eic_community/includes/preprocess/views.inc
4hardfire: my bad. It's done!

lib/themes/eic_community/includes/preprocess/views.inc
4hardfire: done!

lramarojaona (approved):

PR #852, opened 2021-11-15
Feature/eicnet 1573 topics activity stream

lramarojaona:
### Tests

- [ ] Go on a Topic detail page
- [ ] Check that you have the activity stream showing topic related messages only

### Remarks

- The block for the group context activity stream has been updated accordingly
- Solr needs to be re-indexed first
- The react block should handle empty members data (do not show the block if there's no data) => Theo


lib/modules/eic_search/src/Plugin/Block/ActivityStreamBlock.php
```diff
@@ -121,27 +121,116 @@ public function __construct(
...
+
+    $has_delete_permission = FALSE;
+    if ($group) {
+      $has_delete_permission = EICGroupsHelper::userIsGroupAdmin($group, $this->currentUser);
+    }
+
+    $build['#attached']['drupalSettings']['overview'] = [
+      'has_permission_delete' => $has_delete_permission,
+    ];
+
```
Dromadoros, line 179: double line space

lramarojaona: Thanks, this is fixed now

Dromadoros (approved):

PR #853, opened 2021-11-15
[EICNET-1493]feat: put an url on the user teaser for group

Dromadoros:
How to test:

- [x] Re-index everything in search api
- [x] Go to /groups and check if there is a link for the group owner user


lramarojaona (approved):
Works fine

PR #855, opened 2021-11-15
EICNET-1546: Fix group block/publish operation links

4hardfire:
### Fixes

- Fix block/publish operation links so that the destination URL takes into account the base path.

### Tests

- [ ] Go to a published group page
- [ ] Click on block
- [ ] Add a reason a press block
- [ ] You should be redirected back to the group page and you shouldn't get any error like: `Redirects to external URLs are not allowed by default, use \Drupal\Core\Routing\TrustedRedirectResponse`

### DIFF
--- a/lib/modules/eic_flags/src/Hooks/EntityOperations.php
+++ b/lib/modules/eic_flags/src/Hooks/EntityOperations.php
@@ -165,6 +165,16 @@ class EntityOperations implements ContainerInjectionInterface {
       ];
     }
 
+    if ($entity->hasLinkTemplate('block-entity')) {
+      if ($entity->toUrl('block-entity')->access()) {
+        $operations['block'] = [
+          'title' => $this->t('Block'),
+          'url' => $entity->toUrl('block-entity')
+            ->setRouteParameter('destination', $this->currentRequest->getRequestUri()),
+        ];
+      }
+    }
+
     return $operations;
   }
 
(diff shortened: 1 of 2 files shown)


lib/modules/eic_flags/src/Hooks/EntityOperations.php
```diff
@@ -165,6 +165,16 @@ public function getOperations(EntityInterface $entity) {
       ];
     }
 
+    if ($entity->hasLinkTemplate('block-entity')) {
```
Dromadoros, line 168:
```suggestion
    if ($entity->hasLinkTemplate('block-entity') && $entity->toUrl('block-entity')->access()) {
```

lib/modules/eic_flags/src/Hooks/EntityOperations.php
```diff
@@ -184,6 +194,17 @@ public function getAdminOperations(EntityInterface $entity) {
       return $handler->getActions($entity);
     }
 
+    $operations = [];
+    if ($entity->hasLinkTemplate('block-entity')) {
```
Dromadoros, line 198: same, just add &&

4hardfire: Done!

lib/modules/eic_flags/src/Hooks/EntityOperations.php
4hardfire: Done!

Dromadoros (approved):

PR #856, opened 2021-11-16
Feature/eicnet topics details fe

Askerx9:
### What I did

- [ ] add topic in storybook (http://localhost:9000/?path=/story/bundles-topic--topic-overview)
- [ ] slice elements

### DIFF
--- a/lib/themes/eic_community/data/featured-content-collection/companies.data.js
+++ b/lib/themes/eic_community/data/featured-content-collection/companies.data.js
@@ -28,17 +28,5 @@ export default {
     },
     4
   ),
-  // items: mockItems(
-  //   {
-  //     content: TeaserTemplate(
-  //       Object.assign(
-  //         {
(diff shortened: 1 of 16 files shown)


filipvt1 (changes requested):
The group teaser looks different. In a recent commit there was some padding added to the teaser component. Because of this the image received some additional padding and doesn't align with the border of the teaser anymore. Either remove the padding for group teaser or pull the image back with negative margins.

Askerx9:
ha yes you right @filipvt1 now it's better

filipvt1 (approved):

filipvt1 (changes requested):
I would apply the zero padding specifically for the group teaser. Because the group teaser is shown on different pages and is still broken there.

Askerx9:
@filipvt1 so weird but normally with the recent commit it's good now

filipvt1 (approved):

PR #861, opened 2021-11-16
[EICNET] Feat: create home teaser event in storybook

Askerx9:
### What I did

- http://localhost:9000/?path=/story/compositions-teaser--event-home-teaser
- https://www.figma.com/file/WZH3qhNynnmm0H6FKmD3hi/EASME_v1?node-id=2677%3A14950


filipvt1 (changes requested):
- Alignment of the "Remote event" icon and the image is off. There is a margin-left on ".ecl-teaser--event .ecl-teaser__meta-column". Can you check if this serves a purpose because it also seems to mess up the alignment in the regular event teaser
- The homepage teaser shouldn't show views and likes counter
- The counter with the amount of registrations is missing

PR #874, opened 2021-11-17
EICNET-296 Homepage upcoming events

4hardfire:
### Features

- New view mode "Small teaser" created for group type Event;
- Preprocessed group type event variables for the small_teaser view mode;
- Glue back-end with front-end.

### Tests

- [ ] Create a couple of events (on going and upcoming)
- [ ] Join the event with 3 different users
- [ ] Go to the homepage and check if the look and feel of the upcoming events list matches the mockups

### Todo - follow up

- [x] The map pin icon has a blue color and in the mockups it is shown in grey;
- [x] Make the tags clickable and redirect the user to the events overview page with applied topic filter;
- [ ] Improve event teaser template so that it shows the title before the topic tags.


Dromadoros:
Teaser date should show the start and end. Also months should be in short version

image

filipvt1:
- The icon isn't showing correctly. In storybook this is either the camera icon for remote or the marker for on site event. Now it tries to show the icon called "event" but that doesn't exist.
- In ECL it is documented that an ongoing event should have a yellow date block (by adding the "ecl-date-block--ongoing" class). This is currently not the case

4hardfire:
> * The icon isn't showing correctly. In storybook this is either the camera icon for remote or the marker for on site event. Now it tries to show the icon called "event" but that doesn't exist.
> * In ECL it is documented that an ongoing event should have a yellow date block (by adding the "ecl-date-block--ongoing" class). This is currently not the case

- Currently we only show the marker because we can have both locations (on site and remote) for the event. I think this should be re-worked on a different ticket after discussing the final solution. CC @lenglebert.

- Currently ongoing and upcoming events are showing correctly on my local (check image) 
image

Dromadoros (approved):

PR #882, opened 2021-11-18
[EICNET] fix: Group groups by bundle

korhankale:
## To test

- [x] Go to a discussion
- [x] Click on share with another group > The list should be grouped by bundle


filipvt1:
Frontend OK

lib/themes/eic_community/includes/preprocess/common.inc
```diff
@@ -76,9 +76,16 @@ function _eic_community_get_share_group_content_link(
         continue;
       }
 
-      $formatted_groups[] = [
+      if (!isset($formatted_groups[$group->bundle()])) {
+        $formatted_groups[$group->bundle()] = [
+          'label' => ucfirst($group->bundle()),
```
lramarojaona, line 81: You can get the real label with this (just tested):
`$group->type->entity->label()`

lramarojaona (changes requested):

PR #887, opened 2021-11-18
[EICNET]feat: Add state label of the event in group overview

Dromadoros:
Send the state of the event (ongoing, future, past) to solr. Also add a Drupal cron that will update event each 12hours

How to test:

- [x] Re-index elements in search api
- [x] Clear your cache
- [x] Go to group events overview
- [x] Be sure it's ordered 1. Ongoing 2. Future 3. Past
- [x] You can also filter by this state

### DIFF
--- a/config/sync/context.context.group_events_overviews.yml
+++ b/config/sync/context.context.group_events_overviews.yml
@@ -21,7 +21,7 @@ conditions:
 reactions:
   blocks:
     blocks:
-      5192d3d3-7de9-4cad-baa2-4195a79584f0:
+      5b2d7e48-2503-4cfc-bf97-ea9585c5cc86:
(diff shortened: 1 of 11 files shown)


4hardfire (commented):
Everything seems pretty fine for me. Although, I found an issue when trying to edit the overview page. Apparently the sorting options are not updated after I save the overview page entity.

4hardfire (commented):
Everything seems pretty fine for me. Although, I found an issue when trying to edit the overview page. Apparently the sorting options are not updated after I save the overview page entity.

4hardfire (approved):
Approved :). We need a follow-up ticket to solve the issue when saving the overview page entity.

PR #889, opened 2021-11-19
[EICNET-1683] add share icon in group flags

Askerx9:
### before
image

### new
image


filipvt1 (changes requested):
It is strange that you can't click on the icon to share. Can you check if it is possible to remove the icon from the html and add it with some extra css on the "a" tag that is added by the javascript library?

Askerx9:
@filipvt1 not possible because the "a" tag is create by a plugin un js. I can try to add an "onClick" attribute with a JS redirection on the svg but it's ugly

Askerx9:
@filipvt1 I checked again and it is not possible to do the onClick on the svg so we have to keep it that way.

filipvt1 (approved):

PR #890, opened 2021-11-19
[EICNET-1719]chore: Refactor the way we get global overview page

Dromadoros:
Now we get overview page by an ID directly linked to the entity via the new field "field_overview_id"

### DIFF
--- a/lib/themes/eic_community/includes/preprocess/blocks/block--eic-group-header-block.inc
+++ b/lib/themes/eic_community/includes/preprocess/blocks/block--eic-group-header-block.inc
@@ -182,7 +182,7 @@ function eic_community_preprocess_eic_group_header_block(array &$variables) {
   ];
 
   // Adds url to navigate back to the list of groups.
-  $back_url = GlobalOverviewPages::getGlobalOverviewPageUrl('groups');
+  $back_url = GlobalOverviewPages::getGlobalOverviewPageUrl(GlobalOverviewPages::GROUPS);
   $variables['group_values']['back_url'] = $back_url->toString();
 
 }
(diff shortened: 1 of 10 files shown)


lib/themes/eic_community/includes/preprocess/blocks/block--eic-group-header-block.inc
```diff
@@ -182,7 +182,7 @@ function eic_community_preprocess_eic_group_header_block(array &$variables) {
   ];
 
   // Adds url to navigate back to the list of groups.
-  $back_url = GlobalOverviewPages::getGlobalOverviewPageUrl('groups');
+  $back_url = GlobalOverviewPages::getGlobalOverviewPageUrl(GlobalOverviewPages::GROUPS);
```
4hardfire, line 185: I think you still need to change the parameter type in `getGlobalOverviewPageUrl` so that it accepts an integer instead of a string right?

4hardfire (changes requested):

lib/themes/eic_community/includes/preprocess/blocks/block--eic-group-header-block.inc
Dromadoros: Yes right, it was working because autoconvert the int to string

4hardfire (approved):
Seems ok to me! Do not forget to update overview pages in QA 😄

korhankale:
@Dromadoros @4hardfire is it really a good idea to use id to retrieve overview pages? Shouldn't we use the source type with/or any other field which helps identifying the queried overview? <br> This approach isn't a robust one according to me (run the fixtures locally twice and you'll get a 500 on every group pages)

PR #892, opened 2021-11-19
Feature/eicnet 1715 topics overview page

lramarojaona:
### Tests

- [x] Go to the /topics page
- [x] Check that you have one view of terms for each top level term

### Remarks

- FE and taxonomy teaser will be part of other tickets


4hardfire (changes requested):
I have an error when trying to access `/topics/horizontal`:
`Error: Call to a member function get() on null in eic_topics_preprocess_page__taxonomy__term() (line 18 of /var/www/html/lib/themes/eic_community/includes/preprocess/taxonomy/taxonomy_term.inc)`

lib/modules/eic_overviews/src/Controller/TaxonomyOverviewController.php
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * The ID of the views display to be used.
+   *
+   * @var string
+   */
+  const TAXONOMY_VIEWS_DISPLAY_ID = 'block_taxonomy_overview';
+
+  /**
+   * The entity_type.manager service.
```
4hardfire, line 31: Usually I type `The Entity type manager service` instead of `The entity_type.manager service`. Because the service machine name can suffer some changes in the future and in that case it will not make sense anymore.

4hardfire (changes requested):

lib/modules/eic_overviews/src/Controller/TaxonomyOverviewController.php
lramarojaona: Mmm, I've seen this example somewhere, but I can change it, I checked examples in core and indeed they don't use the machine name to explain

4hardfire (approved):
Seems fine to me!

lramarojaona:
> I have an error when trying to access `/topics/horizontal`: `Error: Call to a member function get() on null in eic_topics_preprocess_page__taxonomy__term() (line 18 of /var/www/html/lib/themes/eic_community/includes/preprocess/taxonomy/taxonomy_term.inc)`

Yes, this has been fixed now

PR #896, opened 2021-11-22
[EICNET] fix: Local setup dependencies

korhankale:


lib/modules/eic_statistics/src/StatisticsHelper.php
```diff
@@ -49,32 +50,32 @@ class StatisticsHelper {
   protected $commentsHelper;
 
   /**
-   * Constructs a StatisticsHelper object.
```
lramarojaona, line 52: Is there a reason to remove thos description or is it a merge issue?

korhankale: Actually the type-hinting was wrong for one of them and since this kind of comment doesn't bring any added value I removed them 😅 
image

lramarojaona (approved):

PR #897, opened 2021-11-22
[EICNET-1628] correct design for user profile and member overview

Askerx9:
https://citnet.tech.ec.europa.eu/CITnet/jira/secure/RapidBoard.jspa?rapidView=6628&projectKey=EICNET&view=detail&selectedIssue=EICNET-1628&quickFilter=26822
https://citnet.tech.ec.europa.eu/CITnet/jira/secure/RapidBoard.jspa?rapidView=6628&projectKey=EICNET&view=detail&selectedIssue=EICNET-1626&quickFilter=26822

https://www.figma.com/file/WZH3qhNynnmm0H6FKmD3hi/EASME_v1?node-id=2677%3A23993

http://localhost:9000/?path=/story/bundles-member--full-private


filipvt1 (changes requested):
- The styling of the secondary buttons seems off. For some reason the padding is defined like this: `calc(.75rem - 2px) calc(1rem - 2px)` while for the primary button it is like this `.75rem 1rem`. This causes the secondary button to be 4px smaller than the primary button. Can you check if removing that calc function in the padding of the secondary button causes any issues (the secondary button is used in other places as well)
- The "My events" tab doesn't do anything
- The star icon is aligned to high compared to the user's name. In the designs there is a different icon shown there (a trophy)

Askerx9:
> * The styling of the secondary buttons seems off. For some reason the padding is defined like this: `calc(.75rem - 2px) calc(1rem - 2px)` while for the primary button it is like this `.75rem 1rem`. This causes the secondary button to be 4px smaller than the primary button. Can you check if removing that calc function in the padding of the secondary button causes any issues (the secondary button is used in other places as well)
> * The "My events" tab doesn't do anything
> * The star icon is aligned to high compared to the user's name. In the designs there is a different icon shown there (a trophy)

@filipvt1 done

filipvt1 (approved):

PR #902, opened 2021-11-22
Feature/eicnet 1716 topic term teaser

lramarojaona:
### Tests

- [x] Make sure you have at least one Topic with an image to test
- [x] Go on `/topics` page and check that the teaser has: image, name, experts and organisations elements

### Remarks

- This PR needs https://github.com/EIC-EA/EIC-community-D8/pull/899 to be merged first


Askerx9 (changes requested):
image

The front-end is broken 

maybe you can use this template: topic.overview.item.html.twig with topics like here : https://dev-eic-afb067c90dae.victhorious.com/themes/custom/eic_community/storybook-static/?path=/story/compositions-topics-list--topics-list

lib/themes/eic_community/includes/preprocess/taxonomy/taxonomy_term.inc
```diff
@@ -5,26 +5,86 @@
...
+      // Get the thumbnail.
+      /** @var \Drupal\media\MediaInterface $media */
+      $media = $term->get('field_topic_banner')->referencedEntities()[0];
+      if ($media && !empty($media->get('oe_media_image')->getValue())) {
+        $style_name = 'gallery_teaser_crop_160x160';
+        $teaser['image'] = ImageValueObject::fromStyledImageItem($media->get('oe_media_image')->first(), $style_name);
+      }
+      $teaser['stats'] = [];
+      $teaser['term_item'] = $teaser;
+      break;
```
4hardfire, line 47: Logic seems fine to me. I would just create a new file to handle logic related to vocabulary topic and call those methods in the taxonomy_term.inc using a switch.
So the structure would be like:

- taxonomy_term.inc (for common methods)
- taxonomy-term--topics.inc
- ....
- taxonomy-term--[vocabulary-name].inc

What do you think?

4hardfire (approved):
Seems ok to me!

PR #903, opened 2021-11-22
[EICNET-1597]chore: Re-do the way we build prefiltered query params

Dromadoros:


lib/modules/eic_topics/eic_topics.module
```diff
@@ -108,15 +116,15 @@ function eic_topics_views_pre_render(ViewExecutable $view) {
...
+        'sm_content_field_vocab_topics_string' => $term->label(),
+        'ss_global_content_type' => ['wiki_page'],
       ];
       break;
 
     case 'experts':
       $route = 'eic_search.people';
       $params = [
-        'itm_aggregated_field_vocab_topics' => $term_id,
+        'sm_user_profile_field_vocab_topic_interest_array' => $term->label(),
```
lramarojaona, line 127: This is wrong, it should be the topic of expertise, you can use this one:
`user_profile_topic_expertise_string`

I don't know if it already exists as a facet though

lramarojaona (changes requested):

lib/modules/eic_topics/eic_topics.module
```diff
@@ -77,23 +85,23 @@ function eic_topics_views_pre_render(ViewExecutable $view) {
...
       ];
       break;
 
     case 'files':
       $route = 'eic_search.global_search';
       $params = [
-        'itm_aggregated_field_vocab_topics' => $term_id,
-        'ss_global_content_type' => 'document,gallery,video',
+        'sm_content_field_vocab_topics_string' => $term->label(),
+        'ss_global_content_type' => ['document', 'gallery', 'video'],
```
lramarojaona, line 104: In this case the search page doesn't return any result, I guess the facet should treat this as an `OR `instead of an `AND`, but this is probably not in the scope of this ticket

Dromadoros: it's already the case for topics but not content type but you are right, we can do this in another ticket

lib/modules/eic_topics/eic_topics.module
Dromadoros: Ok it's done and yes the facet exists, thx

lramarojaona (approved):

PR #904, opened 2021-11-23
[EICNET-1709] chore: Allow delete/archival requests for draft groups

korhankale:
## To Test

- [ ] Create a group and switch it to draft
- [ ] You should be able to make delete/archival requests


lib/modules/eic_groups/src/Access/FlagRequestAccessCheck.php
```diff
@@ -72,7 +72,6 @@ public function access(Route $route, RouteMatchInterface $route_match, AccountIn
     // Deny access to flag if the group IS in pending or draft state.
```
lramarojaona, line 72: Maybe also adapt the comment?

lramarojaona (approved):
Works, just one small remark with comment

PR #906, opened 2021-11-23
[EICNET-1694]fix: ADd correct permission + icon state for highlight content

Dromadoros:
How to test:

- [x] SA/SCM/GO/GA should be able to highlight content
- [x] Go as trusted user (not group owner) and check if highlight link is disable


lramarojaona:
Should the star icon be always hidden if the user does not have the permission to (un-)higlight the item?

Dromadoros:
> Should the star icon be always hidden if the user does not have the permission to (un-)higlight the item?

Yes! Done

PR #907, opened 2021-11-23
Feature/eicnet 1718 topics overview fe

lramarojaona:
### Improvements

- There is a new template that can be used: `views-view-unformatted--nude.html.twig`. It basically removes the surrounding `<div>` around the row

### Tests

- [x] Go on `/topics` page
- [ ] Check that you have a correct display of the overview

### Remarks

- This PR needs followingPRs to be merged first:
  - https://github.com/EIC-EA/EIC-community-D8/pull/892
  - https://github.com/EIC-EA/EIC-community-D8/pull/899
  - https://github.com/EIC-EA/EIC-community-D8/pull/902
- Page does not have a correct container (it is full width right now)


lramarojaona:
> Could you add "ecl-container" in the class of the global div ? image

@Askerx9 , This is indeed missing but I believe this shouldn't apply to this specific element but on a higher level somewhere in https://github.com/EIC-EA/EIC-community-D8/blob/develop/lib/themes/eic_community/patterns/layout/base.html.twig
So that it would apply to all pages, all elements. But this requires another ticket it"s not part of this ticket.

Askerx9:
@lramarojaona  no, you can add it on the .ecl-topic__overview, we don't add it on the base.twig.html because we have components that use full width and we don't need a container for them.

PR #911, opened 2021-11-23
[EICNET-1700]feat: Add cache tags to comments block + condition to hide if comment system not open

Dromadoros:
How to test it:

- [x] Check if "Document" has no more comment field settings in the node form display
- [x] Go to discussion and try to close comment system, check if you do not see it anymore in the FE
- [x] Open the comment system again et go again on the page. It should be closed


lramarojaona:
The **Closed** option from Comment module states `Users cannot post comments, but existing comments will be displayed. `
So I guess we should just hide the comment form and not all the comments.

The **Hidden** option should hide form + comments: `Comments are hidden from view. `

Dromadoros:
> The **Closed** option from Comment module states `Users cannot post comments, but existing comments will be displayed. ` So I guess we should just hide the comment form and not all the comments.
> 
> The **Hidden** option should hide form + comments: `Comments are hidden from view. `

My bad, forgot the closed state. Now the close will hide the comment form + action on comments (reply, like)

lramarojaona (approved):
Works well!

PR #913, opened 2021-11-23
[EICNET-1674]feat: Re-index dependant entities of group membership group content

Dromadoros:
How to test:

- [x] Go to group members front-end
- [x] Leave a group as a group member
- [x] Re-check if it has been sync to search api (check if it has been removed from the members group)
- [x] Go to groups list, check a group owner of a group
- [x] For the same group change the group owner
- [x] Go again on groups list and verify if group owner has been updated


Askerx9:
BGGGG!!! it works well

Askerx9 (approved):

PR #915, opened 2021-11-23
EICNET-1725: Error when adding a topic of interest in user profile

4hardfire:
### Fixes

- Do not flag topic of interest if the user is already following the topic.

### Tests

- [x] Go to `/topics/financial-development` and follow the topic
- [x] Edit the profile of the current user you are logged in with and add the **Financial development** topic in the topics of interest field
- [x] Make sure the page doesn't throw any error and check if there is no error in the logs such as: `Drupal\Core\Entity\EntityStorageException: The user has already flagged the entity with the flag. in Drupal\Core\Entity\Sql\SqlContentEntityStorage->save() (line 810 of /var/www/html/web/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php).`

### DIFF
--- a/lib/modules/eic_flags/src/Hooks/EntityOperations.php
+++ b/lib/modules/eic_flags/src/Hooks/EntityOperations.php
@@ -402,6 +402,13 @@ class EntityOperations implements ContainerInjectionInterface {
 
     // Follow new topics.
     foreach ($topics as $topic) {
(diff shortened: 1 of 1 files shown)


lramarojaona (approved):
It works but I guess that if someone follows a topic, it should also be reflected on his profile (Topics of interest).

I'll create a followup ticket

PR #916, opened 2021-11-24
EICNET-1689: Homepage last updated groups (rework)

4hardfire:
### Improvements

- Rename group teaser template and hook_preprocess_group into hook_preprocess_group__group;
- Add search api cache tags in Homepage events and groups so that it shows the correct updated items;

### Fixes

- Fix issue in events teaser when trying to generate links for the event tags.

### Tests

- [x] Go to the homepage;
- [x] Make sure you have 3 upcoming events and 3 groups;
- [x] Update one of the groups and make sure that group is shown as the first item in the list;
- [x] Update one of the events and make sure the list gets updated accordingly.


lramarojaona:
There are differences between logged-in and anonymous.
When I updated a group or event:

- the teaser was updated
- the order was updated for annoymous but not logged-in

However I don't know if we really need to have instantly updated content everywhere, this is kindof killing the cache feature

4hardfire:
> There are differences between logged-in and anonymous. When I updated a group or event:
> 
> * the teaser was updated
> * the order was updated for annoymous but not logged-in
> 
> However I don't know if we really need to have instantly updated content everywhere, this is kindof killing the cache feature

@lramarojaona on my local the order was updated also for logged-in. Concerning the cache, we should discuss with the other DEVs to see their opinion on it.

lramarojaona (approved):

PR #918, opened 2021-11-24
EICNET-1727: Event admin view - creation of owner

4hardfire:
### Improvements

- Create owner role for group type event;
- Update member and outsider permissions.

### Tests

- [x] Create a new Group type Event
- [x] Go to `/admin/community/events` and make sure the event is listed there


lramarojaona (approved):
Events or groups that don't have an owner don't show up, this can mayb be imrpoved later

4hardfire:
> Events or groups that don't have an owner don't show up, this can mayb be imrpoved later

CC: @lenglebert

PR #924, opened 2021-11-25
EICNET-1510: Email notification for a group transitioned to unblocked

4hardfire:
### Features

- Create new generic message type to use when an entity has changed status;
- Send notification message to group owner when SA/SCM unblocks a group.

### Tests

- [x] As SA/SCM, block a group
- [x] After blocking the group, go to the group edit form and published the group
- [x] Run cron
- [x] Make sure an email notification has been sent to the group owner

### DIFF
--- a/lib/modules/eic_messages/src/Hooks/EntityOperations.php
+++ b/lib/modules/eic_messages/src/Hooks/EntityOperations.php
@@ -104,19 +104,21 @@ class EntityOperations implements ContainerInjectionInterface {
       return;
     }
 
+    $message_uid = $this->getContentStatusChangedMessageReceiver($entity);
+    // If there is no receiver for the message, we don't send the notification.
+    if (!$message_uid) {
+      return;
+    }
+
     // Gets the entity type label to use in the message notification.
-    $entity_type_label = isset(self::ENTITY_TYPE_LABELS[$entity->getEntityTypeId()])
-      ? self::ENTITY_TYPE_LABELS[$entity->getEntityTypeId()]
-      : (isset(self::ENTITY_TYPE_LABELS["{$entity->getEntityTypeId()}__{$entity->bundle()}"])
-          ? self::ENTITY_TYPE_LABELS["{$entity->getEntityTypeId()}__{$entity->bundle()}"]
-          : $entity->getEntityTypeId());
+    $entity_type_label = strtolower($entity->type->entity->label());
 
     // By default notification is not skipped.
     $skip_notification = FALSE;
     // Default message fields.
     $message = [
       'template' => 'notify_content_status_changed',
-      'uid' => $entity->getOwnerId(),
+      'uid' => $message_uid,
       'field_message_subject' => NULL,
       'field_referenced_entity_label' => $entity->label(),
       'field_entity_prev_status_label' => $previous_state,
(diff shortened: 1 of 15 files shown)


lib/modules/eic_messages/src/Hooks/EntityOperations.php
```diff
@@ -55,26 +91,98 @@ public static function create(ContainerInterface $container) {
...
+    if ($previous_state === $current_state) {
+      return;
+    }
+
+    // Gets the entity type label to use in the message notification.
+    $entity_type_label = isset(self::ENTITY_TYPE_LABELS[$entity->getEntityTypeId()])
+      ? self::ENTITY_TYPE_LABELS[$entity->getEntityTypeId()]
+      : (isset(self::ENTITY_TYPE_LABELS["{$entity->getEntityTypeId()}__{$entity->bundle()}"])
+          ? self::ENTITY_TYPE_LABELS["{$entity->getEntityTypeId()}__{$entity->bundle()}"]
+          : $entity->getEntityTypeId());
```
lramarojaona, line 112: You should be able to get the labels easily with something like:
`$entity->type->entity->label()`
instead of harcoding names that we already have in the DB.

I guess you should first test if there is a type.

lib/modules/eic_messages/src/Hooks/EntityOperations.php
```diff
@@ -55,26 +91,98 @@ public static function create(ContainerInterface $container) {
...
+      'field_referenced_entity_label' => $entity->label(),
+      'field_entity_prev_status_label' => $previous_state,
+      'field_entity_status_label' => $current_state,
+      'field_entity_url' => [
+        'uri' => $entity->toUrl()->toString(),
+        'title' => $entity->label(),
+      ],
+    ];
+
+    if ($entity->getEntityTypeId() === 'group') {
```
lramarojaona, line 130: Maybe put this (all this code block) in a separate function?

We may need to sned out notifications for other entity types, and this function is already quite big.

lramarojaona (changes requested):

lramarojaona:
Works well, please check comments ;)

lib/modules/eic_messages/src/Hooks/EntityOperations.php
4hardfire: Done!

lib/modules/eic_messages/src/Hooks/EntityOperations.php
4hardfire: Done!

lib/modules/eic_messages/src/Hooks/EntityOperations.php
```diff
@@ -19,18 +22,51 @@ class EntityOperations implements ContainerInjectionInterface {
...
+   * @var string
+   */
+  const MODERATION_STATE_BLOCKED = 'blocked';
+
+  /**
+   * The entity type labels to use in notification message.
+   *
+   * @var array
+   */
+  const ENTITY_TYPE_LABELS = [
```
lramarojaona, line 43: I guess we can now get rid of this?

lramarojaona (approved):

lib/modules/eic_messages/src/Hooks/EntityOperations.php
lramarojaona: Great!

lib/modules/eic_messages/src/Hooks/EntityOperations.php
lramarojaona: It's now done

PR #929, opened 2021-11-25
Bugfix/20211125 ludo

lramarojaona:
### Fixes

- I had following error when re-indexing content:

_Unable to decode output into JSON: Syntax error_
_Error: Call to a member function id() on null in Drupal\eic_groups\EICGroupsHelper->getGroupByEntity() (line 261 of /app/lib/modules/eic_groups/src/EICGroupsHelper.php)._

This is probably due to a message referencing a group that doesn't exist anymore.

- Anonymous and outsiders couldn't see public/published event groups


lib/modules/eic_groups/src/EICGroupsHelper.php
```diff
@@ -281,9 +281,10 @@ public function getGroupByEntity(EntityInterface $entity) {
     }
 
     if ($entity instanceof MessageInterface) {
-      $group_ref_id = $entity->hasField('field_group_ref') ?
-        $entity->get('field_group_ref')->entity->id() :
-        NULL;
+      $group_ref_id = NULL;
+      if ($entity->hasField('field_group_ref') && !empty($entity->get('field_group_ref')->entity)) {
```
Dromadoros, line 285: `!$entity->get('field_group_ref')->entity instanceof GroupInterface`

isn't it better so we know it's not null and the correct Entity type (even if we are pretty sure we will have group)

lramarojaona: I've added a new commit, you still need the `hasField()` before the `get()`

Dromadoros (approved):

PR #930, opened 2021-11-25
[EICNET] correct style for pagination

Askerx9:
https://www.figma.com/file/WZH3qhNynnmm0H6FKmD3hi/EASME_v1?node-id=2880%3A18204
http://localhost:8080/groups


filipvt1 (changes requested):
The styling of the inactive item (regular state, hover state and focus state) isn't correct yet.
image
Also check styling on hover of next/prev arrows.

filipvt1 (changes requested):
The current item is currently shown in a blue text, it should be black. Probably caused by the link inside the active item. In storybook it isn't clickable. This is part of the library used for pagination I guess so I would just update the styling so an active link is styled in black.

Askerx9:
it's done and now it's not clickable @filipvt1

filipvt1 (approved):

PR #932, opened 2021-11-26
EICNET-1673: Fix issue with wrong visibility for stories/news.

lramarojaona:
Bugfix: when the value of private field is empty (NULL), it was showing Community members only while remaining available for anonymous users.

This mostly applies to contents not created throught the UI where the value would be set automaticlaly depending on the checkbox. (like content generation)


4hardfire (approved):
Looks fine to me!

PR #933, opened 2021-11-26
EICNET-1771: Configure pathauto to not create new aliases automatically on title change.

lramarojaona:
### Tests

- [x] Change the title of node, don't touch URL alias field
- [x] URL alias should remain the same
- [x] Change the URL alias of the node
- [x] Node alias should now be the new one and the old one should be redirected to the new one


4hardfire (approved):
It works well!

PR #934, opened 2021-11-26
[EICNET] rework style of the comments modal

Askerx9:
### What I did

- [ ] Refactor the code to reuse components for each comments modal
- [ ] Apply a correct style


lib/themes/eic_community/react/components/Block/CommentsDiscussion/Partials/PopupRequest.js
```diff
@@ -1,91 +1,74 @@
...
+    };
 
     this.submit = this.submit.bind(this);
   }
 
   submit(e, closeFunction) {
     closeFunction();
-    console.log(this.props.reason)
-    this.props.makeRequest('comment', this.props.comment.comment_id, this.props.type, this.state.reason)
+    console.log(this.props.reason);
```
Dromadoros, line 21: we can remove this, I know that was my fault

filipvt1 (changes requested):
Whenever I open the "Edit comment" dialog the focus is on the user profile picture. It would be more logical to have the focus on the input field. For delete and archival the focus is correct so should only be updated for "edit" action. All other things look good. 
Don't know if it is an option to make the user not clickable. That would make the next focus-able item the input field. Otherwise you'll have to implement some custom logic.

Askerx9:
@filipvt1  not possible for the moment in this version we can't disable autofocus on the first element, maybe in the next version of the plugin https://github.com/yjose/reactjs-popup/issues/278

Askerx9:
@filipvt1 
I have made some changes to this:

- TaggedUser are implemented
- TaggedUser have a good design
- TaggedUsers can be tagged on the children of comments
- Some corrections on the css

filipvt1 (approved):

PR #936, opened 2021-11-26
[EICNET] create mail template

Askerx9:
check if the mail is correct; http://localhost:8080//themes/custom/eic_community/mails/mail-dist/group_ownership_transfer.html


filipvt1 (changes requested):
The text inside the yellow button doesn't have the right color. It is blue now but the standard ECL button has a black text color.

Askerx9:
done @filipvt1

filipvt1 (approved):

PR #941, opened 2021-11-29
[EICNET-1774]feat: Add tagging user system on comment

Dromadoros:
How to test it:

- [x] Go to a discussion detail
- [x] Add a comment and you will see all possible users depending the visibility of the group
- [x] Select one/multiple user(s)
- [x] send comment
- [x] You should see your new comments with a new line into the comment box "With: XXX"


lib/modules/eic_groups/src/Controller/DiscussionController.php
```diff
@@ -135,6 +137,11 @@ public function addComment(Request $request, $discussion_id) {
         'value' => $text,
         'format' => 'plain_text',
       ],
+      'field_tagged_users' => array_map(function($tagged_user) {
```
lramarojaona, line 140: This variable isn't set anywhere `$tagged_user`. I guess this should be `$tagged_users`?

lib/themes/eic_community/react/components/Block/CommentsDiscussion/Partials/CommentForm.js
```diff
@@ -1,64 +1,167 @@
...
+                    />}
+                    modal
+                    nested
+                  >
+                    {close => (
+                      <div className="modal">
+                        <button className="close" onClick={close}>
+                          &times;
+                        </button>
+                        <div className="header">{"test"}</div>
```
lramarojaona, line 107: This shouldn't be here I guess

lib/themes/eic_community/react/components/Block/CommentsDiscussion/Partials/CommentForm.js
```diff
@@ -1,64 +1,167 @@
...
+                        <div className="content">
+                          <div className="ecl-comment-overview__form-wrapper">
+
+                            <div className="ecl-comment-form ">
+                              <div className="ecl-comment-form__author">
+                                <div className="ecl-author ecl-author--is-default  ecl-author--hide-info">
+                                  <div className="ecl-author__information">
+                              <span className="ecl-author__label ">
+                                    <a href="?path=user-john-doe"
```
lramarojaona, line 116: Do we need to show an author here?

lramarojaona (changes requested):
It works althoug the modal window doesn't seem complete to me (it's showing a test message + showing an author, is this needed?)

I added a comment in the ticket regarding emails, cause to me it seems useless if tagged users don't get an email, but it's maybe not in the scope of this ticket

lib/modules/eic_groups/src/Controller/DiscussionController.php
Dromadoros: It's an array_map so it's coming from tagged_users (second param of array_map)

Dromadoros:
> It works althoug the modal window doesn't seem complete to me (it's showing a test message + showing an author, is this needed?)
> 
> I added a comment in the ticket regarding emails, cause to me it seems useless if tagged users don't get an email, but it's maybe not in the scope of this ticket

I removed the author + test title. And I agree, we can see the email point in a discussion/ticket

lramarojaona (approved):

PR #946, opened 2021-11-29
Feature/eicnet 1783 organisation visibility

lramarojaona:
### Tests

- [x] Create 2 organisations: Organisation A and Organisation B
- [x] Change the visibility of Group 1 to Custom restriction => Specific organisations => Organisation A
- [x] Change the visibility of Group 2 to Custom restriction => Specific organisations => Organisation B
- [x] Add User 1 (TU) to Organisation A
- [x] Add User 2 (TU) to Organisation B
- [x] Login as User 1
- [x] Check that you can access Goup 1 detail page and not Group 2
- [x] Check that you see all content from Group 1 in all Solr views and no content from Group 2
- [x] Login as User 2
- [x] Check that you can access Goup 2 detail page and not Group 1
- [x] Check that you see all content from Group 2 in all Solr views and no content from Group 1

### Remarks

- The React entity_tree widget is not ready yet, it should be a sub-ticket

### DIFF
--- a/lib/modules/oec_group_flex/src/Plugin/CustomRestrictedVisibility/RestrictedUsers.php
+++ b/lib/modules/oec_group_flex/src/Plugin/CustomRestrictedVisibility/RestrictedUsers.php
@@ -31,6 +31,8 @@ class RestrictedUsers extends CustomRestrictedVisibilityBase {
     $form = parent::getPluginForm();
 
     // Specific trusted users text box.
+    // @todo Make use of the entity_tree widget once available for
+    //   organisations.
     $form[$this->getPluginId()][$this->getPluginId() . '_conf'] = [
       '#title' => $this->t('Select trusted users'),
       '#type' => 'entity_autocomplete',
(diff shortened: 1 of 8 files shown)


lib/modules/oec_group_flex/src/Plugin/CustomRestrictedVisibility/RestrictedUsers.php
```diff
@@ -31,6 +31,8 @@ public function getPluginForm(): array {
     $form = parent::getPluginForm();
 
     // Specific trusted users text box.
+    // @todo Make use of the entity_tree widget once available for
+    //   organisations.
```
Dromadoros, line 35: too much spaces before organisations

lramarojaona: I commented in the wrong file, this is fixed now

4hardfire (approved):
It works!

PR #949, opened 2021-11-30
[EICNET-1761]feat: Add first section of the footer menu

Dromadoros:
How to test it:

- [ ] Check if footer is showing like the design/slicing

image


lramarojaona:
@Dromadoros , I think this was working before, with an editable block, can you try this?
I guess they want to be able to manage this:
https://github.com/EIC-EA/EIC-community-D8/blob/d644642c13cc437d3f27853ce8cc0c98afbb2288/lib/modules/eic_default_content/content/block_content/34a810c9-d608-4979-ac40-00657bce344b.yml

korhankale (approved):

PR #958, opened 2021-12-01
EICNET-1808: User profile My Interests

4hardfire:
### Features

- Show topics of interest and topics of expertise when viewing user profile.
- Create helper functions in the theme to grab vocabulary terms in hierarchical tree.

### Tests

- [ ] Edit a user profile `/user/<uid>/member` and add some topics of interest and expertise
- [ ] Go to the user detail page and make sure there is a new tab "Interests" that shows the topics of interest and expertise selected by the user. Also make sure those are shown as a hierarchical tree.

### Notes

- [ ] This PR should be merged after #950

### DIFF
--- a/lib/themes/eic_community/includes/preprocess/common.inc
+++ b/lib/themes/eic_community/includes/preprocess/common.inc
@@ -15,6 +15,7 @@ use Drupal\eic_flags\FlagType;
 use Drupal\eic_groups\Constants\GroupVisibilityType;
 use Drupal\group\Entity\GroupInterface;
 use Drupal\node\NodeInterface;
+use Drupal\taxonomy\TermInterface;
 
 /**
  * Returns the build of the social share block.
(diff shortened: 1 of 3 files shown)


lib/themes/eic_community/includes/preprocess/common.inc
```diff
@@ -269,3 +270,54 @@ function _eic_community_get_entity_last_comment(EntityInterface $entity, $first_
...
+ *   The hierarchical tree of terms.
+ * @param array $loaded_parents
+ *   The array of loaded parents.
+ *
+ * @return array
+ *   Nested array of all loaded parents of the term.
+ */
+function _eic_community_process_term_parents_tree(TermInterface $term, array $terms, array $loaded_parents) {
+  $parent = array_shift($loaded_parents);
+  if (!isset($terms[$parent->id()])) {
```
korhankale, line 308: Shouldn't happen but it throws a 500 when the term doesn't have a parent (e.g: when you select "Horizontal" as topic"

korhankale (approved):

PR #959, opened 2021-12-01
EICNET-1739: Update member profile Topics of interest when following a topic

4hardfire:
### Improvements

- Implement logic in event subscriber so that when a user flags/unflags a topic it will update the topics of interest in the user profile accordingly.

### Tests

- [x] As TU, make sure you fill in your profile via `/user/<user-id>/member`
- [x] Go to a topic detail page and follow the topic
- [x] Go to your profile detail page `/user` and make sure the topic that you follow is shown in the "Interests" tab
- [x] Go to the topic detail page again and unfollow the topic
- [x] Go back to the profile detail page and make sure the topic is not shown anymore in the "Interests" tab.

### Notes

There is an issue when the user doesn't have a profile yet that's why we need to make sure the user fills in his profile when the creating a new account in the platform. A follow-up ticket has been created https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1834.


lib/modules/eic_user/src/EventSubscriber/FlagEventSubscriber.php
```diff
@@ -30,34 +51,125 @@ public static function getSubscribedEvents() {
...
+   * @param \Drupal\taxonomy\TermInterface $taxonomy_term
+   *   The topic term to add/remove from user profile.
+   * @param string $flag_event
+   *   The flag event. Either FlagEvents::ENTITY_FLAGGED or
+   *   FlagEvents::ENTITY_UNFLAGGED.
    */
-  public function invalidateUnflaggedEntityCache(UnflaggingEvent $event) {
-    $flaggings = $event->getFlaggings();
+  private function updateUserProfileTopics(ProfileInterface $profile, TermInterface $taxonomy_term, $flag_event) {
+    $vocab_field_name = 'field_vocab_topic_interest';
```
lramarojaona, line 135: Here we could probably introduce a new constant for this field name in `\Drupal\eic_topics\Constants\Topics`

And replace hardcoded `field_vocab_topic_interest` everywhere in the code.

lramarojaona: I guess we should also get rid of the `$vocab_field_name` variable and use the constant itself

lib/modules/eic_user/src/EventSubscriber/FlagEventSubscriber.php
```diff
@@ -30,34 +51,125 @@ public static function getSubscribedEvents() {
...
+    }
 
-      // Invalidate entity cache tags.
-      Cache::invalidateTags($flagging->getFlaggable()->getCacheTagsToInvalidate());
+    if (!$needs_update) {
+      return;
     }
+
+    $profile->$vocab_field_name = $topics;
+    $profile->save();
```
lramarojaona, line 172: Again, why not
```php
    if ($needs_update) {
      $profile->$vocab_field_name = $topics;
      $profile->save();
    }
```
:)

config/sync/user.role.trusted_user.yml
```diff
@@ -34,6 +34,7 @@ permissions:
   - 'post comments'
   - 'skip comment approval'
   - 'unflag recommend_group'
+  - 'update own member profile'
```
lramarojaona, line 37: I guess we should add this permission to all authenticated users.

lramarojaona (changes requested):

lib/modules/eic_user/src/EventSubscriber/FlagEventSubscriber.php
```diff
@@ -30,34 +51,125 @@ public static function getSubscribedEvents() {
...
+    switch ($flagging->getFlagId()) {
+      case FlagType::FOLLOW_USER:
+        $invalidate_cache = TRUE;
+        break;
+
+      case FlagType::FOLLOW_TAXONOMY_TERM:
+        /** @var \Drupal\taxonomy\TermInterface $taxonomy_term */
+        $taxonomy_term = $flagging->getFlaggable();
+
+        if ($taxonomy_term->bundle() !== 'topics') {
```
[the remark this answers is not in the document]
4hardfire: Done!

lib/modules/eic_user/src/EventSubscriber/FlagEventSubscriber.php
```diff
@@ -30,34 +51,125 @@ public static function getSubscribedEvents() {
...
+        break;
+
+    }
+
+    if (!$invalidate_cache) {
       return;
     }
 
-    // Invalidate entity cache tags.
+    $this->invalidateFlaggedEntityCacheTags($flagging);
```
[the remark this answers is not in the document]
4hardfire: Done!

lib/modules/eic_user/src/EventSubscriber/FlagEventSubscriber.php
4hardfire: @lramarojaona I created a new constant and replaced the hardcoded `field_vocab_topic_interest` in custom modules.
Concerning the variable `$vocab_field_name` I usually make it that way otherwise if we change the constant name in the future, there will be more places to replace. This is a kind of trick that I usually do. If you think it's useless and not needed at all I can do it the way you suggested.

lib/modules/eic_user/src/EventSubscriber/FlagEventSubscriber.php
4hardfire: Done!

lib/modules/eic_user/src/EventSubscriber/FlagEventSubscriber.php
lramarojaona: Yes, I've thought of that.

Normally a constant name should not change, but since we're still in development phase, your opinion is also true.
I just find a bit strange to put a constant in a variable.

As you wish I'd say :)

config/sync/user.role.trusted_user.yml
4hardfire: I added permission to view and edit own profile.

lramarojaona (approved):

PR #962, opened 2021-12-01
Feature/eicnet 1828 organisation type visibility

lramarojaona:
### Tests

- [x] Create 2 organisations: Organisation A and Organisation B with 2 different Organisation types
- [x] Change the visibility of Group 1 to Custom restriction => Specific organisation types => Type of Organisation A
- [x] Change the visibility of Group 2 to Custom restriction => Specific organisation types => Type of Organisation B
- [x] Add User 1 (TU) to Organisation A
- [x] Add User 2 (TU) to Organisation B
- [x] Login as User 1
- [x] Check that you can access Goup 1 detail page and not Group 2
- [x] Check that you see all content from Group 1 in all Solr views and no content from Group 2
- [x] Login as User 2
- [x] Check that you can access Goup 2 detail page and not Group 1
- [x] Check that you see all content from Group 2 in all Solr views and no content from Group 1


lib/modules/eic_groups/modules/eic_organisations/src/OrganisationsHelper.php
```diff
@@ -0,0 +1,108 @@
...
+      $group = $membership->getGroup();
+
+      // We only check organisations, if this is another bundle, just skip it.
+      if ($group->bundle() == Organisations::GROUP_ORGANISATION_BUNDLE) {
+        $organisations[] = $group;
+      }
+    }
+
+    return $organisations;
+  }
```
4hardfire, line 74: @lramarojaona in the future we should consider improving this method because I don't know how slow it can become if the user is a member of a lot of groups. Maybe a DB query would be better in this case.

4hardfire (approved):
It works!

lib/modules/eic_groups/modules/eic_organisations/src/OrganisationsHelper.php
lramarojaona: Yes, I was also thinking we can use a static cache for this result (memory cache per request, not persistent cache), and I'm also surprised the "groupMembershipLoader" doesn't provide a method where you can filter on group types

PR #970, opened 2021-12-06
Feature/eicnet 1782 group news overview

lramarojaona:
### Tests

- [x] Enable the News feature in an Organisation
- [x] As GO/GA/TU, check that you can access the overview through the main menu
- [x] As GO/GA/TU, check that you can create news items
- [x] Repeat the same for an Event (group type)

### DIFF
--- a/config/sync/context.context.group_news_overviews.yml
+++ b/config/sync/context.context.group_news_overviews.yml
@@ -1,0 +1,63 @@
+uuid: 05476799-affe-4640-899a-7b504fd59a42
+langcode: en
+status: true
+dependencies:
+  module:
+    - eic_search
+    - route_condition
+name: group_news_overviews
+label: 'Group - News - Overviews'
+group: Groups
+description: 'This context is used for all group news overviews'
+requireAllConditions: false
+disabled: false
+conditions:
+  route:
+    id: route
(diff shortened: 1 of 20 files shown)


4hardfire (changes requested):
Issues found:

- Trusted users can only create draft News. Currently is not possible to publish them as TU.
- After creating a News in group type "Event" as TU, I got an access denied when trying to access the news detail page.
- After creating a News (in both Organisation and Event groups) the overview list didn't show any results and all the facets disappeared (maybe I am missing something??).

lramarojaona:
> Issues found:
> 
>     * Trusted users can only create draft News. Currently is not possible to publish them as TU.
> 
>     * After creating a News in group type "Event" as TU, I got an access denied when trying to access the news detail page.
> 
>     * After creating a News (in both Organisation and Event groups) the overview list didn't show any results and all the facets disappeared (maybe I am missing something??).

I guess this is due to the fact that News & Stories have a specific workflow. TU should not be able to publish NEws directly (in global context). However indeed we seem to have an issue with News inside groups.

I'll create a ticket.

4hardfire (approved):
Remaining issues will be handled in follow-up tickets!

PR #973, opened 2021-12-06
EICNET-1835: User profile - My groups

4hardfire:
### Features

- Create new view of groups to show user profile groups;
- Create new theme suggestion and templated for the view of user profile groups;
- Show tab "My groups" in user profile page.

### Tests

- [x] As logged in user, become member of multiple groups
- [x] Go to the user profile page and make sure the groups you are a member are shown in the tab "My groups"

### Todo

- [x] Test cache issues.


lib/themes/eic_community/includes/preprocess/users/user.inc
```diff
@@ -515,3 +521,18 @@ function _eic_community_get_user_topics(UserInterface $account, $taxonomy_term_f
...
+ * Returns view output of user profile groups.
+ *
+ * @param \Drupal\user\UserInterface $account
+ *   The user account for which we want to get list of groups.
+ *
+ * @return array|null
+ *   A renderable array containing the view output or NULL if the display ID
+ *   of the view to be executed doesn't exist.
+ */
+function _eic_community_get_user_rendered_groups(UserInterface $account) {
```
lramarojaona, line 535: Why not create a generic function then?

```php
function _eic_community_get_rendered_view($view_id, $display_id, $args= [])
```

config/sync/views.view.groups.yml
```diff
@@ -0,0 +1,261 @@
...
+    display_options:
+      display_extenders: {  }
+      display_description: ''
+    cache_metadata:
+      max-age: -1
+      contexts:
+        - 'languages:language_content'
+        - 'languages:language_interface'
+        - url
+      tags: {  }
```
lramarojaona, line 261: We can probably improve with an no results text.
Something like this?
`You currently aren't member of any group. Start browsing the groups at /groups`

lramarojaona (changes requested):
Works well, please check the 2 comments

lib/themes/eic_community/includes/preprocess/users/user.inc
lramarojaona: That's nice, but if the display ID doesn't exist, what will happen?

config/sync/views.view.groups.yml
4hardfire: Done!

lramarojaona (changes requested):

lib/themes/eic_community/includes/preprocess/users/user.inc
4hardfire: It will return NULL.

lib/themes/eic_community/includes/preprocess/views.inc
```diff
@@ -155,3 +156,27 @@ function eic_community_preprocess_views_view(&$variables) {
...
+ *   of the view to be executed doesn't exist.
+ */
+function _eic_community_get_rendered_view($view_id, $display_id, array $args = []) {
+  $view = Views::getView($view_id);
+
+  if (!$view) {
+    return NULL;
+  }
+
+  return $view->preview($display_id, $args);
```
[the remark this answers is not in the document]
4hardfire: I can test that but usually it should return NULL if the display doesn't exist.

lib/themes/eic_community/includes/preprocess/users/user.inc
lramarojaona: Ok then

lramarojaona (approved):

lib/themes/eic_community/includes/preprocess/views.inc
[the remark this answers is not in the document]
lramarojaona: Indeed, it returns FALSE

PR #974, opened 2021-12-07
EICNET-1780: Create News & Stories and Events overview pages in default content

lramarojaona:
### Improvements

- There is a new MenuGenerator that will create the main menu links

### Tests

- [x] Run `drush fixtures:reload all`
- [ ] You should have now **News & Stories** and **Events** overview pages in _/admin/community/overview-pages_

### DIFF
--- a/lib/modules/eic_default_content/src/Generator/OverviewPageGenerator.php
+++ b/lib/modules/eic_default_content/src/Generator/OverviewPageGenerator.php
@@ -6,6 +6,7 @@ use Drupal\eic_events\Constants\Event;
 use Drupal\eic_overviews\Entity\OverviewPage;
 use Drupal\eic_overviews\GlobalOverviewPages;
 use Drupal\eic_search\Search\Sources\GlobalSourceType;
+use Drupal\eic_search\Search\Sources\GlobalEventSourceType;
 use Drupal\eic_search\Search\Sources\GroupSourceType;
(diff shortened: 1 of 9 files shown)


lib/modules/eic_default_content/src/Generator/OverviewPageGenerator.php
```diff
@@ -37,16 +39,41 @@ public function load() {
...
+
+    $this->createOverview([
+      'enable_search' => TRUE,
+      'facets' => [
+        'ss_group_topic_name',
+        'sm_group_field_location_type',
+        Event::SOLR_FIELD_ID_WEIGHT_STATE_LABEL,
+        'ss_group_event_country',
+      ],
+      'source_type' => GlobalEventSourceType::class,
```
korhankale, line 73: The namespace of this source type isn't defined. This is resulting in an error when you query the overview as it uses a relative one instead of the correct one
Current:
image
Should be:
Drupal\eic_search\Search\Sources\GlobalEventSourceType

korhankale (approved):

lib/modules/eic_default_content/src/Generator/OverviewPageGenerator.php
lramarojaona: Good catch, this is fixed

PR #977, opened 2021-12-07
EICNET-1839: Private images re-used in homepage banner are not shown for anonymous

4hardfire:
### Improvements

- Add custom_block and group entities as entity usage source so that anonymous users can view private images that are attached to custom_blocks and groups.

### Fixes

- Create and apply contrib patch that fixes endless loop during batch operation to update all entity usages (see https://www.drupal.org/project/entity_usage/issues/3168418#comment-14331779).

### Tests

- [x] Go to `/admin/config/entity-usage/batch-update` and click on **Recreate all entity usage statistics**
- [x] Make sure the batch doesn't end up in an endless loop
- [x] Clear all cache
- [x] As anonymous user, navigate to the frontpage and make sure you can view the Banner image
- [x] As anonymous user, navigate to groups overview page `/groups` and make sure you can view the group banner images

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3144,6 +3144,9 @@
                     "services": {
                         "drush.services.yml": "^9"
                     }
+                },
+                "patches_applied": {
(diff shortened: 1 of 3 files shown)


lramarojaona (approved):
Works fine!

PR #980, opened 2021-12-08
[EICNET] Teaser elements add click on empty state image

Askerx9:
Normally now we have a click on all empty state's teaser elements (twig and react)


filipvt1 (changes requested):
There's still a dump present in Group teaser.
What about attachments? This is not using the teaser view, but maybe the icon should also be clickable?
For Video/Filelist/Gallery the teaser image isn't clickable

filipvt1 (approved):

PR #982, opened 2021-12-08
[EICNET] correct width on mobile for .ecl-editorial-article__content

Askerx9:
### before
image

### now 
image


filipvt1 (changes requested):
Don't think it is necessary to add extra css here. It seems to me that the width: 66.66% on the article content doesn't do anything when you get to the desktop view (as flexbox is used and it has a flex-basis of 100%, so it flows to full width there. Might just be easier to remove that width. That way you can remove the media query for the "sm" breakpoint.

filipvt1 (approved):

PR #985, opened 2021-12-08
[EICNET] Groups tables change view on mobile

Askerx9:
### BEFORE
image

### NOW
image


filipvt1 (changes requested):
The table view switches at a resolution of 995px. But the list-style on the ul is removed at 768px. Which means that between 768px and 995px, no bullet is shown next to the role.

Askerx9:
done @filipvt1

filipvt1 (approved):

PR #987, opened 2021-12-10
Feature/eicnet 1860 data export [Part 1]

lramarojaona:
### Improvements

This PR provides following views field plugins to the `oec_group_flex` module:

- Visibility
- Visibility Custom: Restricted email domains
- Visibility Custom: Restricted users

Also following views field plugins to the `eic_groups` module:

- Visibility Custom: Restricted organisations
- Visibility Custom: Restricted organisation types
- Group members

### Tests

- [x] Create a group based view
- [x] Make sure you have one group based on restricted domains, one with restricted users, one with restricted organisations and one with restricted organisation types
- [x] Add all the mentioned fields
- [x] Check that the view outputs correct values for all the fields

### Remarks

This PR doesn't provide in any way a data export, this is just a first chunk.


4hardfire (approved):
Seems ok to me!

PR #997, opened 2021-12-15
[EICNET-1820]chore: use static method to generate field entity tree settings

Dromadoros:


lib/modules/eic_content/src/Plugin/Field/FieldWidget/EntityTreeWidget.php
```diff
@@ -230,6 +233,9 @@ public static function getEntityTreeFieldStructure (
     string $target_type,
     string $preselected_terms,
     int $number_to_load,
+    string $base_url,
+    string $base_url_search,
+    string $base_url_search_children,
```
lramarojaona, line 238: @Dromadoros , 
Is it worth to re-add this? It's not being used.

lramarojaona (changes requested):

lramarojaona:
@Dromadoros , Could you also provide a working example in a README file as how to use this form element in a custom form?
Just tried a bit, but I'm getting the following error:

`InvalidArgumentException: The #default_value property has to be an entity object or an array of entity objects. in Drupal\Core\Entity\Element\EntityAutocomplete::valueCallback() (line 114 of core/lib/Drupal/Core/Entity/Element/EntityAutocomplete.php). `

Thanks!

lenglebert:
What's the status on this PR? @Dromadoros

PR #1001, opened 2021-12-16
Feature/eicnet group wiki detailed rework

Askerx9:
### What I did

- [ ] nav list is expended by default
- [ ] On mobile the sidebar menu should be expanded vertically

### What I need

- [x] The wording of the buttons is incorrect
- [x] Management buttons are not on the same line
- [x] Missing request delete and archival action in the manage button
- [x] Add extra class on navigation list container to remove the padding on mobile (.ecl-editorial-article__aside)

### DIFF
--- a/lib/themes/eic_community/js/components/drupal.navigation-list.js
+++ b/lib/themes/eic_community/js/components/drupal.navigation-list.js
@@ -140,7 +140,7 @@
     });
 
     if (readSession('collapsed')) {
-      navigationList.setAttribute('aria-collapsed', 'true');
+      navigationList.setAttribute('aria-collapsed', 'false');
       collapseParentElement(navigationList);
     }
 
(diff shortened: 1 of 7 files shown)


lenglebert:
Back-end work needs to be done: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1871

lib/themes/eic_community/js/components/drupal.navigation-list.js
```diff
@@ -140,7 +140,7 @@
     });
 
     if (readSession('collapsed')) {
```
filipvt1, line 142: What is the definition of "expanded by default"? Does this mean that on every page load it is expanded again? In that case, what's the point of storing it in the session? If it is on the initial pageload then I think this implementation isn't correct. Because now it is forcing it to be expanded every time you reload the page or you move to another page.

filipvt1 (changes requested):
On mobile there still seems to be some extra padding being added to the ".ecl-editorial-article__aside" item. Check why this padding is there and if not necessary, remove it. Otherwise overwrite it for the wiki page.

filipvt1 (approved):

PR #1005, opened 2021-12-17
EICNET-1858: Group ownership transfer confirmation (Rework)

4hardfire:
### Improvements

- GO transfer confirmation: don't provide a reason for accepting the ownership transfer
- GO transfer confirmation: owner should only have owner role and not both admin and owner
- GO transfer confirmation: add pending status
- On deny, redirect to user to the group homepage
- New timeout field in request flags to automatically close a request
- Owner checkbox removed from the membership edit form (Jira ticket -> https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1840)

### Todo

- [x] Send an email to tell the user when request has reached the timeout
- [x] Send email to tell the GO when request has reached the timeout
- [ ] When GO tries to access the accept/deny form of a request that reached the timeout limit, the user should see a message "The invitation has expired"

### Tests

- [x] Follow the same tests from PR #994 

### Test improvements

- [x] Make sure there is no field to provide a reason when accepting ownership transfer
- [x] Make sure the new owner doesn't have the group admin role after accepting ownership transfer
- [x] When requesting ownership transfer to a group admin, when viewing the list of group members `/groups/<group-id>/members` make sure there is a text "pending ownership transfer" is shown bellow the username of the user that is being requested
- [x] When the new GO accepts/denies a request, if the destination is not presented in the URL make sure the user gets redirected to the group homepage. The same behaviour should happen if the user clicks on "Cancel".
- [x] When requesting a new ownership transfer, make sure there is a new field to provide the timeout for the request to be automatically closed
- [x] Try to edit the owner membership by clicking on the "Edit member" link and make sure the "Owner" option is not presented in the list of group roles to choose.
- [x] Save the membership and make sure the user still has the owner role


lramarojaona (approved):

lramarojaona:
Looks good, but I have few remarks/suggestions:

- Request confirmation (accept or deny) should be allowed only for the targeted user? (for now SA can also accept/deny)
- Timeout field: provide preselected options instead of number field? (No timeout/2 days/5 days/10 days/30 days)
- Message template: inform the user about the timeout?
- Members overview: display time left before expiration?

PR #1007, opened 2021-12-24
EICNET-1711: Topic Detail FE.

filipvt1:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1711
### What has been fixed:
**General**
- [x] Latest member activity should be "member activity" > In design it is "Latest activity", updated to that value
- [x] Grey used in design (statistics for example) is darker than currently implemented.	

**Subtopics**
- [x] The subtopics show a line (link), design does not show this.
- [x] The subtopics have a smaller font inside the badges on design

**Latest activity**
- [x] Space between event-icon & user who created this event.
- [x] Contrast is very low (does not pass). Design also lists a darker gray for this text.

**Related groups**
- [x] Groups are not fixed height (this is the case on list with all groups)

**Related stories**
- [x] Design does not have a gray background

### Remarks:
Spacing between the headings displayed on the right (between subtopics - experts)**
> There are some additional wrapper divs around the "contributors" block which are causing the CSS to not be applied correctly. Can they be removed?

Spacing between the two columns used (Marketing - subtopics)
> We use the same spacing on other pages now. Check with Lucas based on feedback whether we should keep this spacing or increase it.

**Experts**
- The experts only have an email on design. Implementation also shows social buttons
	The icon is also showing on the right side (not below the adres)
- The date of joining is missing 
- Profile (avatar) looks big compared to design
> Join date needs to be added in Storybook. 
> Implementation is not correct, uses "member" node html while in storybook it uses the "author" html.
> Regarding the icon, this is something that was generalized based on other pages/components to be more consistent.


Askerx9:
### What has been fixed:

**General**
- [ ] I have a feeling that a lot of the images (teasers & also banners on other pages) gets compressed hard and have a lot of artifacts due to low quality. This is very noticeable on mobile because the image is quite large.

**Breadcrumbs**
- [ ] Breadcrumb bar is above the header, it should be below on topics & subtopics

**Latest activity**
- [ ] Design also lists the what kind of item is displayed

**Related groups**
- [x] Change "View all groups" to "Show more"

**Related news & stories**
- [x] Date in the upper right corner
- [ ] Statistics are not displayed

**Related discussions**
- [x] Everything needs to be aligned left (same as title + description).
- [x] Missing the text "started by" before the author
- [x] The comment box does not stretch the full width on design

**Related files**
- [x] Design shows alternating background-color (white - gray)
- [ ] Like button is missing
- [ ] I tried adding a video, but it does not display in the stream
- [ ] Missing "show more"

**Related events**
- [ ] Remote/location is missing
- [ ] Time left to register


### Remarks:
**General**

_I have a feeling that a lot of the images (teasers & also banners on other pages) gets compressed hard and have a lot of artifacts due to low quality. This is very noticeable on mobile because the image is quite large._
> We need to check with the backend why the image it sends is 180px/180px.

**Breadcrumbs**

_Breadcrumb bar is above the header, it should be below on topics & subtopics_
> done by Bernardo on other PR

**Latest activity**

_Design also lists the what kind of item is displayed_
> Need to be done in the BE I think.

**Related news & stories**

_Statistics are not displayed_
> Because we don't receive stats in the data

**Related files**

_Missing "show more"_
> Could we check with a BE

_Like button is missing_
> Could we check with a BE

_I tried adding a video, but it does not display in the stream_
> Could we check with a BE

**Related events**
problem with event 
image

Dromadoros (approved):
Ok for BE

PR #1013, opened 2022-01-03
[EICNET-1608]feat: Set big number on maxlength for selecting users in group

Dromadoros:
How to test:

- [ ] Go to a group contact page (/groups/group-1/contact)
- [ ] Try to add multiple recipients 
- [ ] Be sure you can add multiple recipients without having any limitations issues

### DIFF
--- a/lib/modules/eic_private_message/src/Form/PrivateMessageForm.php
+++ b/lib/modules/eic_private_message/src/Form/PrivateMessageForm.php
@@ -134,7 +134,7 @@ class PrivateMessageForm extends FormBase {
         '#title' => $this->t('Select users in group', [], ['context' => 'eic_private_message']),
         '#description' => $this->t('Seperate users by ",".', [], ['context' => 'eic_private_message']),
         '#tags' => TRUE,
-        '#maxlength' => 5000,
+        '#maxlength' => NULL,
         '#target_type' => 'group_content',
         '#selection_handler' => 'default:group_membership',
         '#selection_settings' => [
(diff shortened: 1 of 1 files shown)


lib/modules/eic_private_message/src/Form/PrivateMessageForm.php
```diff
@@ -134,6 +134,7 @@ public function buildForm(array $form, FormStateInterface $form_state) {
         '#title' => $this->t('Select users in group', [], ['context' => 'eic_private_message']),
         '#description' => $this->t('Seperate users by ",".', [], ['context' => 'eic_private_message']),
         '#tags' => TRUE,
+        '#maxlength' => 5000,
```
4hardfire, line 137: @Dromadoros I had the same problem for the custom restrictions by user when creating a custom restricted group and used the following `'#maxlength' => NULL,` which it worked. Maybe you should do the same to avoid any limitation.

4hardfire (changes requested):

lib/modules/eic_private_message/src/Form/PrivateMessageForm.php
4hardfire: You can check the plugin class `RestrictedUsers.php` in oec_group_flex module

Dromadoros: Done!

4hardfire (approved):
Seems ok to me!

PR #1016, opened 2022-01-03
[EICNET-1101]feat: Handle likes differently when indexing

Dromadoros:
How to test it:

- [x] Re-index all elements in solr
- [x] go to group library overview
- [x] try to like one/multiple item(s)
- [x] on the overview, sort by likes and be sure you have correct/logic numbers


4hardfire (approved):
It works!

PR #1020, opened 2022-01-04
Feature/eicnet 1860 data export 2

lramarojaona:
### Improvements

- Introduced a Group Metrics views field, which is based on GroupMetric plugins

### Tests

- [ ] Create a Group based view
- [x] Make sure you have content for: group content, members, memberhsip requests, invitations, flags, downloads, comments inside your groups
- [ ] Add a `Group: Group Metrics` field for each of the plugins and configure them to test them as you wish: `Group flags`, `Group invitations`, `Group members`, `Group comments`, `Group downloads`, `Group membership requests`, `Group content`, `Group contributors`
- [ ] Make sure you have relevant/correct counters


lib/modules/eic_groups/eic_groups.module
```diff
@@ -115,6 +115,13 @@ function eic_groups_views_data() {
       'id' => 'group_members_by_role',
     ],
   ];
+  $data['groups_field_data']['group_metrics'] = [
+    'title' => t('Group metrics'),
+    'help' => t('Displays the group metrics.'),
+    'field' => [
+      'id' => 'group_metrics',
+    ],
+  ];
```
4hardfire, line 124: @lramarojaona why do we have the same implementation in `eic_group_statistics.module`?

4hardfire (changes requested):

4hardfire:
@lramarojaona the group contributors always show 0. I don't know if this is an issue or not. Is there a way I can test it?

lib/modules/eic_groups/eic_groups.module
lramarojaona: Oh sorry indeed, you're right, this is a leftover, just removed it.

lramarojaona:
> @lramarojaona the group contributors always show 0. I don't know if this is an issue or not. Is there a way I can test it?



> @lramarojaona the group contributors always show 0. I don't know if this is an issue or not. Is there a way I can test it?

This counter should show active members that own published nodes. Do you have that?

4hardfire (changes requested):
Group contributors keeps showing 0 even though I created new content in a published group as TU.

PR #1021, opened 2022-01-04
[EICNET-1590]feat: Add correct key for "changed" value in sort

Dromadoros:
How to test:

- [x] Go to a group library overview
- [x] Verify that the second default sort is by updated date of the document node
- [x] You can also force the first sort by the dropdown


4hardfire (approved):
Seems ok to me!

PR #1022, opened 2022-01-04
EICNET-1725: Error when adding same topic of interest multiple times in user profile

4hardfire:
### Improvements

- Apply entity tree widget in user profile interests.

### Tests

- [ ] Try to update a member profile and make sure the entity tree widget is presented in the **Topics of expertise** and **Topics of interest** fields.

### DIFF
--- a/config/sync/core.entity_form_display.profile.member.default.yml
+++ b/config/sync/core.entity_form_display.profile.member.default.yml
@@ -16,6 +16,7 @@ dependencies:
     - profile.type.member
   module:
     - address
+    - eic_content
     - link
     - social_link_field
     - text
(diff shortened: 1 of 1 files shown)


config/sync/core.entity_form_display.profile.member.default.yml
```diff
@@ -87,22 +88,26 @@ content:
...
     weight: 6
     settings:
-      match_operator: CONTAINS
-      match_limit: 10
-      size: 60
-      placeholder: ''
+      match_top_level_limit: '0'
+      disable_top_choices: '1'
+      items_to_load: '25'
+      auto_select_parents: '1'
```
Dromadoros, line 106: I think we need to disable this in your case ? Do you want to auto select all parents ?

4hardfire: That was a mistake indeed. It should be fixed!

Dromadoros (approved):

PR #1023, opened 2022-01-04
[EICNET-1474]fix: Be able to search in fullname and not only first or last name

Dromadoros:
- [x] Edit a user and go to the field "delete by"
- [x] Select one user by his first AND last name, you should see a result


4hardfire (approved):
Seems ok to me!

PR #1024, opened 2022-01-05
[EICNET-1897] Group search - Large font size of group title

Askerx9:
### How to test

1. got to your favorit group
2. make an empty search
3. put an H2 inside ".ecl-teaser__description"

### Before
image

### Now
image


filipvt1 (changes requested):
If the teaser description can contain html tags, it shouldn't be a p tag. This is invalid HTML. Can you convert the paragraph to a div? Check that styling is still correct with that new tag.

filipvt1 (approved):

PR #1027, opened 2022-01-05
[EICNET-1643] responsive form improvement

Askerx9:
https://citnet.tech.ec.europa.eu/CITnet/jira/secure/RapidBoard.jspa?rapidView=6628&projectKey=EICNET&view=detail&selectedIssue=EICNET-1643&quickFilter=26822

### What I did
image

### DIFF
--- a/lib/themes/eic_community/sass/components/_form.scss
+++ b/lib/themes/eic_community/sass/components/_form.scss
@@ -3,11 +3,6 @@
   width: 100%;
   margin-left: auto;
   margin-right: auto;
-  @include ecl-media-breakpoint-down('md') {
-    padding-left: 1rem;
-    padding-right: 1rem;
-    max-width: calc(100% - 2rem);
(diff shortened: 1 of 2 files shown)


lib/themes/eic_community/sass/components/_form.scss
```diff
@@ -3,6 +3,11 @@
   width: 100%;
   margin-left: auto;
   margin-right: auto;
+  @include ecl-media-breakpoint-down('md') {
```
filipvt1, line 6: I would actually just remove this styling and see if you can add the `<div class="ecl-container">` element around the form. This element is used in other components to add spacing and to determine the width of the container. Without that element the form doesn't have the correct width between a screen width of 480px and 995px.

filipvt1 (changes requested):

filipvt1 (approved):

PR #1028, opened 2022-01-05
[EICNET-1899]feat: Create header for organisation + refactor the way we get global overview page link + breadcrumb

Dromadoros:
How to test it:

- [ ] Create an organisation
- [ ] Go to the detail and check if breadcrumbs + header info are correct
- [ ] Be sure you have an overview page content with type "organisations"

### DIFF
--- a/config/sync/core.entity_form_display.group.organisation.default.yml
+++ b/config/sync/core.entity_form_display.group.organisation.default.yml
@@ -3,12 +3,16 @@ langcode: en
 status: true
 dependencies:
   config:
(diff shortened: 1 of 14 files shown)


4hardfire (changes requested):
@Dromadoros the last step is not ok. I don't have an overview page of organisations on my local. I reloaded the fixtures and still don't have any new overview page. Also I have a link in the breadcrumb "Overview" which in my opinion should be called "Organisations" but maybe this is an issue because of the missing overview page.

PR #1030, opened 2022-01-06
EICNET-1860: Group latest activity

lramarojaona:
### Improvements

- Provide logic to calculate group latest activity, with cache
- Provide a views field plugin to display the value

### Tests

- [ ] Create a test view based on group entity
- [ ] Add the `Group latest activity` field to the view (display also seconds for better testing)
- [x] Add a content to a group
- [x] Add a second content to a group => check that the field is updated with the new date
- [x] Add a comment to a content of the group => check that the field is updated with the new date
- [x] Delete the last content created in the group => check that the field is updated with the new date (should be older)

### Remarks

- The date should also be updated when deleting a comment, but since we don't hard-delete comments, it doesn't apply

### DIFF
--- a/lib/modules/eic_comments/src/Constants/Comments.php
+++ b/lib/modules/eic_comments/src/Constants/Comments.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Drupal\eic_comments\Constants;
+
+/**
+ * Defines constants for comments stuff.
+ *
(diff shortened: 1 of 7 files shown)


4hardfire (approved):
Code seems good to me!
Tested in PR #1038 and it shows only the date without the time. However, I tried to modify the format to show also the time and it's being updated properly. So I will consider this one as passed.

PR #1033, opened 2022-01-07
EICNET-1649: Display local tasks in a select field (rework)

4hardfire:
### Improvements

- Fix access to invite users when user is not a member of a group and also depending on the group status.
- Allow SA/SCM users to update book outlines.

### Tests

Invite members:

- [x] As GO of a public group, edit the group and enable the option "Invite members"
- [x] As TU and not member of the group, navigate to the group page and make sure the link to invite member is not visible and you cannot access the form via URL
- [x] Join the group, and make sure you can now see the link to invite a member in the group header 
- [x] As SA/SCM, send the group back to the pending status
- [x] As GO, make sure the link to invite member is not shown and you cannot access the form to invite a member via URL
- [x] As SA/SCM, change the group status to draft
- [x] As GO, make sure you can now see the link to invite a member in the group header and you can access the form to invite a member 

Book outline

- [x] As SA/SCM, navigate to a wiki page of a group
- [x] Make sure you can see the link "Outline" via the "Manage content" dropdown

### DIFF
--- a/lib/modules/eic_groups/src/Plugin/GroupContentEnabler/GroupInvitation.php
+++ b/lib/modules/eic_groups/src/Plugin/GroupContentEnabler/GroupInvitation.php
@@ -36,11 +36,34 @@ class GroupInvitation extends GroupInvitationBase {
    * {@inheritdoc}
    */
   public function createAccess(GroupInterface $group, AccountInterface $account) {
+    /** @var \Drupal\Core\Access\AccessResult $access */
     $access = parent::createAccess($group, $account);
+    $is_power_user = UserHelper::isPowerUser($account);
+    $is_group_admin = EICGroupsHelper::userIsGroupAdmin($group, $account);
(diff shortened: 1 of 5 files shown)


lib/modules/eic_groups/src/Plugin/GroupContentEnabler/GroupInvitation.php
```diff
@@ -36,13 +36,36 @@ public function getGroupOperations(GroupInterface $group) {
...
     }
 
+    // Adds default cacheable dependencies.
+    $access->addCacheableDependency($account)
+      ->addCacheableDependency($group);
+
+    // If access is allowed by default, we need to make sure the option to
+    // invite users is enabled and also if the user is a poweruser or a
+    // member of the group. Otherwise we force access denied as default before
+    // moving further to the next conditions.
```
lramarojaona, line 58: Nice to add such comments :)

lramarojaona (approved):
Works!

PR #1035, opened 2022-01-07
EICNET-1860: Group members data export

lramarojaona:
### Improvements

- Views plugins for member latest activity, content count and comment count

### Tests

- [ ] Create a test view based on user entity and add the membership relationship, Add parent group as argument
- [ ] Add `Member latest activity`, `Member content count` and `Member comment count` fields
- [ ] Check that you have correct values
- [ ] Try to add/remove content and comments and check that values are updated accordingly

### DIFF
--- a/lib/modules/eic_statistics/modules/eic_group_statistics/src/Plugin/views/field/GroupMemberLatestActivity.php
+++ b/lib/modules/eic_statistics/modules/eic_group_statistics/src/Plugin/views/field/GroupMemberLatestActivity.php
@@ -1,0 +1,82 @@
+<?php
+
+namespace Drupal\eic_group_statistics\Plugin\views\field;
+
+use Drupal\group\Entity\GroupContentInterface;
+use Drupal\views\Plugin\views\field\Date;
+use Drupal\views\ResultRow;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * A handler to provide a field for group member latest activity.
+ *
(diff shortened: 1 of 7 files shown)


lib/modules/eic_statistics/modules/eic_group_statistics/src/Plugin/views/field/GroupMemberLatestActivity.php
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        // We get the first encountered membership.
+        if ($relationship_entity->getGroupContentType()->getContentPluginId() == 'group_membership') {
+          $membership = $relationship_entity;
+          break;
+        }
+      }
+
+    }
```
4hardfire, line 71: @lramarojaona can't we just return the membership using `getMembership` method from `class GroupMemberHelper`?

4hardfire (changes requested):

lib/modules/eic_statistics/modules/eic_group_statistics/src/Plugin/views/field/GroupMemberLatestActivity.php
4hardfire: In fact I've checked the code in PR #1038 and you already did that change.

4hardfire (approved):
When we soft delete a comment, the counter doesn't take that into account. I believe this improvement can be done in a follow-up ticket. The rest seems ok to me!

PR #1036, opened 2022-01-07
EICNET-1369: EICNET-1369 Adapt comment section on video detail page (rework)

4hardfire:
### Improvements

- Show users who commented nodes as contributors next to the comment section;
- Change comment section title to "Replies"
- Fix cache issues when showing the comment avatar after user logs in with a different account;
- Fix coding standards.

### Tests

- [x] As TU/GM, go to a discussion/document/video detail page where comments are enabled and check if the title of the section is "Replies" instead of "Comments"
- [x] Place a new comment
- [x] Refresh the page and make sure the user avatar is shown in the contributors next to the comment section
- [x] Login as a different GM, go to the discussion/document/video detail page and make sure your avatar is shown next to the comment textbox and not the avatar from the previous logged in GM
- [x] As SA/SCM, remove the comment (soft delete) and make sure the TU/GM is not shown anymore in the list of contributors after refreshing the page


lramarojaona (approved):
Works well, just don't know if we should replace comments with replies in the sentence

PR #1038, opened 2022-01-07
Feature/eicnet 1860 views data export

lramarojaona:
### Improvements

- Provide ContentMetrics view field plugin

### Tests

- [ ] Go to `/admin/community/groups/groups`, click Export, Check that you get an Excel export
- [ ] Go to `/admin/community/groups/events`, click Export, Check that you get an Excel export

### DIFF
--- a/lib/modules/eic_content/src/Plugin/views/field/ContentMetrics.php
+++ b/lib/modules/eic_content/src/Plugin/views/field/ContentMetrics.php
@@ -1,0 +1,196 @@
+<?php
+
+namespace Drupal\eic_content\Plugin\views\field;
+
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\eic_content\ContentMetricPluginManager;
+use Drupal\views\Plugin\views\field\NumericField;
+use Drupal\views\ResultRow;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * A handler to provide a field to display content metrics.
+ *
+ * @ingroup views_field_handlers
+ *
+ * @ViewsField("content_metrics")
+ */
+class ContentMetrics extends NumericField {
+
+  /**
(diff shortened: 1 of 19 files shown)


lib/modules/eic_content/src/Plugin/views/field/ContentMetrics.php
```diff
@@ -0,0 +1,196 @@
...
+    $metric_id = $this->options['metric'];
+
+    if (empty($this->metricsInfo[$metric_id])) {
+      return 0;
+    }
+
+    $plugin = $this->metricsInfo[$metric_id];
+
+    $conf = $this->options[$metric_id . '_conf'] ?? [];
+    return $plugin->getValue($group, $conf);
```
4hardfire, line 193: The plugin is expecting a node entity. So I would change the variable name from `$group` to `$node`.

lib/modules/eic_content/src/Plugin/views/field/ContentMetrics.php
```diff
@@ -0,0 +1,196 @@
...
+   */
+  public function query() {
+    // Do nothing -- to override the parent query.
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getValue(ResultRow $values, $field = NULL) {
+    $group = $values->_entity;
```
4hardfire, line 183: Change the variable name from $group to $node.

4hardfire (changes requested):

lib/modules/eic_content/src/Plugin/views/field/ContentMetrics.php
lramarojaona: It is fixed

lib/modules/eic_content/src/Plugin/views/field/ContentMetrics.php
lramarojaona: It is fixed

4hardfire (approved):
Seems good to me!

PR #1042, opened 2022-01-11
[EICNET-1837] Not Authorized shown after TU leaves private group

korhankale:
## To Test

- [x] Join a private group as TU
- [x] Leave it
- [x] You should be redirected to the groups overview with a message


4hardfire (approved):
Seems ok to me!

PR #1050, opened 2022-01-13
EICNET-1689: Homepage last updated groups (rework 2)

4hardfire:
### Bug fixes

- Change groups homepage view cache to tag based per user because group visibility can be changed and there is an issue where anonymous users can see private groups in the list.

### Tests

- [x] As anonymous, go to the homepage and make sure you see only public groups
- [x] Login as SA/SCM
- [x] Change one of the public groups that shows in the homepage for anonymous users. Change the group to private.
- [x] Log out and navigate to the homepage
- [x] Make sure the private group is not shown in the homepage

### DIFF
--- a/config/sync/views.view.groups_homepage.yml
+++ b/config/sync/views.view.groups_homepage.yml
@@ -5,6 +5,7 @@ dependencies:
   config:
     - search_api.index.global
   module:
+    - eic_search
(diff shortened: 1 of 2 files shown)


lramarojaona (approved):
It works!

PR #1051, opened 2022-01-13
EICNET-1927: TU cannot create draft from published story

4hardfire:
### Fixes

- Apply patch to fix issue where user cannot save draft version of a published story because of book privileges.
- Allow content owners to access latest revision.

### Tests

- [x] As TU, create a new story
- [x] As SA/SCM, publish the story
- [x] As TU, edit the story and create a new draft
- [x] Check that you have access to the latest revision /stories/story-name/latest


lramarojaona (approved):
Works!

PR #1052, opened 2022-01-13
[EICNET]chore: Refresh configs according to the latest core version

korhankale:
This PR doesn't bring any new changes, it only refreshes the configs according to the latest core version
## To Test

- [x] Run `make setup`
- [x] Run some monkey testing on the application (especially on critical parts)


4hardfire (approved):
Seems ok to me!

PR #1053, opened 2022-01-13
EICNET-1697: Show non published groups to power users. (rework)

lramarojaona:
### Tests

- [x] As TU Create a group, don't publish it (Draft or Pending)
- [x] AS GO you should see the unpublished group in `/groups`
- [x] Log in SA/SCM/DA: you should also see the unpublished group in `/groups`


4hardfire (approved):
Seems ok to me!

PR #1054, opened 2022-01-13
EICNET-1754: Footer site name

4hardfire:
### Fixes

- Add footer site info block to the footer section.

### Tests

- [x] Go to the homepage
- [x] Make sure the first section in the footer shows the **Footer - Site info** (from `/admin/content/block-content`)

### Todo:

- [x] The description html tags are shown as plain text and this needs to be fixed in the template (`site-footer.html.twig`)


lramarojaona (approved):
Works!

PR #1057, opened 2022-01-14
EICNET-1937: Custom restriction visibility options not shown in group add form

4hardfire:
### Fixes

- Fix issue where custom restriction visibility options were not shown in group add form;
- Fix error when editing a group without visibility.

### Tests

- [x] As SA/SCM, try to create a group and select the custom restricted visibility
- [x] Make sure the custom restricted options are shown in the form


lramarojaona (approved):
It works

PR #1067, opened 2022-01-17
EICNET-1920: Fix breadcrumb for News and Stories.

lramarojaona:
### Tests

- [x] Go on a Story detail page, check that the breadcrumb shows "News & Stories" link which leads to `/articles`
- [ ] Go on a global News detail page (not in a group), check that the breadcrumb shows "News & Stories" link which leads to `/articles`


Dromadoros:
We just need to be sure that the overview page content has the correct name but that's not an issue

Dromadoros (approved):

PR #1072, opened 2022-01-18
[EICNET-1335] add tooltip on highlight icon

Askerx9:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1335?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel

### What I did

- add tooltip library
- add style on tooltip
- add translation for tooltip

### Example
image


lib/themes/eic_community/sass/components/_highlight.scss
```diff
@@ -47,4 +47,13 @@
   &__label {
     @include visually-hidden;
   }
+
+  &__tooltip {
```
filipvt1, line 51: Isn't it possible to pass the background and border color as attributes to the react tooltip component? That way you won't have to override the background with !important

filipvt1 (changes requested):

filipvt1 (approved):

PR #1076, opened 2022-01-19
[EICNET-1780]feat: add filter private + sort created

Dromadoros:
How to test:

- [ ] Create/edit news and story overview page
image

- [ ] Be sure every facets and sort working well


4hardfire (changes requested):
Everything works fine except the list doesn't show private stories. Is this normal?

Dromadoros:
> Everything works fine except the list doesn't show private stories. Is this normal?

Ok, it was link due to the fact that you were an SCM but not tagged as trusted_user also. I added the poweruser check also. Should work now

4hardfire (approved):
Seems ok to me!

PR #1079, opened 2022-01-20
EICNET-1975: Adapt URL for adding a group/event on homepage

4hardfire:
### Fixes

- Fix URLs to add a group or event in the homepage actions.
- Fix coding standards.

### Tests

- [x] As TU/SA/SCM, navigate to the homepage
- [x] Make sure the links to add a group and event redirects you to the correct page (`/group/add/group` or `/group/add/event`)


lib/modules/eic_content/src/Plugin/Block/ContentOperationsBlock.php
```diff
@@ -122,7 +134,9 @@ public function build() {
     $build = [];
     $supported_entities = [
       'group' => [
-        'add_route' => 'entity.group.add_page',
+        'add_route' => function ($entity, $bundle) {
```
Dromadoros, line 137: Now with PHP7.4 we can do this:

`fn($entity, $bundle) => Url::fromRoute('entity.group.add_form', ['group_type' => $bundle]);`

But I think not ready yet for different environment so let's wait

Dromadoros (approved):

PR #1082, opened 2022-01-20
EICNET-1974: Fix Group feature permissions issue

lramarojaona:
### Tests

- [x] As GO (non SA/SCM/DA), create an event
- [x] After event creattion, edit the event and enable News feature
- [x] Make sure the menu item is visible, that you can create a News and access the News overview
- [x] Edit the event again to disable the News feature
- [x] Make sure the menu item is not visible, that you cannot create a News and can't access the News overview


4hardfire (approved):
It worked, however we need to include this PR #1081 otherwise we cannot create Event groups with Trusted users.

PR #1083, opened 2022-01-20
[EICNET-841]feat: Set dynamic search placeholder and set case-insensitive

Dromadoros:
How to test:

- [ ] Re-index elements in SOLR
- [ ] Be sure your overview page peopleis correctly configured
image

- [ ] Be sure you have Last active and Relevance as sort
- [ ] Sort by Last active and check that info in member card "last activity..." is logic and correctly sorted
- [ ] The search label should be "Search for a member"
- [ ] Check if the search is case-insensitive


lib/themes/eic_community/react/components/Block/Overview/Search/Field/CollapsedOptions.js
```diff
@@ -18,15 +36,23 @@ const CollapsedOptions = ({values, facetsValue, bundle, updateFacet}) => {
...
+          {facets.slice(0, index).map((option) => {
             const facetValue = facetsValue[values[0]];
             const checked = facetValue ? facetValue[option[0]] : false;
 
             return <CheckboxField checked={checked} key={option[0]} facet={values[0]} updateFacet={updateFacet} value={option}/>
           })}
         </div>
+        {facets.length > minItems && (
+          <div className="ecl-filter-sidebar__item-options">
+            <button
```
filipvt1, line 47: Instead of a button, I would turn this into a standard link. That way no additional padding is added and the text aligns properly with the right hand side of the column. Correct example is the "More" link inside the "All News & Stories" on the homepage.

filipvt1 (changes requested):

PR #1086, opened 2022-01-21
[EICNET-1821]feat: Handle blocked group state with group content

Dromadoros:
How to test:

- [x] Re-index solr 
- [x] Go to /search and verify that group content's parent group is blocked is not shown for TU, GM, ... (only SA, SCM and Drupal Admin can see it)


4hardfire (changes requested):
@Dromadoros as a GO shouldn't you be able to see the content of your own blocked group? Also... when I unblock the group and go to /search I still don't see the group content.

Dromadoros:
You're right, I forgot the unpublished to publish and also and filter on group owner + resync when new flag owner accept

This is now done, thanks @4hardfire

4hardfire (approved):
Seems ok to me!

PR #1089, opened 2022-01-24
[EICNET-1879] add current page in url

Askerx9:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1879?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel

### What I did

- add params 'page'  in url
- get params 'page' from url


filipvt1 (changes requested):
If you refresh the page when for example page 2 is active, multiple ajax requests happen and you don't get the correct version of the page. In my case I get page 1 instead of page 2.
image

filipvt1 (approved):

PR #1091, opened 2022-01-24
[EICNET]chore: Upgrade to php 7.4

korhankale:
This PR upgrades the PHP version to 7.4 and brings some improvements to the compose config


4hardfire (approved):
Tested the PHP 7.4 upgrade and contrib modules upgrades.
I am currently not using the same local environment so maybe it's better if another BE Dev tests that part.

Dromadoros (approved):

PR #1092, opened 2022-01-24
(draft)[EICNET-1744]feat: Re-order field offers and needs

Dromadoros:
How to test:

- [x] check if in organisation the order of what we offer and what we need is the same than in back-end
- [x] as GO check if you edit a group you have information about required fields


4hardfire (approved):
Seems ok to me!

PR #1098, opened 2022-01-25
Feature/eicnet 1785 organisation webservice

lramarojaona:
### Improvements

- Added `field_smed_id` to Organisation group type
- Refined the `field_address` of Oragnisation group type: removed first/last name, made all fields optional
- Added a method to convert taxonomy term SMED IDs into Drupal term IDs with the given entity

### Tests

- [x] Create a user in Drupal with the `service_authentication` role, you should use those credentials as Basic Auth
- [x] Create one organisation, including the SMED ID, you should get the created entity in response
- [x] Try to create a new one with the same SMED ID, you should get an error (HTTP code 422 Unprocessable entity) with a message
- [x] Try to update the organisation, you should get the updated entity as a response

### Remarks

- You can find all the details about the endpoints URLs and payload examples here https://citnet.tech.ec.europa.eu/CITnet/confluence/display/EICNET/WS+-+D9+Implementation#WSD9Implementation-Organisations

### DIFF
--- a/config/sync/core.entity_form_display.group.organisation.default.yml
+++ b/config/sync/core.entity_form_display.group.organisation.default.yml
@@ -16,6 +16,7 @@ dependencies:
     - field.field.group.organisation.field_organisation_link
     - field.field.group.organisation.field_organisation_turnover
     - field.field.group.organisation.field_organisation_type
+    - field.field.group.organisation.field_smed_id
     - field.field.group.organisation.field_social_links
     - field.field.group.organisation.field_team_members
     - field.field.group.organisation.field_thumbnail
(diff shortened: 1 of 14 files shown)


4hardfire (approved):
Code seems ok to me.
Tests have passed.

Small remarks:

- Currently when creating an organisation via the webservice, the entity is unpublished by default. We might need to review the requirements.
- The site install process is failing because of dependency on `basic_auth` module

lramarojaona:
Ok, thanks @4hardfire , your remarks will be addressed later (we'll get rid of basic_auth).

PR #1099, opened 2022-01-25
[EICNET-1782] rework design of ecl-teaser story

Askerx9:
https://citnet.tech.ec.europa.eu/CITnet/jira/secure/RapidBoard.jspa?rapidView=6628&projectKey=EICNET&view=detail&selectedIssue=EICNET-1782&quickFilter=26826

### What I did

- [x] We miss a hr between the different news articles
- [x] The upper news article sticks on the upper hr
- [x] Date (time) should be aligned to the right
- [x] Statistics currently do not get updated
- [ ] Not sure on this one, but on organisation detail we have the introduction listed as content below title, on the organisation news overview we have description listed.
- [x] Thumbnail should be cropped automaticly to be consistent on the page.


lib/modules/eic_search/eic_search.services.yml
```diff
@@ -61,4 +61,59 @@ services:
       - { name: eic_search.source }
   eic_search.solr_document_processor:
     class: Drupal\eic_search\Service\SolrDocumentProcessor
-    arguments: ['@database', '@flag.count', '@search_api.post_request_indexing', '@eic_statistics.statistics.storage.node.decorator', '@eic_comments.helper', '@eic_media_statistics.entity_file_download_count', '@oec_group_flex.helper', '@queue', '@plugin.manager.queue_worker', '@entity_type.manager']
+    arguments: ['@database', '@flag.count', '@search_api.post_request_indexing', '@eic_statistics.statistics.storage.node.decorator', '@eic_comments.helper', '@eic_media_statistics.entity_file_download_count', '@oec_group_flex.helper', '@queue', '@plugin.manager.queue_worker', '@entity_type.manager', '@file_url_generator']
```
4hardfire, line 64: @Dromadoros some arguments should be removed since we moved the logic to different services.

4hardfire (changes requested):

lib/modules/eic_search/src/Search/DocumentProcessor/ProcessorGlobalEvent.php
```diff
@@ -0,0 +1,77 @@
...
+   * @inheritDoc
+   */
+  public function process(Document &$document, array $fields, array $items = []): void {
+    $group_type = array_key_exists('ss_group_type', $fields) ?
+      $fields['ss_group_type'] :
+      NULL;
+
+    if ($group_type !== 'event') {
+      return;
+    }
```
4hardfire, line 27: @Dromadoros is there a reason why this part is duplicated from the method `supports`?

4hardfire (changes requested):
@Dromadoros For some reason after a fresh install, when I go to the groups overview page all the groups (public/private/restricted) are showing "Public" in the visibility label. Also when I logout, I still see private groups as anonymous (even though I cannot access the group detail page). Could you have a look?

filipvt1:
Front-end OK

4hardfire:
> @Dromadoros For some reason after a fresh install, when I go to the groups overview page all the groups (public/private/restricted) are showing "Public" in the visibility label. Also when I logout, I still see private groups as anonymous (even though I cannot access the group detail page). Could you have a look?

Seems fixed!

4hardfire (approved):
BE changes seem ok to me!

PR #1107, opened 2022-01-27
[EICNET-1101]fix: Use correct label + sort option

Dromadoros:
How to test:

- [x] Go to group library overviews
- [x] Be sure you have these different sort options (with correct label)

```
Alphabetic order (A-Z): sorted on title (A-Z)
Alphabetic order (Z-A): sorted on title (Z-A)
Date uploaded (most recent first)
Highlighted files: first show all highlighted files, then the remaining files according to date uploaded (most recent first)
Number of downloads (desc): according to highest download count first
Number of views (desc): according to highest view count first
Number of likes
```


4hardfire (approved):
Seems ok to me!

PR #1108, opened 2022-01-28
Feature/eicnet 1785 membership webservice

lramarojaona:
### Tests

- [x] Make sure you have an organisation and a user both with a SMED ID
- [x] Add the user as a member of the organisation, without a role
- [x] Update the membership with the admin role and check if the user role was added correctly
- [x] Delete the membership and make sure the user is not part of the group anymore

### Documentation

https://citnet.tech.ec.europa.eu/CITnet/confluence/display/EICNET/WS+-+D9+Implementation#WSD9Implementation-Groupmemberships


4hardfire (approved):
Seems ok to me!

PR #1112, opened 2022-01-28
Feature/eicnet 1785 api key authentication

lramarojaona:
### Tests

- [x] Create a user acount with `service_authentication` role
- [x] Go to `/admin/config/eic/webservices` to add a dummy config key and Webservice user account
- [x] Change the authentication method in Postman to API Key and use `X-EIC-Auth-Token` with the key you just created in Drupal
- [ ] Try all the webservices

### Documentation

https://citnet.tech.ec.europa.eu/CITnet/confluence/display/EICNET/WS+-+D9+Implementation

@korhankale , I'm assigning this ticket to you for review, thanks! :)

This PR also include this one: https://github.com/EIC-EA/EIC-community-D8/pull/1108

### DIFF
--- a/lib/modules/eic_webservices/src/Authentication/AuthToken.php
+++ b/lib/modules/eic_webservices/src/Authentication/AuthToken.php
@@ -1,0 +1,77 @@
+<?php
+
+namespace Drupal\eic_webservices\Authentication;
+
+use Drupal\Core\Authentication\AuthenticationProviderInterface;
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Authentication provider to validate requests with token in header.
+ */
+class AuthToken implements AuthenticationProviderInterface {
+
+  /**
+   * The config factory.
+   *
+   * @var \Drupal\Core\Config\ConfigFactoryInterface
+   */
+  protected $configFactory;
+
+  /**
+   * The entity type manager service.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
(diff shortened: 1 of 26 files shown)


lib/modules/eic_webservices/src/Authentication/AuthToken.php
```diff
@@ -0,0 +1,77 @@
...
+    if (empty($webservice_user_account_uid)) {
+      return NULL;
+    }
+    if (!$account = $this->entityTypeManager->getStorage('user')->load($webservice_user_account_uid)) {
+      return NULL;
+    }
+
+    // Validate the token.
+    $token = $request->headers->get('X-EIC-Auth-Token');
+    if ($token == $api_key) {
```
korhankale, line 70: Highly advised to use strict equal here

korhankale:
Strange that the update resources for organisation and user are using the POST method, is there a reason for it (instead of using PATCH)?

korhankale:
I think we should create a user with fixtures having the service_authentication role

korhankale:
eic_webservices.settings doesn't seem to be ignored. 
Also to prevent manual changes as much as possible, I would suggest adding a default value in default.settings.local.php for these 2 configs (which is the file being copied in local envs as settings.php)
image

lramarojaona:
> 

If I add the EIC Webservices module in the config split, it also ignores lots of other configs that we need, so I didn't add it, unless you have another suggestion.

Variables have been added to default.settings.local.php

lib/modules/eic_webservices/src/Authentication/AuthToken.php
lramarojaona: You're right, I fixed it.

lramarojaona:
> I think we should create a user with fixtures having the service_authentication role

This is done.

PR #1114, opened 2022-01-29
[EICNET-1747]feat: Add new paragraph external user

Dromadoros:
How to test it:

- [ ] Edit your organisation
- [ ] Go to Team tab and add internal/external users
- [ ] Organisations field in Member paragraph should be now plain text and multiple for both paragraphs
- [ ] Go to front end and see if the look & feel is correct and all informations is shown


Askerx9:
https://citnet.tech.ec.europa.eu/CITnet/jira/secure/RapidBoard.jspa?rapidView=6628&projectKey=EICNET&view=detail&selectedIssue=EICNET-1747&quickFilter=26822

### What I did

- [ ] add user Empty state image
- [ ] The name is clickable. When the employee is a member, clicking on the name leads to the profile page.
- [ ] We have different copy on design + we miss the chevron-down
- [ ] When having a long city/country at another teammember the cards are not equally in height.

filipvt1 (changes requested):
When one of the boxes is longer, the content of the other box isn't aligned properly. It seems somewhat centered, but it should be aligned to the top. Question: should the "last activity" align to the bottom or just follow the other content?
image

PR #1115, opened 2022-01-31
[EICNET-1885]feat: Check if feature is enabled for current organisation to show/hide blocks

Dromadoros:
How to test:

- [x] Try to disable feature and see if the block linked to the feature is disabled
- [x] Try with other features in organisation


4hardfire (changes requested):
@Dromadoros When the announcements feature is enabled and there are no announcements, shouldn't we display the block with an empty message the other blocks?

Dromadoros:
Done, so now if the current user has the permission to edit the group, he will see a CTA to add announcement if there are none. It will redirect the user to the edit group form on the "announcements" tab

4hardfire (approved):
It seems ok to me.

PR #1118, opened 2022-01-31
EICNET-1825: Masquerade a user

4hardfire:
### Features

- Install masquerade module and configure permissions for SA/SCM users;
- Show "unmasquerade" link in user account menu when the user is masquerading.

### Tests

- [x] As SA/SCM, go to the admin view of members of the platform `/admin/people`
- [x] In the operation links click on "masquerade as" (you can only masquerade as TU)
- [x] Make sure you are masquerade as TU
- [x] Click on the user account menu dropdown
- [x] Make sure you have a link to "unmasquerade"
- [x] Click on "unmasquerade" and make sure you switched back to your account

Note: the "masquerade as" link is also accessible via the user profile detail page under the "Manage profile" dropdown


config/sync/user.role.site_admin.yml
```diff
@@ -146,6 +148,9 @@ permissions:
   - 'manage blocked entities'
   - 'manage contact form emails'
   - 'mark private content'
+  - 'masquerade as authenticated'
+  - 'masquerade as service_authentication'
+  - 'masquerade as trusted_user'
```
lramarojaona, line 153: If we should only be able to masquerade as TU, you should then remove authenticated
Also I don't think service_authentication is needed, this role is only for webservices

lib/themes/eic_community/includes/preprocess/layout/page.inc
```diff
@@ -60,11 +60,23 @@ function eic_community_preprocess_page(array &$variables): void {
           'path' => $account->toUrl()->toString(),
         ],
       ],
-      [
+    ];
+
+    // Adds unmasquerade link if the user is masquerading.
+    if (\Drupal::service('masquerade')->isMasquerading()) {
```
lramarojaona, line 66: I don't get why we do this through the theme, this should definitely be a block (User actions).
Anyway, that's another issue.

lramarojaona (changes requested):
It works well, we just need to remove some roles I think.

config/sync/user.role.site_admin.yml
4hardfire: Done!

lramarojaona (approved):

PR #1119, opened 2022-01-31
EICNET-2000: User contact form - add styling options

4hardfire:
### Improvements

- Create new text format "user_contact_form" with limited options to use in the User contact form;
- Fix coding standards.

### Tests

- [x] As TU/SA/SCM, go to the user contact form (via user profile detail page or `/user/<user-id>/contact`)
- [x] Make sure you can style the body field with bold, italic and underline options and add links to external pages as well.


lramarojaona (changes requested):
It works, but please use the `basic_text` format rather than creating a specific one just for contact.

And adapt the format (CKeditor buttons) if needed. We want to avoid having more than 2 HTML formats.

4hardfire:
> It works, but please use the `basic_text` format rather than creating a specific one just for contact.
> 
> And adapt the format (CKeditor buttons) if needed. We want to avoid having more than 3 HTML formats.

I updated the basic_text format and it's now being used. I also removed the user_contact_form text format.

lramarojaona (approved):

PR #1122, opened 2022-02-01
Feature/eicnet 1481 event webservice

lramarojaona:
### Tests

- [x] Try to create an event and check that it's added to Drupal
- [x] Try to create a second event with the same SMED ID and check it's refused (422)
- [x] Try to udpate the event and check that it's updated correctly

Documentation can be found here: https://citnet.tech.ec.europa.eu/CITnet/confluence/display/EICNET/WS+-+D9+Implementation#WSD9Implementation-Events


4hardfire (approved):
Seems ok to me!
Just reviewed logic of events.

PR #1124, opened 2022-02-02
Feature/eicnet 1481 event webservice

lramarojaona:
### Tests

- [x] Try to create a user through WS
- [x] Try to udpate, it should work directly (I was having an issue when updating, which I fixed)
- [x] Create a membership
- [x] Try to update the membership with a new role

Documentation: https://citnet.tech.ec.europa.eu/CITnet/confluence/display/EICNET/WS+-+D9+Implementation#WSD9Implementation-Groupmemberships


4hardfire (approved):
It seems ok to me! However, we will need to rework on the `GroupMembershipConstraintValidator` because currently we can set multiple owners for the same group via WS.

PR #1125, opened 2022-02-02
EICNET-1993: Add Drupal default workflow to all content types

4hardfire:
### Improvements

- New default workflow to use in content types
- Rework activity stream and subscription message triggers based on content moderation
- Rework the eic default content fixtures

### To fix

- [x] **_Subscription notifications_**: when a user creates new draft content in the group and then publishes the content, the notification about the creation will not be sent out because the content was previously not published;
- [x] **_Group overview pages_**: Currently draft content is not shown to the owner in group overview pages. **Answer**: only published content should be displayed.
- [x] **_Activity stream_**: the message about the content creation will never be created if the content is in draft. When the owner publishes the content, the message is showing "User updated Content X" which should be "User published Content X"
- [x] **_Global search_**: draft group content is not shown to the owner in the global search. **Answer**: only published content should be displayed.
- [x] Nodes created by fixtures are unpublished by default.

### Tests

Activity stream messages:

- [x] As GO/GA, add new **DRAFT** discussion and check the option to "Post message in the activity stream"
- [x] Navigate to the group activity stream overview page
- [x] Make sure there is no message about the creation of the draft discussion
- [x] Edit the discussion, and change the status from **DRAFT** to **PUBLISHED** and check the option to "Post message in the activity stream"
- [x] Navigate to the group activity stream overview page again
- [x] Make sure you see the activity message "Username created Discussion"
- [x] Edit the discussion, keep the **PUBLISHED** status and check the option to "Post message in the activity stream"
- [x] Navigate to the group activity stream overview page again
- [x] Make sure you see the activity message "Username updated Discussion"

Message subscriptions:

- [x] Create a new public group or use the one from the previous tests
- [x] As TU, follow the group
- [x] As GO/GA, add new **DRAFT** discussion and check the option to "Send notification"
- [x] Run cron
- [x] Check if the TU **did NOT** received an email notification (SUBSCRIPTION :: MT18)
- [x] As GO/GA, edit the discussion and change the status from **DRAFT** to **PUBLISHED** and check the option to "Send notification"
- [x] Run cron
- [x] Check if the TU received an email notification (SUBSCRIPTION :: MT18)
- [x] As GO/GA, edit the discussion, keep the status published and just check the option to "Send notification"
- [x] As TU, follow the discussion node
- [x] Run cron
- [x] Check if the TU received an email notification (SUBSCRIPTION :: MT20)

Overview pages

- [x] As GO/GA/GM/SA/SCM/TU, make sure draft group content is not shown in group overview pages and global overviews

List of group nodes

- [x] As GO/GA/SA/SCM, check if you can access the list of nodes of the group -> `/groups/group-name/nodes`
- [x] Make sure you can filter the list by moderation state


Dromadoros (approved):
It's working well! Just for everyone, there is a threshold on duplicate message so if you try to resave the published entity with the "send notification", it will not be sent

PR #1126, opened 2022-02-02
[EICNET] Author image empty state

Askerx9:
### Before
image

###  Now

image


filipvt1 (approved):
Frontend looks ok. For some reason the author first name and last name weren't present in the AJAX response on my local environment though.

PR #1129, opened 2022-02-03
EICNET-1573: Add missing fields to gallery and video activity stream messages.

lramarojaona:
### Tests

- [x] Add a Video and a Gallery in a group and tag them with a specific topic
- [x] Check the "Post activity" option
- [x] Go to the topic detail page and check that you have one activity messages for each content in the acivitiy stream

### Remarks

- If you have previous actvity messages of types Gallery or Video, this might break the atcivity stream. Either delete them or start on new install


4hardfire (approved):
Seems ok to me!

PR #1132, opened 2022-02-04
Feature/eicnet 1661 topic detail page rework

lramarojaona:
### Tests - BE

- [x] Create experts for a specific topic
- [x] Make sure you have a "Show more" link below the experts block that leads to the Members page with the topic of experties preselected

### Tests - FE

**Ticket**
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1564

- [ ] add cta in featured-contributors.html.twig
- [ ] remove overlap in teaser avatar

### DIFF
--- a/lib/themes/eic_community/patterns/compositions/featured-contributors.html.twig
+++ b/lib/themes/eic_community/patterns/compositions/featured-contributors.html.twig
@@ -42,7 +42,6 @@
         {% endif %}
 
         {% if call_to_action.link is not empty %}
-          <div>
             {% include "@ecl-twig/ec-component-link/ecl-link.html.twig" with {
               link: call_to_action.link|default({})|merge({
                 type: 'standalone',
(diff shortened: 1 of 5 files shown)


4hardfire (approved):
Tested BE and it works fine.
Just a small remark: I had to clear the cache in order for the experts block to display on the page.

lramarojaona:
Ok thanks,

@filipvt1 , can you also review the FE?

Thank you!

filipvt1 (changes requested):
There shouldn't be a '<div class="ecl-featured-list__item">' around the buttons, because of the collapse styling, those rows are clipped and the button isn't visible
Storybook still has the expand/collapse state of this component. Can you update it to have the fixed link?

lramarojaona:
> There shouldn't be a '
> ' around the buttons, because of the collapse styling, those rows are clipped and the button isn't visible
> Storybook still has the expand/collapse state of this component. Can you update it to have the fixed link?

Coudl you maybe highlight the code you're mentioning?
https://github.com/EIC-EA/EIC-community-D8/pull/1132/files

PR #1139, opened 2022-02-08
[EICNET] mails template

Askerx9:
### What I did

- [ ] create mail template section in storybook
- [ ] my_update.html
- [ ] user_tagged.html
- [ ] full_text.html


filipvt1 (changes requested):
Tagged
	- Text inside the grey box(and "with: " label) should be different grey color: should be #707070 instead of #404040, same as the meta)
	- On mobile the meta data isn't aligning properly (they are spaced out across the entire width)
	- If you go smaller than 599px there's an empty space on the right side of the grey box. Part of the box also is shown off screen to the right
	
Update
	- Text inside the grey box should be different grey color: should be #707070 instead of #404040, same as the meta)
	- The outro text (italic) below the buttons is missing
	- Same issues with grey box as "tagged template"
	
Full text
	- Links in body should be underlined
	
	
General remark:
	- In all the templates there should be an underline on "manage notifications" and "unsubscribe".

Askerx9:
@filipvt1 It's done, 

for two last point I think is only in storybook (bug?), because when I open file html, is good

filipvt1:
> @filipvt1 It's done,
> 
> for two last point I think is only in storybook (bug?), because when I open file html, is good

Yes you are right. Storybook seems to automatically load a css file that resets the styling. So it isn't an issue.

filipvt1 (approved):

PR #1140, opened 2022-02-08
EICNET-2064: As a non logged in (anonymous user) I cannot access wiki pages on the upper level

4hardfire:
### Fixes

- Fix redirect issue when accessing wiki section as anonymous user;
- Update wiki redirection when changing wiki page weights;
- Fix coding standards.

### Tests

- [x] As anonymous user, navigate to a group wiki section (`/groups/<group-name>/wiki`) which contains at least 2 wiki pages in the first level
- [x] Make sure you get redirected to the first wiki page of the section
- [x] As SA/SCM, edit the first wiki page and change the book outline order to the end
- [x] As anonymous user, try to access the wiki section `/groups/<group-name>/wiki` and make sure you get redirected to the first wiki page of the section (it should be the wiki page that was in the second position before changing the orders)
- [x] Try to access the wiki section again `/groups/<group-name>/wiki` and make sure you still get redirected to the wiki page

### DIFF
--- a/lib/modules/eic_groups/eic_groups.module
+++ b/lib/modules/eic_groups/eic_groups.module
@@ -187,6 +187,42 @@ function eic_groups_group_content_insert(EntityInterface $entity) {
     // Re-create entity alias.
     \Drupal::service('pathauto.generator')->createEntityAlias($related_entity, 'insert');
   }
+
+  switch ($entity->bundle()) {
+    case 'wiki_page':
+      // We need to clear the wiki parent book cache in order to redirect the
+      // user to the right wiki page when accessing the book page. Otherwise
+      // the redirect response might point to the wrong wiki page.
+      \Drupal::classResolver(EntityOperations::class)
+        ->clearWikiPageBookCache($entity);
+      break;
+
+  }
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_update().
+ */
+function eic_groups_node_update(EntityInterface $entity) {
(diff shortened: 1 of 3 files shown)


lib/modules/eic_groups/eic_groups.module
```diff
@@ -187,6 +187,42 @@ function eic_groups_group_content_insert(EntityInterface $entity) {
...
+
+  }
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_update().
+ */
+function eic_groups_node_update(EntityInterface $entity) {
+  /** @var \Drupal\group\Entity\GroupContentInterface[] $group_contents */
+  $group_contents = \Drupal::entityTypeManager()->getStorage('group_content')
```
lramarojaona, line 208: Here you could use `EICContentHelper->getGroupContentByEntity()`.
Ideally this function should also check the group_content plugin type (in this case gnode:wiki)
But I guess we can improve that later

lramarojaona (changes requested):
Looks good, I have a only one comment

lib/modules/eic_groups/eic_groups.module
4hardfire: It should now be using `EICContentHelper->getGroupContentByEntity()`.
Should we create a ticket to improve the method?

lramarojaona: Ok, I created a ticket: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2074

lramarojaona (approved):

PR #1143, opened 2022-02-09
EICNET-2063: As non logged in user looking at a wiki page gives CTA which overlaps with the wiki itself

4hardfire:
### Changes

- Hide login block from wiki pages.

### Tests

- [x] As anonymous user, go to a wiki page of a public group
- [x] Make sure the login block (with text: "Please login to see comments and contribute") is hidden from the page since we don't currently use comments in wiki pages

### Remarks from QA to validate with PMs

- Currently we do not have comments on wiki pages, should the CTA show different copy for this?

FA: https://citnet.tech.ec.europa.eu/CITnet/confluence/display/EICNET/CT+-+Wiki+page#CTWikipage-Commentable

- This suggests we should also activate comments on wiki pages?


lib/modules/eic_comments/src/Plugin/Block/LoginBlock.php
```diff
@@ -166,6 +166,15 @@ private function nodeHasCommentsEnabled(): bool {
       return FALSE;
     }
 
+    switch ($node->bundle()) {
```
lramarojaona, line 169: I think we should just remove the field_comments if we don't use it.
In that way, the day we want comments for wiki pages, we just need to add the field and avoid exceptions

lramarojaona (changes requested):

lib/modules/eic_comments/src/Plugin/Block/LoginBlock.php
4hardfire: Indeed. I will remove the field for now.

4hardfire: Done!

lramarojaona (approved):

PR #1144, opened 2022-02-10
EICNET-1940: Email notification for user tagging in comments

4hardfire:
### Features

- Create new message type notification for tagged users in comments;
- Create logic to send message notification to tagged users;
- Create new view mode notification teaser to use when rendering a comment in the notification message;
- Provide example template for the comment notification teaser view mode.

### DIFF
--- a/config/sync/language/en/message.template.notify_user_tagged_on_comment.yml
+++ b/config/sync/language/en/message.template.notify_user_tagged_on_comment.yml
@@ -1,0 +1,7 @@
+text:
+  -
+    value: "<p>[message:field_event_executing_user:entity:field_first_name]&nbsp;[message:field_event_executing_user:entity:field_last_name]&nbsp;tagged you in a comment</p>\r\n"
+    format: full_html
+  -
+    value: "<p><a href=\"[message:field_event_executing_user:entity:url]\">[message:field_event_executing_user:entity:field_first_name]&nbsp;[message:field_event_executing_user:entity:field_last_name]</a> tagged you in a comment:</p>\r\n\r\n<div>[message:raw-value:field_rendered_content]</div>\r\n\r\n<p><a href=\"[message:field_referenced_comment:entity:url]\">View comment</a></p>\r\n"
+    format: full_html
--- a/lib/themes/eic_community/templates/content/comment--notification-teaser.html.twig
+++ b/lib/themes/eic_community/templates/content/comment--notification-teaser.html.twig
@@ -1,0 +1,6 @@
+{# Todo: we need to change this template that will be used in the notifications. #}
+<div class="ecl-comment-overview__item">
+  <div class="ecl-comment-thread">
+  {% include "@theme/patterns/compositions/comment/comment-base.html.twig" with comment|default({}) only %}
+  </div>
+</div>
--- a/config/sync/core.entity_form_display.message.notify_user_tagged_on_comment.default.yml
+++ b/config/sync/core.entity_form_display.message.notify_user_tagged_on_comment.default.yml
@@ -1,0 +1,45 @@
+uuid: c40ac14d-5192-44a6-9373-2912307d943f
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.field.message.notify_user_tagged_on_comment.field_event_executing_user
+    - field.field.message.notify_user_tagged_on_comment.field_referenced_comment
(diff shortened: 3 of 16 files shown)


lramarojaona:
@4hardfire , it seems sopme tokens are not correctly converted, was this part of this ticket?
Subject:
`[message:field_event_executing_user:entity:field_first_name] [message:field_event_executing_user:entity:field_last_name] tagged  you in a comment
`
Message:
`[[message:field_event_executing_user:entity:field_first_name] [message:field_event_executing_user:entity:field_last_name]](http://default/user/1) tagged you in a comment:`

config/sync/language/en/message.template.notify_user_tagged_on_comment.yml
```diff
@@ -0,0 +1,7 @@
+text:
+  -
+    value: "<p>[message:field_event_executing_user:entity:field_first_name]&nbsp;[message:field_event_executing_user:entity:field_last_name]&nbsp;tagged you in a comment</p>\r\n"
```
lramarojaona, line 3: Please always use the "display name" instead of "firstname lastname". For now it's the same but in the end we can always manage the display name in one place instead of hardcoding firstanme/lastname.

In this case: `[message:field_event_executing_user:entity:display-name]`

lib/themes/eic_community/includes/preprocess/users/user.inc
```diff
@@ -253,17 +255,17 @@ function eic_community_get_teaser_user_display(UserInterface $user) {
...
-    $media = \Drupal::service('entity.repository')->getTranslationFromContext($user->get('field_media')->entity);
-    if ($media->bundle() == 'image') {
-      $image_item = ImageValueObject::fromImageItem($media->get('oe_media_image')->first());
-      $author['image'] = [
-        'src' => $image_item->getSource(),
-        'alt' => $author['name'],
-      ];
-    }
-  }
+  // if (!$user->get('field_media')->isEmpty()) {
```
lramarojaona, line 258: What's the reason for commenting this? Should we keep it?

lib/themes/eic_community/templates/content/comment--notification-teaser.html.twig
```diff
@@ -0,0 +1,130 @@
+{# Todo: we need to change this template that will be used in the notifications. #}
```
lramarojaona, line 1: Is there a reason we need a specific template??

lib/themes/eic_community/templates/mimemail/mimemail-message--message-notify--notify-user-tagged-on-comment.html.twig
```diff
@@ -0,0 +1,231 @@
+{% extends "@eic_community/mimemail/mimemail-message.html.twig" %}
```
lramarojaona, line 1: Is there a reason we need a specific template??

lramarojaona (changes requested):

lib/themes/eic_community/templates/content/comment--notification-teaser.html.twig
4hardfire: This is the rendered content of the comment that goes in the notification. Currently I've implemented this way and there is a field_rendered_content in the notification message which we use it as a token in the body of the message.
Should we do it this way or do you think we should add this html in the message body itself (using tokens)?

lib/themes/eic_community/includes/preprocess/users/user.inc
4hardfire: This was a mistake!

4hardfire:
> @4hardfire , it seems sopme tokens are not correctly converted, was this part of this ticket? Subject: `[message:field_event_executing_user:entity:field_first_name] [message:field_event_executing_user:entity:field_last_name] tagged you in a comment ` Message: `[[message:field_event_executing_user:entity:field_first_name] [message:field_event_executing_user:entity:field_last_name]](http://default/user/1) tagged you in a comment:`

@lramarojaona were you testing with Drupal admin locally? Because by default that user doesn't have a First and Last name.

config/sync/language/en/message.template.notify_user_tagged_on_comment.yml
4hardfire: Fixed!

lramarojaona (approved):

PR #1151, opened 2022-02-15
EICNET-2076: Disable like action and statistics in comments

4hardfire:
### Improvements

- Disable like + reply + statistics if a comment was soft deleted or archived.
- Improve access checks on comment flags.
- Get rid of custom service decorators for Flag/Unflag in `eic_groups/src/Access` access checks and implement hook_flag_action_access() instead.

### To discuss

- [x] We need to check access to flag/unflag a comment when a comment has been archived. Currently we cannot know if a comment has been archived without looping through all request archivals. We should create a property or a field to identify if a comment is archived.

### Tests

- [x] As GM1, add a comment in a discussion
- [x] As GM2, reply to the previous comment
- [x] As SA/SCM, delete the first comment
- [x] As GM/GO/GA/SA/SCM, make sure you cannot like or reply to the deleted comment and you don't see any like statistics.
- [x] As GM1, request archival of the second comment
- [x] As SA/SCM, accept archival request
- [x] As GM/GO/GA/SA/SCM, make sure you cannot like or reply to the archived comment and you don't see any like statistics.


Dromadoros:
Tested it, it's working well and it's clean!

Dromadoros (approved):

PR #1153, opened 2022-02-15
[EICNET-1782]feat: Set introduction field in teaser of search news

Dromadoros:
How to test:

- [x] Re-index everything
- [x] Go to organisation detail in News overview
- [x] check that the description in the teaser is the field "introduction"


lib/themes/eic_community/react/components/Block/Overview/Search/ResultItem/StoryResultItem.js
```diff
@@ -38,7 +37,7 @@ export default function StoryResultItem(props) {
               <span className="ecl-teaser__title-overflow"><span>{props.result.tm_global_title}</span></span>
             </a>
           </h2>
-          <div className="ecl-teaser__description" dangerouslySetInnerHTML={{__html: text}} />
+          <div className="ecl-teaser__description" dangerouslySetInnerHTML={{__html: props.result.ss_content_introduction_string.replace(/(<([^>]+)>)/gi, "")}} />
```
lramarojaona, line 40: You should use the `HTML filter` processor provided by Search API ( we need to check if all text fields are applicable or if we should exclude some of them).
This way we strip tags only on index process and not on each display in the browser.

lramarojaona (changes requested):

lramarojaona:
@Dromadoros , also in organisation news, I see both the Introduction and the body displayed in the overview list I guess it should be introduction only?

Dromadoros:
> @Dromadoros , also in organisation news, I see both the Introduction and the body displayed in the overview list I guess it should be introduction only?

I think you need to clean all data in solr and re-index it, it doesn't take anymore the basic render of news type

lib/themes/eic_community/react/components/Block/Overview/Search/ResultItem/StoryResultItem.js
Dromadoros: > You should use the `HTML filter` processor provided by Search API ( we need to check if all text fields are applicable or if we should exclude some of them). This way we strip tags only on index process and not on each display in the browser.

It's a good idea but fulltext is used a lot of times and I'm scared that it will break some stuff on some components. So for performance wised like you said I removed on the front itself and added it on the index process. So when we will load the overview it will not apply the striptag function one by one anymore

lramarojaona (approved):

PR #1156, opened 2022-02-16
[EICNET-1741]fix: Add tag in header organisation and make it clickable + fix permissions

Dromadoros:
How to test:

- [x] First be sure your overview page has the facet "Organisation type" enabled
- [x] Create a new Organisation
- [x] Check if as GO you see actions select button on the header
- [x] Also on the group header, you should see now organisation types as tags
- [x] You should be able to click on the tag and redirected to organisations overview prefiltered by organisation type


4hardfire (commented):
Seems ok to me!

Just a small issue that I found (not related though):
- As GO (trusted user), when the group is in DRAFT state I cannot publish it. Apparently only SA/SCM can do that.

4hardfire (approved):

PR #1158, opened 2022-02-16
[EICNET-1881]fix: Replace translation + addresses label + autocomplete internal link

Dromadoros:
How to test it:

- [x] Edit an organisation
- [x] Add social links
- [x] Try to put an internal link and check if auto suggestion is working
- [x] Put multiple addresses and verify that in the front-end we see correct title in top of each address item (it should be company or city)
- [x] Add just 1 location, now it's a paragraph and verify that in front end it's says: 1 location
- [x] Add another location and check if you see: 2 locations

### DIFF
--- a/lib/modules/eic_private_message/src/Plugin/EntityReferenceSelection/MembershipSelection.php
+++ b/lib/modules/eic_private_message/src/Plugin/EntityReferenceSelection/MembershipSelection.php
@@ -3,7 +3,6 @@
 namespace Drupal\eic_private_message\Plugin\EntityReferenceSelection;
 
 use Drupal\Core\Database\Database;
-use Drupal\Core\Entity\Annotation\EntityReferenceSelection;
 use Drupal\Core\Entity\Plugin\EntityReferenceSelection\DefaultSelection;
 use Drupal\user\Entity\User;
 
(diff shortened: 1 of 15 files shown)


lramarojaona:
@Dromadoros , for this one: _Try to put an internal link and check if auto suggestion is working_, I guess you're talking about the Link field?

lib/modules/eic_private_message/src/Plugin/EntityReferenceSelection/MembershipSelection.php
```diff
@@ -28,6 +27,36 @@ class MembershipSelection extends DefaultSelection {
...
+
+      // Adding the permission check is sadly insufficient for users: core
+      // requires us to also know about the concept of 'blocked' and 'active'.
+      if (!$this->currentUser->hasPermission('administer users')) {
+        $query->condition('status', 1);
+      }
+
+      return $query;
+    }
+
```
lramarojaona, line 59: @Dromadoros , does this have anythin to do with this PR?

lramarojaona (approved):
It's working, I only have 2 questions

Dromadoros:
> @Dromadoros , for this one: _Try to put an internal link and check if auto suggestion is working_, I guess you're talking about the Link field?

Yes that's the one

lib/modules/eic_private_message/src/Plugin/EntityReferenceSelection/MembershipSelection.php
Dromadoros: Yes, it's that one change that fix the issue with link internal suggestions

PR #1174, opened 2022-02-18
[EICNET-1943]feat: Add subnavigation on the profile header

Dromadoros:
How to test:

- [x] Go to /my-profile/activity
- [x] Check if you see the subnavigation with different menu items
- [x] The current menu item should be active, in our cases it should be: "Interesting for you"

PS: we do not have other overviews/pages so for the moment other menu items have the same URL behind


4hardfire (changes requested):
Everything seems to be working fine, although I would change the router path of the activity feed to `/user/{user}/activity`. Note that this would impact your access method because it currently assumes that you are trying to access your own profile but if you change to `/user/<user-id>/activity` of another profile you should have access denied. In my opinion this change makes the website more SEO friendly as well.

4hardfire (approved):
The changes will be done in another PR.

PR #1175, opened 2022-02-20
[EICNET-1945]feat: Implement correct filter for profile activity feed + set correct datasource feed

Dromadoros:
How to test:

- [x] Re-sync search API
- [x] Go to /my-profile/activity
- [x] Verify that you have filters: "Only my groups", "Type" and "Topics"
- [x] Test different filters, especially the "only my groups"

In this one I also updated the feed results:

- [x] Now we are getting messages sources
- [x] Check if you have correct content ordered by creation date
- [x] Operation type should be there "created", "updated"

### DIFF
--- a/config/sync/context.context.my_profile.yml
+++ b/config/sync/context.context.my_profile.yml
@@ -38,40 +38,6 @@ reactions:
         unique: 0
         context_id: my_profile
         context_mapping: {  }
-      6c509f98-b687-4e8d-a7f9-8d8f81c5bc9d:
-        id: eic_search_overview
-        label: ''
-        provider: eic_search
-        label_display: visible
-        context_mapping: {  }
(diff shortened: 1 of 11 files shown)


Dromadoros:
> I've tested the filters and it seems to be working properly. However, I still have some questions concerning the content that should be shown in the list. I thought it should be showing content which are tagged with topics of your interest but apparently it's not. Is this the normal behaviour that we have defined for this feature? Also when I create a news or story, it's not shown in the list.

I kind of agree with you, we can create news and story but it's not listed on the results and that's strange. But in the FA it says that this overview is an extend to Activity stream we have. And it's taking content from all groups or only my groups. News and story are not part of groups so it will never be listed, that's why.

For what I understood, it lists all content from groups except his contents and for MVP we list only all topics and not his topics of interests on the sidebar. We had some call for that and it was complicated to only list his topics of interests in this generic overview

4hardfire (approved):

PR #1178, opened 2022-02-21
Feature/eicnet 1924 share group content rework

lramarojaona:
### Improvements

- Created a new group_content relation type: `group_shared_content`

### Tests

- [x] As TU try to share a node from a public group to another public group (you must be member of the target group)
- [x] Try to share again to the same group, you should get a message "This content is already shared with this group"
- [x] Go to the target group and check that the content appears in the group homepage and/or content-specific overview
- [x] As GO/Ga of the target group, check that you can't edit the shared node

### DIFF
--- a/config/sync/group.content_type.event-group_shared_content.yml
+++ b/config/sync/group.content_type.event-group_shared_content.yml
@@ -1,0 +1,18 @@
+uuid: 3e678500-41a5-4839-8dad-51b5bc60e31d
+langcode: en
+status: true
+dependencies:
+  config:
+    - group.type.event
+  module:
+    - eic_share_content
+    - node
+id: event-group_shared_content
+label: 'Event: Group Shared content'
+description: 'Adds shared content to groups.'
+group_type: event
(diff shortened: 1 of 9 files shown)


4hardfire (changes requested):
The last step didn't work on my local.

Also, when testing the 1st step I as TU, I could share any content of a public group with another public even though I was **not** a member of any of both groups.

lramarojaona:
> The last step didn't work on my local.
> 
> Also, when testing the 1st step I as TU, I could share any content of a public group with another public even though I was **not** a member of any of both groups.

@4hardfire , Thanks for pointing that.

1. Just tested as GO and GA of the target group, I can't edit the shared node. Are you sure that when testing you were not already GO/GA of the srouce group?
2. You're right, I just created a ticket for that: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2110

4hardfire (approved):
Besides the follow-up ticket that needs to be tackled later, everything works properly!

PR #1179, opened 2022-02-21
[EICNET-1728]fix: Change search fields in solr for user, change label, check user email exists

Dromadoros:
How to test:

- [ ] Go to a group detail and click on "invite multiple users" action
- [ ] Try to invite internal users and external, check that it's case insensitive for internal
- [ ] Try also to invite users that have not an email setup (take QA DB there are some cases)
- [ ] On emails, check if you have the link to the group: "Visit also the group: URL_GROUP"


lramarojaona (approved):
Works good!

PR #1182, opened 2022-02-22
[EICNET]feat: Change route url for my interests

Dromadoros:
How to test:

- [x] Go to: /user/USER_ID/activity (replace USER_ID) by your uid
- [x] Verify that you can access and see the overview
- [x] Go to the same URL for another user id, check if you do not have permission to check it


4hardfire (approved):
Looks good to me!

PR #1183, opened 2022-02-22
EICNET-2107: Allow admin roles to edit all comments in all group types.

lramarojaona:
### Tests

- [x] As TU post a comment in a group
- [x] as SCM check that you can edit the comment

### Remarks

- Permissions have been changed on group type level, so it won't apply automatically for existing groups, it will work only for new groups


4hardfire (approved):
Seems ok to me!

PR #1184, opened 2022-02-22
EICNET-2092: Make sure group overviews return access denied when group is blocked.

lramarojaona:
### Tests

- [ ] As GM open all group overview pages (including about, discussions, latest activity, wiki, `/groups/<group-name>/members` etc...)
- [ ] As SCM block the group
- [ ] As GM refresh all overview pages and make sur you get an access denied with the "Group under review" block


4hardfire:
@lramarojaona I fixed the access check so that it works only for view pages and also returns access allowed if the group is not blocked. We need to return access allowed if the group is not blocked because the final access will be combined with the '_group_permission' requirement and if the user doesn't have permission it will return access denied even if our custom access check returns allowed.
Could you test it and put the PR ready for review?

lramarojaona:
Thanks @4hardfire , it works but I can't approve the PR myself since I'm the author

4hardfire (approved):
It works!

PR #1189, opened 2022-02-23
[EICNET]fix: use load more for activity feed

Dromadoros:
How to test:

- [x] Go to /user/USER_ID/activity
- [x] Check if you have a loadmore with 10 items  by batch

PS: if you do not have enough items, add group contents with another use. Cause your user's content will be prefiltered on results


lramarojaona (approved):
It works, but when loading new items it's a bit confusing, cause the page seems to "jump" to another place and we get a bit lost.

Dromadoros:
> It works, but when loading new items it's a bit confusing, cause the page seems to "jump" to another place and we get a bit lost.

I agree, we can maybe create an improvement ticket for a FE

PR #1190, opened 2022-02-23
[EICNET]fix: reindex messages and node when downloading media

Dromadoros:
How to test:

- [ ] Go to /user/USER_ID/activity and be sure you have a document activity listed
- [ ] You should have download statistics
- [ ] Try to download again on the document node and go back to your interests and see if it has been updated


lib/modules/eic_media/modules/eic_media_statistics/src/Controller/MediaFileDownloadController.php
```diff
@@ -56,6 +65,7 @@ public function download(MediaInterface $media) {
     $config = $source->getConfiguration();
     $field = $config['source_field'];
     $request_query = $this->requestStack->getCurrentRequest()->query;
+    $request_query->get('node_id', NULL);
```
korhankale, line 68: Second param's default value is null anyway, we can remove it

korhankale (approved):

PR #1195, opened 2022-02-25
[EICNET-190]fix: do not enable add news/event add content for GM

Dromadoros:
How to test:

- [x] Create an organisation
- [x] Enable feature news & event
- [x] as a GM of the group, verify that you cannot add news/event


4hardfire (approved):
Seems ok to me!

PR #1196, opened 2022-02-25
[EICNET-1909]fix: Set event type instead of topics in teaser tag + format image

Dromadoros:
How to test:

- [x] Go to an organisation detail page
- [x] Be sure you have different events
- [x] In the block events, verify you have in the tag the event type field
- [x] Click on the"View all events" and see if in the overview we still have the event type field
- [x] Now check if the teaser image is no more in Portrait mode but more square to match design


4hardfire (approved):
Seems ok to me!

PR #1200, opened 2022-02-28
EICNET-2110: Share content: make sure users can share to the groups they belong only

lramarojaona:
### Improvements

- Moved the logic to get target groups to the ShareManager module
- Now filtering traget groups based on user
- Created a static method to check if group is blocked
- Created a static method to check if group is archived
- Changed the Archived state to remain published

### Tests

- [x] Make sure you have a mix of published public/private groups
- [x] Add TU as GM to some public and private groups
- [x] As TU create a public group and as SA/SCM publish it
- [x] As TU go to a public group and try to share a content
- [x] Make sure you can share only to _public_ groups you are member of
- [x] As SA/SCM, try to sahre the same content
- [x] Make sure you can share to all public groups, even if you're not a member of all groups (e.g. the group created by TU)
- [x] As SA/SCM block one of those public groups
- [x] As SA/SCM/TU make sure you can't share to the blocked group
- [x] As SA/SCM archive one of those public groups
- [x] As SA/SCM/TU make sure you can't share to the archived group

### DIFF
--- a/config/sync/views.view.admin_groups.yml
+++ b/config/sync/views.view.admin_groups.yml
@@ -8,6 +8,7 @@ dependencies:
   module:
     - content_moderation
     - group
+    - oec_group_flex
     - user
     - views_autocomplete_filters
 id: admin_groups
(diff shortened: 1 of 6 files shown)


4hardfire (changes requested):
The last step failed when testing edge cases:

- As TU, go to a discussion of a public group, click on the share button and leave the form open
- Open another browser and as SA/SCM block the group
- As TU (still with the share form open), I can still see the blocked group in the list and if I select the blocked group, the content gets shared anyways. We should have a validation before the form submit to make sure that we can actually share the content.

lramarojaona:
@4hardfire ,
You're right I added a check when actually submitting the share, also I added an additional check for archived groups.
I've just updated the tests accordingly.

Can you re-test?

4hardfire (changes requested):
There is still an issue:

- As TU, go to a discussion of group 1, click on the share button and leave the form open
- Open another browser and as SA/SCM block the group 1
- As TU (still with the share form open but now the group is blocked), when you select group 2 the content still gets shared even though it belongs to a blocked group.

4hardfire (approved):
Currently when a shared content belongs to a blocked group, the content is still visible in the overview lists of the group when it was shared. We should fix this in a follow-up ticket.

lramarojaona:
> Currently when a shared content belongs to a blocked group, the content is still visible in the overview lists of the group when it was shared. We should fix this in a follow-up ticket.

Just created the followup ticket: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2139

PR #1201, opened 2022-02-28
EICNET-1991: Make image fields optional for organisations.

lramarojaona:
### Tests

- [x] Create a new organisation without banner and thumbnail images
- [x] Make sure the display is not broken on organisation homepage and organisation overview


4hardfire (approved):
Seems ok to me!

PR #1202, opened 2022-02-28
[EICNET-1985]fix: If user is not logged in, do not authorize organisations

Dromadoros:
How to test:

- [x] Go as TU in /organisations overview and check if you have access
- [x] Now go as anonymous user and check if you have a permission denied


4hardfire (changes requested):
Just a small fix. The rest looks ok to me!

PR #1203, opened 2022-02-28
EICNET-2116: News articles within organisation can be promoted on homepage

4hardfire:
### Improvements

- Create new views filter plugin for group visibility;
- Update entity browser for selecting News and Stories to show in the homepage so that it only allows SA/SCM users to select published content (including content from public groups).

### Tests

- [x] As SA/SCM, navigate to -> `/admin/content/block-content`
- [x] Edit the block "Latest News & Stories"
- [x] In the edit form, make sure you can only select News or Stories that are published + not private (community members only checkbox disabled) + from public events

### DIFF
--- a/config/sync/views.view.latest_news_and_stories.yml
+++ b/config/sync/views.view.latest_news_and_stories.yml
@@ -9,9 +9,12 @@ dependencies:
     - node.type.story
   module:
     - entity_browser
+    - group
     - media
     - node
+    - oec_group_flex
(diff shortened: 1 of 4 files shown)


Dromadoros:
When editing the block, search for article, I have nothing proposed in the widget:

image

No JS error, no call error. Didn't have this issue before

Dromadoros (approved):
Working well

PR #1205, opened 2022-02-28
[EICNET-1746]feat: Add field search api event weight state + teaser event

Dromadoros:
How to test:

- [x] Go to organisations details and be sure you have at least more than 4 events
- [x] See if you have max 3 events listed in the block events
- [x] Verify if you see the teaser image in the teaser
- [x] Also check that you have remote event label if remote or address if on site (city, country)
- [x] Check if it's ordered by ongoing, future then past

### DIFF
--- a/config/sync/context.context.following.yml
+++ b/config/sync/context.context.following.yml
@@ -1,0 +1,70 @@
+uuid: 093c1feb-051c-4362-94b5-54470e338c04
+langcode: en
+status: true
+dependencies:
+  module:
+    - eic_search
+    - route_condition
+label: Following
+name: following
+group: 'My profile'
+description: ''
+requireAllConditions: false
+disabled: false
+conditions:
+  route:
+    id: route
+    negate: false
+    uuid: 32d8c454-78f3-473f-a548-b90f18ade3cc
+    context_mapping: {  }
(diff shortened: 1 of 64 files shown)


4hardfire (changes requested):
I am having 2 issues when updating an event from the organisation overview page:

`Notice: Undefined index: in Drupal\eic_groups\EICGroupsHelper::formatAddress() (line 980 of /var/www/html/lib/modules/eic_groups/src/EICGroupsHelper.php)`

`Twig\Error\RuntimeError: The merge filter only works with arrays or "Traversable", got "object" as first argument. in twig_array_merge() (line 142 of /var/www/html/lib/themes/eic_community/patterns/compositions/event/event.teaser.html.twig).`

4hardfire:
@Dromadoros the previous issue seems fixed. However, when using location "On site" it only shows the city and country is missing. Is it supposed?

Dromadoros:
> @Dromadoros the previous issue seems fixed. However, when using location "On site" it only shows the city and country is missing. Is it supposed?

You need to put the taxonomy Country field and also the City in the address from Location field

4hardfire (approved):
It works!

PR #1207, opened 2022-03-01
[EICNET-1742]feat: Implement template + refactor about page organisation

Dromadoros:
How to test:

- [x] Go to an organisation detail
- [x] Click on "About" tab
- [x] Check if you see informations in it within a correct template


4hardfire (approved):
Seems ok to me!

PR #1208, opened 2022-03-01
[EICNET-2111]feat: Add highlight comment parameter on the URL

Dromadoros:
How to test it:

- [x] Go in a discussion
- [x] tag user in a comment (try with different level)
- [x] run cron
- [x] check if you receive email
- [x] Click on view comment
- [x] The top comment should be the one where you tagged in (always top comment is at the top, not the answer to avoid strange behaviour)


4hardfire (changes requested):
@Dromadoros I have an issue when I try to tag users in a reply. For some reason the form disappears after I confirm the tagged users. Can you have a look?

There is also another issue where I cannot request archival or deletion of a reply. When I press submit, it does nothing.

PR #1209, opened 2022-03-01
[EICNET-2112]fix: Handle also archived moderation state

Dromadoros:
How to test:

- [ ] Archive a group
- [ ] Go to /groups and see if your group has the label "Archived"


4hardfire (approved):
It works!

PR #1210, opened 2022-03-01
[EICNET-1961]feat: Create the "following" activities overview

Dromadoros:
How to test:

- [x] Follow different users/groups/contents/topics
- [x] Go to /user/USER_ID/following page
- [x] check if you have correct results link to your interests
- [x] Verify that results are dynamic, unfollow a content and check if you do not see the content anymore (be sure you do not follow the user of the content, the topic or the group so it can disappear)


4hardfire (approved):
Seems ok to me!

PR #1211, opened 2022-03-02
Feature/eicnet 1319 node share migration

lramarojaona:
### Improvements

- Created a `eic_d7_og_membership` process plugin to get the gid of an entity

### Tests

- [x] Run `drush mim upgrade_d7_node_share_to_group_content` and check that nodes are being shared with the dedicated group relationship (group_content: `<group-type>-group_shared_content`)

### Remarks

- There's a body field that exists on D7 for the Share nodes that we don't have/use on D9. This is the text that user enter when sharing a node. We need to check if this is needed for us.
- If a share node is referencing a non-exisintg node or group, it will be skipped. Currently I have  `[notice] Processed 92 items (60 created, 0 updated, 0 failed, 32 ignored) - done with 'upgrade_d7_node_share_to_group_content'`


4hardfire (approved):
It seems ok to me!

I've checked the database and apparently the skipped rows belong to organisations and we didn't implement that migration yet.
Also, it looks like some shared content relates to News. Is this something that we will allow? Currently there we don't show shared News in groups.

lramarojaona:
> It seems ok to me!
> 
> I've checked the database and apparently the skipped rows belong to organisations and we didn't implement that migration yet. Also, it looks like some shared content relates to News. Is this something that we will allow? Currently there we don't show shared News in groups.

It should be supported: https://github.com/EIC-EA/EIC-community-D8/blob/develop/lib/modules/eic_groups/modules/eic_share_content/src/Service/ShareManager.php#L130

PR #1212, opened 2022-03-02
EICNET-2113: Add 'administer members' permissions to relevant roles in Organisation group type

lramarojaona:
### Tests

- [x] As SA/DA/SCM/GO/GA, create an organisation
- [x] As SA/DA/SCM/GO/GA, try to access /members page of this organisation
- [x] Make sure you have access to it


4hardfire (changes requested):
I still don't have access as SA/SCM. In any case, you need to change the permission and enable "View individual group members" instead because we are currently using a custom view for the /members page. We cannot use the "administer members" permission because AFAIK GA shouldn't be able to edit memberships otherwise the user will be able to remove his own admin role.

lramarojaona:
@4hardfire ,
You're right, it seems group permissions are not created when generated by fixtures. I created a follow up-ticket: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2164

Could you re-test and create an organisation instead?
I'll update the tests

4hardfire (approved):
Indeed, it seems to be an issue with fixtures. It works if I create a new organisation manually.

PR #1219, opened 2022-03-03
[EICNET]fix: Set type as multi array in teaser event

Dromadoros:
How to test:

- [x] Be sure to have global event
- [x] Clear cache
- [x] Go to homepage, see if events have tag and page does not return 500


4hardfire (approved):
Seems ok to me!

PR #1227, opened 2022-03-07
[EICNET-1943] send icon_file_path to my activity view

Askerx9:
### To test

- check on mobile if the burger icon is here


filipvt1 (changes requested):
It works in storybook but doesn't work in the Drupal implementation yet. Guess this also needs to be added inside the Drupal templates.

PR #1229, opened 2022-03-07
[EICNET-2138] add space between image and info in organization member teaser

Askerx9:
### What I did

- add margin right on  image  in organization member teaser

### DIFF
--- a/lib/themes/eic_community/sass/compositions/_teaser.scss
+++ b/lib/themes/eic_community/sass/compositions/_teaser.scss
@@ -146,7 +146,7 @@
     object-fit: cover;
     border-radius: 50%;
     overflow: hidden;
-    margin-right: $ecl-spacing-m;
+    margin-right: map-get($ecl-spacing, 'l');
   }
 
   &--organisation#{$node}--as-card-white &__main-wrapper{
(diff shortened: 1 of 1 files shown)


lib/themes/eic_community/sass/compositions/_teaser.scss
```diff
@@ -144,6 +146,7 @@
     object-fit: cover;
     border-radius: 50%;
     overflow: hidden;
+    margin-right: $ecl-spacing-s;
```
filipvt1, line 149: On other teasers this spacing is 1.25rem and not 0.75rem. Might be best to make this the same.

filipvt1 (changes requested):

filipvt1 (changes requested):
It's a bit too big now. It's annoying because the spacing is a bit inconsistent. But it seems the original image wrapper uses ecl-box-model('padding')

filipvt1 (changes requested):
There is a weird inconsistency between the spacing variables.
$ecl-spacing-l is 1rem while in the $ecl-spacing object the l value is 1.25rem. 
Can you make sure to use the value from in side the project instead of the ecl one by using this value: map-get($ecl-spacing, 'l')

filipvt1 (approved):

PR #1254, opened 2022-03-11
EICNET-1993: Drupal default workflow improvements in message creations 

4hardfire:
### Improvements

- Skip entity activity stream and subscription message creation when user creates a new DRAFT from a PUBLISHED version.

### Tests

- [x] As GM go to a group, create a new published discussion
- [x] Edit the discussion, set the moderation state to draft and check the option "Post message in the activity stream"
- [x] Go to the lastest activity -> `/groups/<group-name>/latest-activity` and make sure there is no message about the discussion update.


lramarojaona (approved):
Ok, it works!

PR #1255, opened 2022-03-11
EICNET-2120: Make sure we don't show the group sharing link if user doesn't have access to it

lramarojaona:
### Tests

- [x] As TU, go to a (fresh) group and check that you can share content to antoehr group
- [x] Remove the `share content between groups` permission from the group
- [x] Clear the cache
- [x] As TU, check that you don't have the share to group link anymore

### DIFF
--- a/lib/modules/eic_groups/modules/eic_share_content/src/Service/ShareManager.php
+++ b/lib/modules/eic_groups/modules/eic_share_content/src/Service/ShareManager.php
@@ -161,7 +161,7 @@ class ShareManager {
     }
 
     // Check if we can share from source group to target group.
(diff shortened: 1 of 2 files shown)


4hardfire (commented):
It works. However, I still have some remarks:

- As TU, I can only share content with groups where I am a member. The problem is when I am a TU and I didn't join any group, when I go to the discussion I see the "share with another group" but when I click it doesn't open the form and the button disappears. And then I join another group, and then I need to clear the cache in order for the share button to open the form.

Created new follow-up ticket https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2167.

4hardfire (approved):

PR #1257, opened 2022-03-11
[EICNET] Enable UIT

Dromadoros:
Implementing the the user interface translation.

You can now edit every labels, texts, etc. that have been implemented via the t() method.


lramarojaona (approved):
Ok, it works!

PR #1259, opened 2022-03-11
EICNET-2137: Deny access to GO/GA to group edit page for blocked groups.

lramarojaona:
### Tests

- [x] As GO/GA, edit your group page
- [x] As SA/SCM/DA, block the group
- [x] As GO/GA try to edit the group again => you should get an access denied with "Group under review"


4hardfire (commented):
It works! However in my opinion we should allow GO/GA to still be able to edit the group information.

How do expect to be able to fix for example the group description if that is the reason for blocking the group? Should be the SA/SCM to edit this information? For me it doesn't make much sense and we already make sure that the user cannot change the group state from blocked to published.

In my opinion we should discuss this with the PM.

PR #1262, opened 2022-03-14
[EICNET-1728]feat: Add links to the invite mail notification + error text

Dromadoros:
How to test:

- [ ] Try to invite a user twice so you can have the message error. Verify that you have this: User @email already received an invitation.
- [ ] Now try to invite another existing user and a non existing user on the platform 
- [ ] You should received 2 e-mails
- [ ] Verify that you have links in there and also check if redirects are working correctly


4hardfire (approved):
Seems ok to me!

PR #1265, opened 2022-03-14
[EICNET]chore: Clear indexes on local setup

korhankale:
## To Test
- [x] Run `make setup`
- [x] Check that your solr index doesn't contain more documents than what's existing in your drupal database


lramarojaona (approved):
It works!

PR #1266, opened 2022-03-14
[EICNET-1908]fix: Add teams cta + ignore anonymous user + add language filter

Dromadoros:
How to test:

- [x] Add internal/external user
- [x] Be sure you can click on user link name but external doesnt have a link on the fullname
- [x] Check in teams overview page if you cannot see anonymous user
- [x] In the overview you should have language facet


Dromadoros:
> @Dromadoros for some reason I can't see any results in team overview. Tried to add team members + re-index solr + clear all cache but still no results. Can you have a look?

Are you sure that you have a profile link to your user? Otherwise it cannot work. Be sure to have a profile, save it and re-index

PR #1269, opened 2022-03-14
[EICNET-2176]chore: Move styles for user tagged messages within template

korhankale:
## To Test

- [x] Tag a user on a comment
- [x] Run cron
- [x] Check that the mail is displaying correctly (should mainly contain classes)


4hardfire (commented):
@korhankale I thought the idea was to move the link "View comment" from the message text to the template `comment--notification-teaser.html.twig`.
Also the button is still on the left because we are missing html and some styles. You can check the `user_tagged.html` file in `/themes/eic_community/mails/mail-dist` which should contain all the necessary html + inline styles for the button. Can you have a look?

You might also need to preprocess some variables in order to send the View comment URL to the theme variables.

4hardfire (approved):
Looks good to me!
Just the **View comment** button that needs FE rework on the styles but that will be fixed later.

PR #1270, opened 2022-03-14
EICNET-2160: Make sure only allowed workflow states show up in groups.

lramarojaona:
### Tests

- [x] As SA/SCM/DA, create a user and add it as GA of an organisation
- [x] As GA of the organisation, create a News Article, don't submit
- [x] As GA, make sure you don't have "Submit for approval" button


lramarojaona:
> For some reason when I try to create a global News or Story I get the following error: `Drupal\Core\Entity\EntityStorageException: Cannot load GroupContent entities for an unsaved entity. in Drupal\group\Entity\Storage\GroupContentStorage->loadByEntity() (line 92 of /var/www/html/web/modules/contrib/group/src/Entity/Storage/GroupContentStorage.php).`
> 
> Could you have a look?

Nice catch!
It is fixed now.

4hardfire (approved):
Seems ok to me!

PR #1276, opened 2022-03-15
Feature/eicnet comments edit tag user

Askerx9:
### What I did

- now you can tag users in edit popup


filipvt1 (changes requested):
The icon isn't centered in the button.  There is more spacing on the right side.
image

filipvt1 (approved):

PR #1277, opened 2022-03-15
[EICNET-2055] change icone in my profile btn in notif management

Askerx9:
https://citnet.tech.ec.europa.eu/CITnet/jira/secure/RapidBoard.jspa?rapidView=6628&projectKey=EICNET&view=detail&selectedIssue=EICNET-2055&quickFilter=26822

### What I did

- I change icon from gear to user in btn my profile

### DIFF
--- a/lib/themes/eic_community/patterns/compositions/member/email-notification.html.twig
+++ b/lib/themes/eic_community/patterns/compositions/member/email-notification.html.twig
@@ -1,61 +1,57 @@
 {% set title_element = title_element|default('h2') %}
 
 <section class="ecl-my-notifications">
-  <div class="ecl-my-notifications__header">
-    <div>
-      {% if title is not empty %}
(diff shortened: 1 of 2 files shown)


lib/themes/eic_community/patterns/compositions/member/email-notification.html.twig
```diff
@@ -14,7 +14,7 @@
         }),
         icon: icon_file_path ? {
           path: icon_file_path,
-          name: action.icon.name|default('gear'),
+          name: action.icon.name|default('user'),
```
filipvt1, line 17: I don't think this is the icon they mentioned in the ticket. This icon is shown next to the "Edit interests" button, the one in the ticket is the button higher up the page. Changing the default here also has no impact as the icon is determined by "action.icon.name" value in the data files.

filipvt1 (changes requested):

filipvt1 (approved):

PR #1280, opened 2022-03-15
EICNET-2182: Create eic_dummy_content_type module.

lramarojaona:
### Tests

- [x] Enable `eic_dummy_content_type` module
- [x] You should have a `Dummy` content type created

### Remarks

- This module is for development purpose only and should not be enabled on PROD


4hardfire (approved):
It seems ok for me! The FE will have all the fields types to style.

PR #1287, opened 2022-03-15
EICNET-2175: Fix overviews access for SA/SCM/DA in Events and Organisations.

lramarojaona:
### Tests

- [x] As TU create an Event (group type)
- [x] As DA, create an Organisation
- [x] As SCM, make sure you can access `/events/<event-name>/people`
- [x] As SCM, make sure you can access `/organisations/<organisation-name>/people`


4hardfire (approved):
Seems ok to me!

PR #1288, opened 2022-03-15
EICNET-2132: TU can not reply on comment in News article (organisations)

4hardfire:
### Changes

- Hide reply button if user does not have access to post comments.

### Tests

- [ ] As TU (non-member), go to an organisation News with open comments
- [ ] Make sure you can view comments but cannot post or reply
- [ ] As DA, add the TU as a member of the organisation
- [ ] As TU (member), go to the organisation news detail page
- [ ] Make sure you can view, post or reply to comments


Dromadoros (approved):
Big react dev here

PR #1292, opened 2022-03-16
[EICNET-1745]fix: set correct statistics + invalidate cache when event like

Dromadoros:
How to test:

- [x] Add news and open comments
- [x] Go to organisation detail and check the news block
- [x] Check if you have views, likes and comment stats as 0 (or 1 for view)
- [x] Like the content, get back to the block news in organisation and verify it correctly increased (without rebuild cache)
- [x] Now try to comment, do the same check
- [x] Close the comments in news and verify the stats dissapeared

### DIFF
--- a/lib/modules/eic_statistics/src/NodeStatisticsDatabaseStorage.php
+++ b/lib/modules/eic_statistics/src/NodeStatisticsDatabaseStorage.php
@@ -6,6 +6,7 @@ use Drupal\Core\Cache\Cache;
 use Drupal\Core\Database\Connection;
 use Drupal\Core\State\StateInterface;
 use Drupal\eic_statistics\Event\PageViewCountUpdate;
+use Drupal\node\Entity\Node;
 use Drupal\statistics\NodeStatisticsDatabaseStorage as CoreNodeStatisticsDatabaseStorage;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\HttpFoundation\RequestStack;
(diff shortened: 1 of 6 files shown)


lramarojaona:
@Dromadoros ,

- Like: works
- Comments: counter is 't updated it stays to 0
- Views: counter isn't updated, it stays to 0

Dromadoros:
> @Dromadoros ,
> 
> * Like: works
> * Comments: counter is 't updated it stays to 0
> * Views: counter isn't updated, it stays to 0

This has been fixed. Can you recheck again?

lramarojaona:
It works, but I believe the comments counter should be hidden only when comments are hidden (different from closed).
The "closed" status still displays the comments on the node, while the "hidden" doesn't show them anymore:
2022-03-16 15_28_48-Edit News article My news _ EIC-community-D8

lib/modules/eic_statistics/src/NodeStatisticsDatabaseStorage.php
```diff
@@ -65,6 +66,9 @@ public function recordView($id) {
     $this->eventDispatcher->dispatch($event, PageViewCountUpdate::EVENT_NAME);
     // On each stat update we invalidate the stat cache for the given node and all nodes (this one isn't used for the moment)
     Cache::invalidateTags(["eic_statistics:node:$id"]);
+    if ($node = Node::load($id)) {
+      Cache::invalidateTags($node->getCacheTags());
+    }
```
lramarojaona, line 71: Invalidating the node cache on each node view makes the cache useless.
We should either accept that the counter is not always up-to-date or (best approach) include the statistics in a block instead of the node itself

lramarojaona: There a ticket for that so you can merge this PR: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1358

PR #1307, opened 2022-03-18
[EICNET] add template for page 4xx

Askerx9:
### What I did

- add template for page 4xx

### How to test

- http://localhost:8080/dfdfdsfsd
- see the new page
image

### DIFF
--- a/lib/themes/eic_community/templates/layout/page--4xx.html.twig
+++ b/lib/themes/eic_community/templates/layout/page--4xx.html.twig
@@ -16,7 +16,7 @@
         name: 'warning',
         type: 'notifications'
       },
-      description: "Retun in the previous page <a href='#' onclick='history.back()' > here </a>"
+      description: 'Return to the homepage <a href="@front_url">here</a>'|t({'@front_url': front_url})
     }%}
   </div>
 {% endblock %}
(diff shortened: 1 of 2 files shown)


lib/themes/eic_community/templates/layout/page--4xx.html.twig
```diff
@@ -0,0 +1,22 @@
...
+  <div class="ecl-container" style="padding:2rem 0;">
+    {% include "@theme/patterns/compositions/state-banner.html.twig" with {
+      extra_classes: 'ecl-state-banner--is-blue',
+      title: page.content,
+      icon_file_path: eic_icon_path,
+      icon: {
+        name: 'warning',
+        type: 'notifications'
+      },
+      description: "Retun in the previous page <a href='#' onclick='history.back()' > here </a>"
```
filipvt1, line 19: 1. This text should come from the translations
2. I don't think linking to the previous page is a good solution here. If the user lands on a page from google they will be sent back to google. It might be better to link to the homepage here (needs to take the base url into account here)

filipvt1 (changes requested):

lib/themes/eic_community/templates/layout/page--4xx.html.twig
Dromadoros: I agree, for MVP it's maybe better to redirect to the homepage using generate url by Drupal. And we should use translations function from drupal like @filipvt1 said

4hardfire: I already changed that

Dromadoros (approved):

lenglebert:
@Askerx9 here's the issue for this PR: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2210

PR #1308, opened 2022-03-21
[EICNET-688]feat: Create endpoint to create taxonomy_term for entity tree widget

Dromadoros:


Dromadoros:
> In general it works great! Here is some remarks:
> 
> * When I was selecting topics in a discussion, the button to load more topics at the bottom of the entity tree widget doesn't work
> * Adding tags by pressing 'Enter' doesn't work

1. I already created a ticket for Jimmy and I
2. We removed this feature cause the plugin react by default does some sh*t

PR #1310, opened 2022-03-21
EICNET-1378: Automatically unfollow group + group content when user leaves a group

4hardfire:
### Features

- Implement logic to create queue items to update user content follow flags when leaving a group and when a group visibility changes to private;
- Implement drush commands to clear the queues;
- Prevent notifications from being sent when user cannot view the related entity.

### Tests

Before doing the following tests, make sure you enable the logic to decrement the followers in the configuration -> `config/sync/eic_group_membership.settings.yml`. Update to `unfollow_group_content_enqueue_items: true` and `unfollow_group_content_dequeue_items: true`

**Unfollow group:**

- [x] As TU, join a public group
- [x] Make sure you automatically follow the group
- [x] Leave the group
- [x] Make sure you automatically unfollow the group

**Join public group and change visibility:**

- [x] As TU, join a public group
- [x] Make sure you automatically follow the group
- [x] Follow some content in the group
- [x] As SA/SCM, change the group visibility to private or custom restricted
- [x] As TU, leave the group
- [x] Make sure the group has 1 less follower
- [x] Run drush command `unfollow_group_content_from_non_members`
- [x] Make sure the group contents the TU was following have 1 less follower

**Follow group and change visibility:**

- [x] As TU, follow a public group and some content in the group
- [x] As SA/SCM, change the group visibility to private or custom restricted
- [x] Run drush command `unfollow_group_content_on_visibility_change`
- [x] Make sure the group follow counter has decremented
- [x] Make sure the group contents follow counter has decremented as well

### To discuss

There might be some cases where GO wants to temporarily change the group visibility of a group. This will cause all non members of a group to possibilly unfollow the group and all group contents if this feature is enabled by default.


lramarojaona (changes requested):
drush command didn't work

4hardfire:
> drush command didn't work

@lramarojaona Don't forget to update `config/sync/eic_group_membership.settings.yml` before doing the tests.

lramarojaona:
@4hardfire ,
I'm getting this erro when runnin the last drush command:
EIC-1378

4hardfire:
> @4hardfire , I'm getting this erro when runnin the last drush command: EIC-1378

@lramarojaona for some reason when we do a fresh install, anonymous user is following groups and group content (which should not be the case). I've added an extra condition to remove a flag if it's from anonymous user.
Can you re-test? I would suggest to create a new group and re-test the feature.

lramarojaona (approved):
Ok, it works

PR #1312, opened 2022-03-21
[EICNET-1990]fix: Use entity reference autocomplete widget for file document topics

Dromadoros:
How to test it:

- [x] Create a document content type and add file
- [x] Check if you can select multiple topics


4hardfire (approved):
It works!

PR #1313, opened 2022-03-21
[EICNET-2191]fix: Do not authorized the group node overview for all outsider

Dromadoros:
How to test:

- [x] As a TU check if you do not see the /nodes of an organisation (should be 401)

### DIFF
--- a/config/sync/group.role.organisation-4c7e676f7.yml
+++ b/config/sync/group.role.organisation-4c7e676f7.yml
@@ -17,7 +17,6 @@ permissions_ui: false
 permissions:
   - 'access content overview'
   - 'access group content menu overview'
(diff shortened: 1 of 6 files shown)


4hardfire (approved):
It works.
Question: will we need to update the permissions of all organisations in QA?

Dromadoros:
> It works. Question: will we need to update the permissions of all organisations in QA?

Yes but we have the beautiful feature that I develop to select organisation and update it automatically in the admin

4hardfire:
> > It works. Question: will we need to update the permissions of all organisations in QA?
> 
> Yes but we have the beautiful feature that I develop to select organisation and update it automatically in the admin

What a real machine!!!

PR #1314, opened 2022-03-21
[EICNET-2197]fix: Add access to SA SCM to edit their own book

Dromadoros:
How to test it:

- [x] Connect as SCM
- [x] Create a draft book
- [x] You should be able to edit it


4hardfire (approved):
It works!

PR #1324, opened 2022-03-22
[EICNET]fix: remove view unpublished group permission from TU

korhankale:
## To Test

- [x] Send an archival request for an organisation & accept it
- [x] AS TU you should not be able to view its page
- [x] It should not be displayed in the overview page


lramarojaona (commented):
@korhankale , it works, but I4m wondering if archived organisation should remain read-only and not unpublished

I asked the question in our Slack

PR #1328, opened 2022-03-22
EICNET-2190: Related groups does not shift left when certain groups are not visible to that user

4hardfire:
### Improvements

- Skip view rows from being rendered when html is empty.

### Test

- [x] As SA/SCM, edit a public group and add a private group + a public group in the "Related groups" field (**make sure the private group is in the first position**)
- [x] Go to the group homepage and make sure all the related groups are shown
- [x] As TU, go to the group homepage as well and check if you can only view the public group and make sure there is no empty space (**check the ticket description in Jira for a clear understanding about the empty space** -> https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2190)


lramarojaona:
@4hardfire , it works but I think there is a better approach.

The problem is that it seems Views cannot figure out by itself if the user has access to the group.
So we need to do it ourselves, and it would be better to work directly at the source (It's a shame group_flex access doesn't work with Views)

I suggest this:

```php
/**
 * Implements hook_views_pre_render().
 */
function my_module_views_pre_render(ViewExecutable $view) {
  if ($view->id() == 'group_related_groups') {
    foreach ($view->result as $key => $row) {
      /** @var \Drupal\group\Entity\GroupInterface $group */
      foreach ($row->_relationship_entities as $group) {
        if (!$group->access('view')) {
          unset($view->result[$key]);
        }
      }
    }
  }
}
```

lramarojaona (approved):

PR #1330, opened 2022-03-22
EICNET-2220: Add class on CKEDITOR return

4hardfire:
### Improvements

- Add custom css classes to long text fields.

### Tests

- [x] Create a News content type and add text to the body field
- [x] Make sure you have a css class in the html generated for the body field when viewing the News detail page

### DIFF
--- a/config/sync/context.context.group_group_homepage.yml
+++ b/config/sync/context.context.group_group_homepage.yml
@@ -51,12 +51,13 @@ reactions:
         weight: '-2'
         custom_id: views_block_group_discussions_block_1
         theme: eic_community
-        css_class: ''
+        css_class: ecl-section-wrapper--is-white
         unique: 0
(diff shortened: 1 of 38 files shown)


Askerx9:
https://www.figma.com/file/WZH3qhNynnmm0H6FKmD3hi/EASME_v1?node-id=13181%3A48874

### What I did

- Style the CKEditor table

### How to test

- create a discussion with a table in wysiwyg with header and 3 lines and check if the design is ok

4hardfire:
> It's a nice approach, but it seems it doesn't work everywhere, I just created an event (node) and it seems it deosn't go through

@lramarojaona I applied this also for text_with_summary fields. It should be ok now for all group contents.

lramarojaona (approved):
Ok, it works

PR #1331, opened 2022-03-22
[EICNET]chore: refactor the solr search controller

Dromadoros:
Refactor the way we structure the solr search query and move it to a service so we can call it from everywhere.


lramarojaona:
@Dromadoros , I'm getting an error when doing a fresh install:

` [error]  ParseError: syntax error, unexpected 'new' (T_NEW) in Composer\Autoload\includeFile() (line 66 of /app/lib/modules/eic_search/src/Controller/SolrSearchController.php)`

Dromadoros:
Can you test again @lramarojaona. It was due to a rollback issue that I did...

PR #1333, opened 2022-03-23
EICNET-2227: Create a text filter to add divs around tables.

lramarojaona:
### Tests

- [x] Create a content with one more tables
- [x] Edit the filter you re using (e.g. full HTML) and add classes to the filter configuration `/admin/config/content/formats/manage/full_html` in _Wrap tables with div tag_
- [x] Reload your content: you should have a `div` with your classes around the `table` elements


4hardfire (approved):
It works!

PR #1335, opened 2022-03-23
Implement cypress automated test

Dromadoros:
How to test:

- [ ] Set correctly the .env with new cypress_base_url (it needs to be the QA env with username/password in the url for basic auth)
- [ ] Rebuild your docker containers
- [ ] make ssh_cypress
- [ ] npm install
- [ ] npm cypress run
- [ ] You should have 3 tests passed


.gitlab-ci.yml
```diff
@@ -192,3 +193,18 @@ slack_notifier:
       curl -X POST $SLACK_URL -H 'Content-Type: application/json' -d '{"text": "New deployment running on QA. Environment will be updated in a few minutes..."}'
   only:
     - /^release/.*$/
+
+cypress_test_qa:
+  stage: cypress
```
korhankale, line 198: We better put that in the test stage also no?

lib/modules/eic_user/eic_user.module
```diff
@@ -173,6 +173,8 @@ function _eic_user_refresh_user_data(array $accounts) {
   $solr_helper->reIndexEntities($accounts);
 
   foreach ($accounts as $account) {
+    if (!$account instanceof \Drupal\Core\Session\AccountInterface)
+      continue;
```
korhankale, line 177: Missing brackets here

.gitlab-ci.yml
Dromadoros: It was in stage before but I need that changes are deployed to the env before firing auto test

korhankale: indeed 😄

lib/modules/eic_user/eic_user.module
Dromadoros: I never used anymore the brackets for one line but we can say that we standardize it

korhankale (approved):

PR #1339, opened 2022-03-24
[EICNET-1910] Remove field_vocab_topics in media

Dromadoros:
How to test? Just verified that it's correctly removed


4hardfire (approved):
Seems ok to me!

PR #1340, opened 2022-03-25
[EICNET] improve Front end navigation

Askerx9:
### What I did

- I add the [Hotwire/turbo](https://turbo.hotwired.dev/handbook/drive) to the website, now we have an Ajax navigation, no reloading of the whole HTML but just the body.


filipvt1:
I guess this is just the loading of the library. Now it needs to be applied to every individual link where you want to use this? This will be handled in separate tickets?

filipvt1 (approved):

Askerx9:
@filipvt1 
No, when you load the library it works directly, if you check the network console when you click on a link, the css is not reloaded every time.

PR #1344, opened 2022-03-28
EICNET-899: Migration of users (phase 2)

4hardfire:
### Improvements

- Migrate missing user profile taxonomy term fields coming from SMED: **Topics of expertise**, **Topics of interest**, **Regions and Countries**, **Spoken languages** and **Job titles**.
- Fix user roles migration.

### Tests

- Run a fresh install (**NOTE: DO NOT IMPORT CONTENT FROM FIXTURES**)
- Run `drush mim upgrade_d7_file_image_to_media --execute-dependencies`
- Run `drush mim upgrade_d7_user --update` (15932 should be migrated) -> this migration needs to run twice because we need to update the user profile picture after the files have been migrated. 
- Run `drush mim upgrade_d7_user_to_profile` (15932 should be migrated)

### Remarks

- There is still one remaining taxonomy term field to be migrated in the user profile: **User type**
- There is no piwik administrator role in D9, and therefore all users with that role will become trusted_users after the migration

### DIFF
--- a/config/sync/migrate_plus.migration.upgrade_d7_user.yml
+++ b/config/sync/migrate_plus.migration.upgrade_d7_user.yml
@@ -12,7 +12,7 @@ migration_tags:
 migration_group: migrate_drupal_7
 label: 'User accounts'
 source:
-  plugin: d7_user
+  plugin: eic_d7_user_with_smed_ids
   track_changes: true
 process:
   name: name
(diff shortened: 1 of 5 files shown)


lramarojaona (commented):
Ok, it works.
I'm just wondering why the private files were not migrated as it was a dependency for `upgrade_d7_file_image_to_media` (maybe I CTRL-Ced).
Also we should remove the stub as it's generating wrong images (to be checked)

lramarojaona (approved):
Ok, it works!

PR #1346, opened 2022-03-28
[EICNET-1970]feat: Create my drafts overview

Dromadoros:
How to test:

- [x] Go to your activity profile feed
- [x] Check if in subnavigation you see "My drafts"
- [x] Click on it
- [x] You are directed to "My drafts" overview
- [x] You should have at least: search input, topic filter and content type filter
- [x] In results, it should list all of your drafts contents witht the same teaser than global search

### DIFF
--- a/config/sync/context.context.drafts.yml
+++ b/config/sync/context.context.drafts.yml
@@ -1,0 +1,70 @@
+uuid: 2f22f8d7-f8d4-4599-a531-00ce6b9c2f4a
+langcode: en
+status: true
+dependencies:
+  module:
+    - eic_search
+    - route_condition
+label: Drafts
+name: drafts
+group: 'My profile'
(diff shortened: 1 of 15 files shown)


4hardfire (changes requested):
Everything works really great. However, I found an issue concerning draft content of a published node:

- As SA/SCM, I created a new draft discussion
- The draft discussion is shown in "My drafts"
- I published the discussion and is no longer shown in "My drafts"
- I created a new draft from the published discussion
- My new draft is not shown in "My drafts"

I am not sure how hard will be to fix this but what I would suggest is:

- When you process the solr document, we need to know if there is at least a draft version and save that information. Also the search query should look for that information.

4hardfire (approved):
It works like a charm!

PR #1358, opened 2022-03-30
Feature/eicnet 2273 event group invitations

lramarojaona:
### Improvements

- Created a patch to fix the access check in ginvite module

### Tests

- [x] Go to an event as GO/GA or higher, click on Invite users
- [x] Select some users and click submit
- [x] Make sure you don't have an error on the confirmation screen and when sending the invitations


4hardfire (approved):
It works!

PR #1360, opened 2022-03-30
[EICNET-2277]fix: If no fields ID define in source, ignore search

Dromadoros:
How to test it:

- [ ] Go to activities stream of a group
- [ ] Verify that you have results (be sure to have activities on it)


4hardfire (approved):
It works!

PR #1362, opened 2022-03-31
EICNET cypress move login and cookies in command and delete helpers

Askerx9:
### What I did

- add Command _login_ with default user and password
- add Command _preserverCookies_
- remove helpers

Now you can use "cy.login()" to login like cupress user or "cy.login("admin", "admin") to login like an Admin user


Dromadoros (approved):
nice

PR #1364, opened 2022-03-31
EICNET-2279: Patch Drupal core to ignore missing files instead of fail.

lramarojaona:
### Improvements

- Now file migrations do not fail anymore when some files are missing, they are just ignored

### Tests

- [x] Run `drush mim upgrade_d7_file --execute-dependencies`, you should not have failed items nor failed migration
- [x] Run `drush mim upgrade_d7_file_private`, you should not have failed items nor failed migration

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -106,16 +106,16 @@
         },
         {
             "name": "aws/aws-sdk-php",
-            "version": "3.216.4",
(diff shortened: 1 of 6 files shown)


4hardfire (approved):
It works!

PR #1366, opened 2022-04-01
EICNET-1315: Node 'Discussion' migration (phase 2)

4hardfire:
### Improvements

- Migrate regions and countries + related documents in node discussions migration.
- Create field_language in Discussions + add it to the migration.

### Todo

- [x] Update FA -> https://citnet.tech.ec.europa.eu/CITnet/confluence/display/EICNET/05+-+Node+migration

### Test

- [ ] Run a fresh install (**NOTE: DO NOT IMPORT CONTENT FROM FIXTURES**)
- [ ] Run `drush mim upgrade_d7_node_complete_discussion --execute-dependencies` -> 958 node documents migrated (1113 revisions in total)


Dromadoros:
When I tested  it, it was missing the region and countries field in Discussion. So I had an error on the FE cause it's a required field, it should work when it will be migrated from SMED.

So maybe we approve it but we create a ticket to not forget to update this feature once we have region and coutries?

4hardfire:
> When I tested it, it was missing the region and countries field in Discussion. So I had an error on the FE cause it's a required field, it should work when it will be migrated from SMED.
> 
> So maybe we approve it but we create a ticket to not forget to update this feature once we have region and coutries?

I created a new ticket to fix it later -> https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2338.

Dromadoros (approved):

Dromadoros (approved):

PR #1368, opened 2022-04-01
Feature/eicnet 1828 organisation type access

lramarojaona:
### Tests

- [x] Create an organisation with a specific type
- [x] Create a published group with custom visibility => organisation type (the same that you used in the organisation)
- [x] Create a node inside the group and publish it
- [x] As TU (non-member of the organisation) make sure you can't find the group in the overviews/search nor the content and you can't access the group page or the node page
- [x] Now add the TU as GM of the organisation
- [x] Check that you can find the group and the content in the overviews, and check that you can access th detail pages


Dromadoros (approved):
Nice

PR #1370, opened 2022-04-04
EICNET-2196: Update book page URL pattern and add missing variables to the Theme (Rework)

4hardfire:
### Improvements

- Add permission to edit/delete book pages for SA/SCM users.

### Test

- [x] As SA/SCM create a book page
- [x] As another SA/SCM try to edit the previous book page and make sure you have access to do it


lramarojaona (approved):
Ok, it works!

PR #1383, opened 2022-04-06
[EICNET-1970]fix: Prefilter the group book page from search API and add groups in drafts

Dromadoros:
How to test it:

- [x] Create a Group type group, Event type group and Organisation type group as draft
- [x] You should see in your drafts (/user/1/activity/drafts) the group, event and organisation
- [x] You should NOT see the book page automatically created


4hardfire (approved):
It works!

PR #1391, opened 2022-04-07
[EICNET-2313]fix: do not always put parent published to true for group

Dromadoros:
How to test:

- [x] Connect as TU
- [x] Be sure groups in pending/blocked/... exists and you are not the owner
- [x] Go to the groups overview, be sure you cannot see them


4hardfire (approved):
It works!

PR #1393, opened 2022-04-08
Feature/eicnet 191 group migration

lramarojaona:
### Improvements

- Applied a patch to `Core` to ignore/skip missing files so that the file migrations don't fail when files are missing
- Applied a patch to `migrate_tools` module to run migrations as user #1 to avoid permissions issues

### Tests

- [x] Run `drush mim upgrade_d7_node_complete_group --execute-dependencies`
- [x] Check some groups and make sure you have values for: Welcome message, body, topics, regions and countries, images, message to admin, related groups, moderation state
- [x] Check that you have the correct visibility and joining methods
- [x] Check that you have the correct features enabled (Since most of the groups have default config for this, you can check the `AI-Call` group that should have Events and Wiki features disabled

### Remarks

- There is a typo in the ticket number, this should be linked to https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2079
- Best way to compare is to log in the D7 ACC environment: https://eic.accp.eismea.eu/community7/


4hardfire (approved):
Seems ok to me!
Just a small remark:

- Organisation restrictions in custom restricted groups are not migrated. Check example -> `/groups/otp-bank-corporate-day-private-group`. I suppose this is not yet done because it depends on the organisation migrations.

PR #1394, opened 2022-04-08
[EICNET-1698] improve loading style in react component

Askerx9:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1698?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel

### What I did

- React overview loading improved 
- Position of loader improved

### DIFF
--- a/lib/themes/eic_community/react/components/Block/Overview/Search/Results.js
+++ b/lib/themes/eic_community/react/components/Block/Overview/Search/Results.js
@@ -34,8 +34,8 @@ class Results extends React.Component {
   }
 
   render() {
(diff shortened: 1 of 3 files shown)


filipvt1 (changes requested):
Placement still seems a bit strange. It's positioned absolute, but centered across the entire container. Don't know if this needs to be absolute. Shouldn't the loading icon be shown where the loading content is supposed to go?
image

Askerx9:
@filipvt1 you right, now it's better

PR #1395, opened 2022-04-08
EICNET-2315: GO can not access Members Pending page

4hardfire:
### Improvements

- Add permission to administer membership request when joining method is set to "tu_group_membership_request";
- Create hook_update to update group permissions for old groups.

### Test

Using QA DB:

- [ ] Run `drush updb`
- [ ] Make sure the following the GO can access the group membership requests page (joining method set to **Request**) -> `/groups/test-private-group/members-pending`
- [ ] Make GO cannot access the group membership requests page for the following group (joining method set to **Open**) -> `/groups/edegem-v2/members-pending`

On a fresh install:

- [x] As TU, create a new group and set the joining method to "Request"
- [x] Make sure the user can access the group membership requests page -> `/groups/<group-name>/members-pending`

### DIFF
--- a/lib/modules/eic_groups/eic_groups.install
+++ b/lib/modules/eic_groups/eic_groups.install
@@ -7,6 +7,7 @@
 
 use Drupal\Core\Field\BaseFieldDefinition;
 use Drupal\eic_groups\Constants\NodeProperty;
+use Drupal\eic_groups\EICGroupsHelper;
 use Drupal\group_content_menu\GroupContentMenuInterface;
 use Drupal\group_permissions\Entity\GroupPermission;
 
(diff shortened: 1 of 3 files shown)


lib/modules/eic_groups/eic_groups.install
```diff
@@ -186,3 +187,87 @@ function eic_groups_update_9003(&$sandbox) {
...
+          $joining_method['plugin_id'],
+          [
+            'group_membership_request',
+            'tu_group_membership_request',
+          ]
+        )
+      ) {
+        $group_type = $group->getGroupType()->id();
+
+        $ownerGroupRoleId = EICGroupsHelper::getGroupTypeRole($group_type, 'owner');
```
lramarojaona, line 249: Here you could use `EICGroupsHelper::GROUP_TYPE_OWNER_ROLE`

Also believe GA should also be able to manage membership requests

lib/modules/oec_group_flex/src/Plugin/GroupJoiningMethod/TuGroupMembershipRequest.php
```diff
@@ -3,6 +3,7 @@
 namespace Drupal\oec_group_flex\Plugin\GroupJoiningMethod;
 
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\eic_groups\EICGroupsHelper;
```
lramarojaona, line 6: Ideally we should not have a any dependency of EIC modules in OEC modules.

Since you are only using a static method `getGroupTypeRole`, maybe just replicate this method as a local one?

lramarojaona (changes requested):

lib/modules/oec_group_flex/src/Plugin/GroupJoiningMethod/TuGroupMembershipRequest.php
4hardfire: Done!

lib/modules/eic_groups/eic_groups.install
4hardfire: Done!

4hardfire:
@lramarojaona I updated the code. Can you re-test the PR to make sure it's still working?

lramarojaona (approved):
Ok  it works!

PR #1399, opened 2022-04-11
News & Stories - fix inconsistenties FE vs design

Askerx9:
https://citnet.tech.ec.europa.eu/CITnet/jira/secure/RapidBoard.jspa?rapidView=6628&projectKey=EICNET&view=detail&selectedIssue=EICNET-2105&quickFilter=26822

### What I did

- edit editorial header style
- edit featured content title style in sidebar
- Removed 'view bookmark content' from anonymous users

### DIFF
--- a/config/sync/user.role.anonymous.yml
+++ b/config/sync/user.role.anonymous.yml
@@ -29,7 +29,6 @@ permissions:
   - 'unflag like_media'
   - 'unflag recommend_group'
   - 'view banner fragments'
-  - 'view bookmark_content'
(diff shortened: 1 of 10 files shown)


filipvt1 (changes requested):
The public tag is positioned correctly but not styled right. It's just a grey text. This should be colored depending on the state of the article (green background for public).
image

Topics title still looks different than the other titles:
image
vs in the design:
image

AlexisWalravens (approved):
👍  Front-end wise

PR #1415, opened 2022-04-13
[EICNET-2035]feat: Send notification reminder to all TU that didnt approve group invite

Dromadoros:
How to test:

- [x] Do multipe group invite
- [x] Run cron
- [x] You should receive a notification for each member who has been invited but didnt accepted/rejected yet

PS: you have a settings to configure if you want to run cron multiple times

### DIFF
--- a/lib/modules/eic_groups/src/Hooks/CronOperations.php
+++ b/lib/modules/eic_groups/src/Hooks/CronOperations.php
@@ -12,9 +12,14 @@ use Drupal\Core\Site\Settings;
 use Drupal\Core\State\StateInterface;
 use Drupal\eic_groups\GroupsModerationHelper;
 use Drupal\eic_messages\Service\MessageBus;
+use Drupal\ginvite\Plugin\GroupContentEnabler\GroupInvitation;
 use Drupal\group\Entity\Group;
+use Drupal\group\Entity\GroupContent;
 use Drupal\group\Entity\GroupInterface;
+use Drupal\message\Entity\Message;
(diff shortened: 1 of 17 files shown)


lib/modules/eic_groups/src/Hooks/CronOperations.php
```diff
@@ -267,4 +272,60 @@ private function processGroupWaitingApprovalReminder() {
...
+  private function processGroupInvitationsReminder() {
+    // Value returned is timestamp.
+    $last_reminder_time = $this->state->get('last_cron_group_invite_time', 0);
+    $now = time();
+
+    if (0 < ($last_reminder_time + Settings::get('cron_interval_group_invite_time', 86400)) - $now) {
+      return;
+    }
+
+    $query = $this->database->select('group_content_field_data', 'gc_fd');
```
lramarojaona, line 287: In fact the `cron_interval_group_invite_time` setting should be checked against the invitation creation date.

Because for now I receive both the invitation and the reminder at the same time, where users should receive the reminder after the defined time.

lramarojaona (changes requested):

lib/modules/eic_groups/src/Hooks/CronOperations.php
Dromadoros: That's because on our case in local, it was the first time ever this cron run so it's triggered directly on the same cron. But in other env, it will be trigger each interval time and not the same time than the group invite

PR #1420, opened 2022-04-14
[EICNET-2091] add last page in pagination on mobile

Askerx9:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2091

### What I did

- Now on mobile, we can always see first page and last page


AlexisWalravens (changes requested):

AlexisWalravens (approved):
LGTM now 👍

PR #1422, opened 2022-04-14
[EICNET-2037]feat: Update message for group/event request

Dromadoros:
How to test it:

- [x] Create a group/event as TU, it will be as pending
- [x] As SA/SCM approve the group pending
- [x] Run cron
- [x] You should receive an email following this:
https://www.figma.com/file/j583oU3vO3YfrU3SnZp4vQ/EIC---Analysis?node-id=324%3A11736


4hardfire (changes requested):
It works but when you create a group (not group type event) the text mentions the group as if it was an Event. We need to show the group bundle using tokens instead of hardcoding it as "Event".

korhankale (approved):

PR #1424, opened 2022-04-14
[EICNET-2232]feat: Send message correctly formated for contact user

Dromadoros:
How to test:

- [x] Go to user/ID/contact
- [x] send a message
- [x] run cron
- [x] You should receive an email following template


4hardfire (changes requested):
The email received by the user doesn't look ok to me. For some reason the subject is not fully shown (only the 1st letter) + the body text (of the contact form) is not shown in the notification. Is there something missing?

image

4hardfire (approved):
It works :)

PR #1430, opened 2022-04-15
EICNET-2281: Not authorized and wrong link in status message after creating gallery

4hardfire:
### Regression fixes

- Decorate pathauto generator service in order to prevent generating URL aliases for group content nodes before the group content relationship is created.

### Test

- [x] As TU, create a new content in the group
- [x] Make sure the link to the content detail page in the status message redirects you to the correct page. (**Before it was redirecting to a wrong URL which would throw a page not found**)


lramarojaona (approved):
It works

PR #1443, opened 2022-04-20
[EICNET-1935] return to page 1 if filter change the number on elements

Askerx9:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1935

### What I did

- in overview, when the filter changes the number of elements, we go back to the page 1


AlexisWalravens (approved):
LGTM front-end wise 👍

PR #1447, opened 2022-04-21
EICNET-1441: Enable Gallery and Document CTs for Organisations.

lramarojaona:
Temporary workaround for content membership migration, we'll need to decide what to do with those contents if we don't want to enable those content types for Organisations


lramarojaona:
This workaround was agreed: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1538?focusedCommentId=5412539&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-5412539

PR #1455, opened 2022-04-25
[EICNET-2015]fix: On reset, reset the state drupal + state to false if smed id group migration

Dromadoros:


lramarojaona (approved):
Ok, it works!

PR #1467, opened 2022-04-26
[EICNET-2022]feat: send notification when group request has been denied

Dromadoros:
How to test it:

- [x] As TU create a group
- [x] Connect as SA/SCM
- [x] Edit the group and add revision log message for the reason then click the workflow button "Refuse" 
- [x] Now run cron
- [x] GO should receive an email saying his group has been denied
- [x] Verify that the state of the group stay as pending


4hardfire (changes requested):
Everything works pretty well.
Just a small remark: after refusing the group, when I try to edit as GO I see the button "Create new group". Shouldn't it be "Update group"?

PR #1473, opened 2022-04-28
[EICNET-2390]fix: do not show links for archived group

Dromadoros:
How to test:

- [x] Archive a group
- [x] Go into it NOT AS SUPER USER but as GO/SA/SCM
- [x] You should not see "post content" and group management operations except the request delete


4hardfire (changes requested):
As GO I get access denied when accessing the archived group. Can you have a look?

Dromadoros:
Should be ok now, I pushed to workflow archived state published to true

4hardfire (commented):
As GA (not SA/SCM), I can now access the archived group. However, I cannot see the "Group managment" dropdown and the last step of your test scenario says I should be able to request delete or archival. Is this only for SA/SCM users?

Dromadoros:
You're right dude, it's handled now in the canRequest method!

4hardfire (approved):
It works!

PR #1477, opened 2022-04-28
[EICNET-2174]feat: Always allow access for latest activity

Dromadoros:
How to test:

- [x] Go as anonymous user to a public group and check that you can see the activity stream


4hardfire (changes requested):
As anonymous user I cannot access the group "Latest activity". Can you have a look?

Dromadoros:
I've forgot to send the config files. Be sure to refresh group permissions with the custom admin action to sync group permission

Normally now should be ok

4hardfire (approved):
It works!

PR #1479, opened 2022-04-29
[cypress] add regretion test to check user invitation access

Askerx9:
### What I did

- create a cypress test to avoid reproducing this regression bug.


e2e/cypress/integration/user/invitation.spec.js
```diff
@@ -0,0 +1,29 @@
+// import { login, preserveCookie } from "../../../helper/login";
```
Dromadoros, line 1: Can we remove this line and also remove empty lines before the context()?

PR #1491, opened 2022-05-02
[EICNET-2022]feat: send notification when group request has been denied

Dromadoros:
How to test it:

- [x] As TU create a group
- [x] Connect as SA/SCM
- [x] Edit the group and add revision log message for the reason then click the workflow button "Refuse" 
- [x] Now run cron
- [x] GO should receive an email saying his group has been denied
- [x] Verify that the state of the group stay as pending

### DIFF
--- a/config/sync/core.entity_form_display.group.group.default.yml
+++ b/config/sync/core.entity_form_display.group.group.default.yml
@@ -14,13 +14,12 @@ dependencies:
     - field.field.group.group.field_vocab_topics
     - field.field.group.group.field_welcome_message
     - group.type.group
-    - workflows.workflow.groups
   module:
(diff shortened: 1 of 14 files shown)


Dromadoros:
Comment form Bernardo:

https://github.com/EIC-EA/EIC-community-D8/pull/1467#pullrequestreview-958832530

4hardfire (changes requested):
As TU, when I try to create a group for the first time the save button has the text "Update group". This should only happen when the group already exists and not during the group creation.

Dromadoros:
For the moment, I removed the workflow buttons, now it's again the select field with states cause it's kind of broken

PR #1495, opened 2022-05-02
[EICNET-2414]fix: If anonymous user do not show post content in sidebar

Dromadoros:
How to test:

- [x] Go as anonymous in a public group
- [x] Go to the Library overview
- [x] You should not see the "Post content" in the left sidebar


4hardfire (approved):
It works!

PR #1507, opened 2022-05-03
EICNET-XXXX: Fix comments migration.

lramarojaona:
### Tests

- [ ] Run `drush mim upgrade_d7_comment`
- [ ] You should have the migration that runs completely without any fail

### Remarks

- I don't know why but when the migration starts, it says 1126 items. Then migration is completed it says `[notice] Processed 1127 items (1125 created, 2 updated, 0 failed, 0 ignored) - done with 'upgrade_d7_comment'`. And when I check the migration result in the UI, it says 1126/1126 imported. But this deosn't the migration to be successful with all content


4hardfire (approved):
Code seems ok to me!

PR #1515, opened 2022-05-04
[EICNET-1892]fix: Improvide title & og_title tags for all contents

korhankale:
## To Test

- [x] Create a document & check that title & og_title meta tags are [page:title] | [site:name]
- [x] Do a quick check for any node also


lramarojaona (approved):
Ok!

PR #1519, opened 2022-05-04
Feature/eicnet 2378 files migration fix

lramarojaona:
### Improvements

- Media are not imported anymore if file is not found
- Entities that reference media that were not migrated don't create an empty row anymore


4hardfire (approved):
Code seems ok!

PR #1522, opened 2022-05-04
EICNET-2435: Migrate Flags

4hardfire:
### Features

- Implement Flag migrations: Follow group, Follow content, Follow event, Highlight content, Like group, Like content

### Tests

- [ ] Run `drush mim --tag=Flags`

image


lramarojaona (approved):
It works but I have 2 suggestions:

- We could also import the creation date (`timestamp`) of the flags. I know it sounds a bit useless at first sight but for the future it can help if we are debugging specific cases
- It would be nide also to have `included_flags` instead of `excluded_flags`, so we know exactly what's being migrated

4hardfire:
> It works but I have 2 suggestions:
> 
> * We could also import the creation date (`timestamp`) of the flags. I know it sounds a bit useless at first sight but for the future it can help if we are debugging specific cases
> * It would be nide also to have `included_flags` instead of `excluded_flags`, so we know exactly what's being migrated

I created a new migration for the follow_event flags and those flags are being excluded from the follow_node flag migration. I also updated the flag migrations according to your suggestions.

lramarojaona (changes requested):
I'm getting an error when runnnig the upgrade_d7_flag_follow_node migration:
```
web@3c3394cd32c4:/app$ drush mim upgrade_d7_flag_follow_node
 [notice] Migrations import started.
 [error]  Query condition 'flagging.entity_id IN ()' cannot be empty.
 [notice] Migrations import ended.
```

lramarojaona (approved):
Ok, now you should just change the migration labels, as they are all called "Flags" for now.

PR #1524, opened 2022-05-04
[EICNET-2426] test to see if comments section is showed

Askerx9:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2426

### What I did

- create tests to see if comments section is showed in some cases


e2e/cypress/integration/group/discussion/comments.spec.js
```diff
@@ -0,0 +1,29 @@
+// import { login, preserveCookie } from "../../helper/login";
```
korhankale, line 1: Let's remove that line if not needed

korhankale (approved):

PR #1525, opened 2022-05-04
Feature/eicnet 2442 fix media borwser filter

lramarojaona:
### Tests

- [x] Login with user A (whatever role)
- [x] Add discussion and add a media
- [x] Make sure you can access to your own files only
- [x] Click on "Table" view
- [x] Make sure you can access to your own files only
- [x] Revert back to "Grid" view
- [x] Make sure you can access to your own files only
- [x] Repeat the same with user B

### DIFF
--- a/config/sync/field.field.node.discussion.field_related_downloads.yml
+++ b/config/sync/field.field.node.discussion.field_related_downloads.yml
@@ -4,6 +4,7 @@ status: true
 dependencies:
   config:
     - field.storage.node.field_related_downloads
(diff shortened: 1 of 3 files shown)


lramarojaona:
> It works. However, there is an issue when changing to the "Table" view as TU so I had to test that as Drupal admin. Did you also have this issue?

Indeed, I created a followup ticket: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2467

PR #1532, opened 2022-05-06
EICNET-2302: Enable and configure oe_webtools_analytics module.

lramarojaona:
### Remarks

- Nothing to test now


config/sync/oe_webtools_analytics.settings.yml
```diff
@@ -0,0 +1,5 @@
+_core:
+  default_config_hash: 3SXRmjEUDj6eUAa5pnmocsRtrjB2AS1_DQKBKZUY8Kw
+siteID: '3'
+sitePath: 'eic.eismea.eu/community '
```
korhankale, line 4: Shouldn't we ignore these settings also?

lramarojaona: I don't think so, I think they will follow statistics on PROD only and I believe that the tracking works only on whitelisted domains

PR #1534, opened 2022-05-09
[EICNET-2043]feat: Add most active for groups + fix overview block config

Dromadoros:
…

### DIFF
--- a/lib/modules/eic_search/src/Search/DocumentProcessor/DocumentProcessorInterface.php
+++ b/lib/modules/eic_search/src/Search/DocumentProcessor/DocumentProcessorInterface.php
@@ -11,7 +11,7 @@ use Solarium\QueryType\Update\Query\Document;
  */
 interface DocumentProcessorInterface {
 
-  public const SOLR_MOST_ACTIVE_ID = "its_most_active_total";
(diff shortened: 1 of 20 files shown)


lib/modules/eic_search/src/Search/DocumentProcessor/DocumentProcessorInterface.php
```diff
@@ -11,6 +11,8 @@
  */
 interface DocumentProcessorInterface {
 
+  public const SOLR_MOST_ACTIVE_ID = "its_most_active_total";
```
korhankale, line 14: I'd call it somethine like 'its_activity_score', isn't that better?

lib/modules/eic_search/src/Search/DocumentProcessor/ProcessorGroup.php
```diff
@@ -0,0 +1,107 @@
...
+    $group_id = $fields['its_group_id_integer'] ?? NULL;
+
+    if (!$group_id) {
+      return;
+    }
+
+    $group = Group::load($group_id);
+
+    if (!$group instanceof GroupInterface)
+    {
```
korhankale, line 64: the bracket on a new line isn't consistent with the others

korhankale (approved):

PR #1536, opened 2022-05-10
[EICNET-2396]chore: Remove node cache invalidate from record view

korhankale:


Dromadoros:
Need to add the correct cache tag to the related news and stories (you can find it in group detail page)

Dromadoros (approved):

PR #1541, opened 2022-05-10
EICNET-1796 Front-end content form for the group type Event

4hardfire:
### Improvements

- Add horizontal tabs to the group event form display.

### Test

- [ ] Try to create a new global event as TU and make sure the fields are displayed according to https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1796


Askerx9 (approved):
FE ok

PR #1542, opened 2022-05-10
EICNET-1797: Front-end content form for the group type Organisation

4hardfire:
### Improvements

- Re-order organisation fields in form display + add more tabs.

### Tests

Test

- [ ] Try to edit an organisation as TU (Group owner) and make sure the fields are displayed according to https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1797

### DIFF
--- a/config/sync/core.entity_form_display.group.organisation.default.yml
+++ b/config/sync/core.entity_form_display.group.organisation.default.yml
@@ -42,15 +42,17 @@ third_party_settings:
     group_organisation:
       children:
         - group_main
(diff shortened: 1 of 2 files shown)


Askerx9 (changes requested):
image
maybe turnover can be a number field ?

4hardfire:
> image maybe turnover can be a number field ?

We will leave the field as a text field.

Askerx9 (approved):

PR #1543, opened 2022-05-10
EICNET-2437: Front-end content form for the content type Event

4hardfire:
### Improvements

- Update node event form display (new tabs + update widgets).

### Test

- [ ] As TU (member), try to create an event in a group and make sure the fields are displayed according to https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2437


Askerx9 (approved):
FE ok

PR #1544, opened 2022-05-10
EICNET-1793: Front-end content form for the content type News

4hardfire:
### Improvements

- Update node news form display (new tabs + update widgets);
- Fix issue where comment settings field was not shown in the correct field group.

### Test

- [ ] As TU, try to create a News article and make sure the fields are displayed according to https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1793


Askerx9 (changes requested):
image

We have a few empty items in the news form, can you check what they are?

4hardfire:
> image
> 
> We have a few empty items in the news form, can you check what they are?

It's fixed. There is still a div related to the vertical tabs that is shown for Admins. But I think the easiest way to hide it is via javascript when there are no tabs. Do you think ou can do that?

AlexisWalravens (approved):
Approved: Front-end wise.

PR #1547, opened 2022-05-10
EICNET-1798: Front-end content form for the content type Discusssions

4hardfire:
### Improvements

- Update node discussion form display (new tabs + update widgets).

### Test

- [ ] As TU (member), try to create a discussion in a group and make sure the fields are displayed according to https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-1798


Askerx9 (approved):
FE ok

PR #1562, opened 2022-05-12
[EICNET-2045]fix: Resync dependant entities when adding comment

Dromadoros:
How to test:

- [x] Go to group members, see the current its_most_active_score
- [x] Add comment on discussions with current user
- [x] See if it has been updated


4hardfire (approved):
Seems to work!

PR #1565, opened 2022-05-13
Feature/eicnet 2300 invite user action form

lramarojaona:
### Tests

- [x] Go a group and click on "Invite users"
- [x] Check that there is no group header block
- [x] Check that you have a title and description block
- [x] Select some users, then put a non-valid email address in the textarea and submit
- [x] Check that your preselection of users still remains

### To do

- [ ] Modify "Back to group" button: add `ecl-button--ghost` class and remove `ecl-button--primary`
- [ ] Put "Back to group" button: after the submit button

### Remarks

- Not all of the FA in the ticket is implemented (e.g. button labels are customisable through Translation UI)


4hardfire (approved):
The back-end seems ok to me!

PR #1577, opened 2022-05-16
EICNET-2461: Sharing content to another group should be filtered on the availability of that CT in that specific group

4hardfire:
### Improvements

- Prevent members from sharing content in groups with content type plugin disabled;
- Improvements to the share group service and controller.

### Test

- [x] As TU (member of public groups and organisations), create a new gallery in a group
- [ ] Try to share the gallery with another group and make sure your organisations are not shown.
- [x] Try to create an event in the group.
- [x] Try to share the event with another group and make sure all of your organisations are shown.

### DIFF
--- a/lib/modules/eic_groups/modules/eic_share_content/eic_share_content.routing.yml
+++ b/lib/modules/eic_groups/modules/eic_share_content/eic_share_content.routing.yml
@@ -15,7 +15,7 @@ eic_share_content.share:
       group:
         type: entity:group
 eic_share_content.groups:
-  path: '/share/group/{group}/user/{user}/groups'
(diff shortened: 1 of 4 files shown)


lramarojaona (approved):
Good, just FYI, I allowed the galleries in Ogranisations because come galleries in D7 belong to companies. But the feature remains not avilable for galleries

4hardfire:
> Good, just FYI, I allowed the galleries in Ogranisations because come galleries in D7 belong to companies. But the feature remains not avilable for galleries

Ok thanks for letting me know. In any case, if we enable the galleries on Organisations we would be able to share.

PR #1579, opened 2022-05-17
Feature/eicnet xxxx various migration fixes

lramarojaona:
### Improvements

- Added the `bacth_size` with 100 as a value for all migration

### Remarks

- There are no specific tests he to run, we need to run the whole migration and see if it improves the stability


4hardfire (approved):
Seems ok to me!

PR #1582, opened 2022-05-17
EICNET-2466: Member statistic for organizations does not show the accurate number

4hardfire:
### Fixes

- Fix group statistics for group type Event and Organisation.

### Test group statistics in Events and Organisations

- [ ] As TU, try to join an Event and Organisation
- [ ] Make sure the number of members in the group header shows the correct number
- [ ] Try to add documents and make sure the number of files is updating correctly
- [ ] Try to add some comments in discussions and make sure the number of comments is updated

### DIFF
--- a/lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.module
+++ b/lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.module
@@ -22,13 +22,15 @@ function eic_group_statistics_group_delete(EntityInterface $entity) {
  * Implements hook_ENTITY_TYPE_insert().
  */
 function eic_group_statistics_group_content_insert(EntityInterface $entity) {
+  $group_type = $entity->getGroup()->bundle();
+
   switch ($entity->bundle()) {
-    case 'group-group_membership':
-    case 'group-group_node-discussion':
-    case 'group-group_node-document':
-    case 'group-group_node-event':
-    case 'group-group_node-gallery':
-    case 'group-group_node-wiki_page':
+    case "$group_type-group_membership":
+    case "$group_type-group_node-discussion":
+    case "$group_type-group_node-document":
+    case "$group_type-group_node-event":
+    case "$group_type-group_node-gallery":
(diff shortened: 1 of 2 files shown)


lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.module
```diff
@@ -22,13 +22,15 @@ function eic_group_statistics_group_delete(EntityInterface $entity) {
...
-    case 'group-group_node-document':
-    case 'group-group_node-event':
-    case 'group-group_node-gallery':
-    case 'group-group_node-wiki_page':
+    case "$group_type-group_membership":
+    case "$group_type-group_node-discussion":
+    case "$group_type-group_node-document":
+    case "$group_type-group_node-event":
+    case "$group_type-group_node-gallery":
+    case "$group_type-group_node-wiki_page":
```
lramarojaona, line 33: We shoud be careful here, as if the plugin ID is too long, then the bundle will end up with a "hashed" name such as `group_content_type_7f62828e3a84b`.

E.g. if we add a new CT with a long machine name, this apporach might not work for that one.

We should put a @todo comment to replace this with `$group_type->getContentPlugin($plugin_id)` where $group_type is the group type entity.

lramarojaona: Maybe we could have a helper function that provides the mapping for each group type and installed plugin:

- group-group_node-long_content_type => group-group_node-long_content_type
- organisation-group_node-long_content_type => group_content_type_7f62828e3a84b

lramarojaona (changes requested):

lib/modules/eic_statistics/modules/eic_group_statistics/eic_group_statistics.module
4hardfire: I found out a different approach which works with the real machine names even if the CT machine name is too long. I've tested and it worked.

lramarojaona (approved):

PR #1585, opened 2022-05-18
EICNET-2467: Media Library: access denied as TU

4hardfire:
### Fixes

- Apply patch to fix media library widget accesses for users other than drupal admin.

### Test

- [ ] As TU, try to create a discussion in a group
- [ ] Try to upload a media in the field "Downloads" and make sure you can switch between "Grid" and "Table" view when selecting the media via the media library widget


lramarojaona (approved):
It works!

PR #1590, opened 2022-05-18
EICNET-1982: TU can access "all entities for (organisation here)"

4hardfire:
### Fixes

- Fix access check on group membership delete
- Remove permission to access all group content overview in organisations.

### Test

Note that this will work only for new organisations. Old organisations will need to rebuild permissions.

- [ ] As TU (non-member), try to access organisation group content overview page -> `/organisations/organisation-name/content`


lramarojaona (approved):
As discussed I couldn't reproduce the intial issue, but the goal is achieved.

PR #1597, opened 2022-05-20
[EICNET-2045]feat:Resync also owner of the content when new content

Dromadoros:
How to test:

- [x] As a user, create a new group content
- [x] Run cron (as DA)
- [x] check if your activity_score has been updated in the /people (check solr field from request response)


4hardfire (changes requested):
It doesn't work until I manually re-index solr

4hardfire (approved):
After increasing the queue worker time and do some extensive testing, it seems to be working properly. The results get updated after running cron.

PR #1602, opened 2022-05-23
Feature/eicnet 2487 searc index sanitize text

lramarojaona:
### Improvements

- Added regexes replacement prior to indexing to strip out `src` attribute for inline images.

### Tests

- [x] Run `drush mim upgrade_d7_node_complete_discussion --idlist=11701:13188:und --update`. Once migrated, this should trigger the item indexing
- [x] Make sure you don't see any Search API related error


4hardfire (approved):
No errors during migration.

PR #1603, opened 2022-05-23
[EICNET-1173] Do not display the social share button on private contents

korhankale:
## To Test

- [x] Create a private group & global event
- [x] Create a content on the group also (discussion, gallery, video)
- [x] Check that social share links are not displayed on the group header but also actions at content level


4hardfire (approved):
It works. However, the social share hasn't yet been implemented for discussions.

PR #1604, opened 2022-05-23
EICNET-899: Create the pathauto pattern for users.

lramarojaona:
### Improvements

- Created a pathauto pettern for users.


4hardfire (approved):
Seems ok to me!

PR #1605, opened 2022-05-23
EICNET-2376: As a user, I want a way to disable my personal contact form so that I can prevent other users to contact me

4hardfire:
### Changes

- Remove duplicated field that enables/disables personal contact form;
- re-factor custom search API processor and user preprocess functions.

### Test

- [x] As SA/SCM, edit the account information of a GM and uncheck the "Personal contact form" option. Also make sure the radio button field "**Members are allowed to contact you through your personal contact form**" has been removed (we only need the "Personal contact form" checkbox)
- [x] Make sure the button "Direct message" is not visible in the user profile page
- [x] Make sure the contact icon is not shown in any user overviews where this GM is shown (**global members overview** and **group members overview**)
- [x] Try to enable the "Personal contact form" again and make sure the contact icon is now visible and you can contact the user via the profile page


lramarojaona (changes requested):
It works but I have 2 remarks:

- I guess you forgot to remove the `field_user_can_contact` field as it is still present in the user edit form
- You should (not remove the function) empty the update hook to avoid breaking on fresh installs. Don't know if it would happen, but still it doesn't make sense to keep it if we don't have the field anymore: https://github.com/EIC-EA/EIC-community-D8/blob/develop/lib/modules/eic_user/eic_user.install#L6

4hardfire:
> It works but I have 2 remarks:
> 
> * I guess you forgot to remove the `field_user_can_contact` field as it is still present in the user edit form
> * You should (not remove the function) empty the update hook to avoid breaking on fresh installs. Don't know if it would happen, but still it doesn't make sense to keep it if we don't have the field anymore: https://github.com/EIC-EA/EIC-community-D8/blob/develop/lib/modules/eic_user/eic_user.install#L6

Done!

lramarojaona (approved):
Ok, it works, I'm just not sure about existing environments if you remove the udpate hook.
If someone needs an update hook it won't run since the same number has already been used on existing environments.

4hardfire:
> Ok, it works, I'm just not sure about existing environments if you remove the udpate hook. If someone needs an update hook it won't run since the same number has already been used on existing environments.

If needed we can reset the update hook number in the database.

PR #1613, opened 2022-05-25
EICNET-2031: Group type notification - group invitation (anonymous user notification)

4hardfire:
### Fixes

- Force using eic_community theme on email templates (ginvite emails were using the admin theme).

### Test

- [x] As SA/SCM, go to a group and try to invite an anonymous user
- [x] Make sure a notification has been sent to the user and it is styled


lramarojaona (approved):
It works!

PR #1614, opened 2022-05-25
EICNET-2164: Group permission fix

lramarojaona:
### Improvements

- Create the Group Permission entity when a new group is created through code (fixtures and webservice)

### Tests

- [ ] Run `drush fixtures:reload OrganisationGenerator` (make sure `eic_default_content` module is enabled)
- [ ] Check `group_permission` table to see if entries have created for the generated groups


4hardfire (changes requested):
There is still an issue where the permission to access group overviews are not set and as TU I can only see the Overview + About links in the group.

PR #1622, opened 2022-05-31
Feature/eicnet 191 global events migration rework

lramarojaona:
### Improvements

- Added a source plugin to migrate paragraph items

### Tests

- [ ] Run `drush mim upgrade_d7_node_complete_event_site --update`
- [ ] Make sure you have the text paragraphs that are migrated

### Remarks

- The text paragraphs are migrated to not lose content, but we do nothing with it for now
- The destination paragraph type deosn't allow formatted text, we'll need to decide what to do with it

### DIFF
--- a/config/sync/core.entity_form_display.group.event.default.yml
+++ b/config/sync/core.entity_form_display.group.event.default.yml
@@ -3,6 +3,7 @@ langcode: en
 status: true
 dependencies:
   config:
+    - field.field.group.event.field_additional_content
     - field.field.group.event.field_body
     - field.field.group.event.field_date_range
     - field.field.group.event.field_document_additional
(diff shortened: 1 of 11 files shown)


4hardfire (changes requested):
You forgot to add the command `drush mim upgrade_d7_paragraph_text` in the test scenario before running the global event site migration. In any case, I ran both migration and it worked fine. However, there is still a change to make to the body of the paragraph because currently is not a full text format and the html tags are shown as plain text. See image bellow:
image

lramarojaona:
@4hardfire ,
You're right I forgot the paragprahs migration command
Regarding the HTML I created a followup ticket: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2516
However, this was already stated in my remarks, that we need to decide, for now the goal is to migrate the content to make sure we don't lose it.

4hardfire (approved):

PR #1623, opened 2022-05-31
EICNET-2512: Ignore duplicated entries when migrating path redirects from URL alias

4hardfire:
### Improvements

- Fix URL redirect migration in order to skip duplicated rows instead of throwing an error.
- Create new processor plugin to search for duplicated redirects.

### Test

- [x] Run drush mim upgrade_d7_node_with_group_url_alias_to_path_redirect --update
- [x] Check if remaining redirects of group content includes also group prefix from D7 PURL module. In the D9 try to go to `/amorph-systems_93/documents/press-release-partnership-announcement` and make sure you get redirected to `/organisations/amorph-systems-gmbh/library/press-release-partnership-announcement`


lramarojaona (approved):
It works!

PR #1640, opened 2022-06-07
[EICNET-2497]feat: Remove field language in wiki page

Dromadoros:
How to test:

- [ ] Go to a detail wiki page and verify you do not see the language field anymore


4hardfire (commented):
The removal seems ok but who decided this? If we need to do this it means the field will not be migrated from D7.

PR #1642, opened 2022-06-07
[EICNET-2521]fix: Do not show flag actions on event group type

Dromadoros:
How to test it:

- [x] In group Event, do not show flag actions on the sidebar, we already show it in the group header


4hardfire (changes requested):
The share button is still shown for public events. Can you have a look?
image

PR #1648, opened 2022-06-08
[EICNET]feat: Introduce group content shares within the digest

korhankale:
## To Test

- [ ] As TU, join & follow a public group
- [ ] As another user, member of another public group, share a content with the group the TU is following
- [ ] Trigger the digest for the TU
- [ ] There should be a content share item within the digest e-mail


4hardfire (approved):
It works like a charm!

PR #1649, opened 2022-06-08
[EICNET-2522]fix: If member is group admin he should see content even if group is restricted

Dromadoros:
How to test:

- [x] Logged in as group admin
- [x] Post a file content
- [x] Go to files library
- [x] Now put group as restricted
- [x] You should be able as group admin to view files in overview

### DIFF
--- a/lib/modules/eic_overviews/src/Controller/GroupOverviewsController.php
+++ b/lib/modules/eic_overviews/src/Controller/GroupOverviewsController.php
@@ -43,6 +43,13 @@ class GroupOverviewsController extends ControllerBase {
   public function access(RouteMatchInterface $route_match, GroupInterface $group, AccountInterface $account) {
     $overview_perm = NULL;
 
+    // First check if user has access to the group itself.
+    if (!$group->toUrl()->access($account)) {
+      return AccessResult::forbidden()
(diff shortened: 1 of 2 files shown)


lib/modules/eic_overviews/src/Controller/GroupOverviewsController.php
```diff
@@ -43,6 +43,13 @@ public function buildGenericPage(GroupInterface $group) {
   public function access(RouteMatchInterface $route_match, GroupInterface $group, AccountInterface $account) {
     $overview_perm = NULL;
 
+    // First check if user has access to the group itself.
+    if (!$group->toUrl()->access($account)) {
```
4hardfire, line 47: Shouldn't it be something like this `!$group->access('view', $account)`?

4hardfire (changes requested):

lib/modules/eic_overviews/src/Controller/GroupOverviewsController.php
Dromadoros: ... Forgot to add the necessary parameter

4hardfire (approved):
It works!

PR #1658, opened 2022-06-10
EICNET-2495: Members should not see restricted groups on other members profile page if they have no access

4hardfire:
### Imrpovements

- Use search api views to show user profile groups.


Dromadoros:
I replaced the views used for "my_groups/groups" from "global_index_group_content" to "groups". Now a TU can see groups from other people (respecting the visibility)

Dromadoros:
On my side everything is working pretty well. Even events default sorting

Dromadoros (approved):

PR #1668, opened 2022-06-14
EICNET-2458: Fix empty avatar state in the latest activity stream.

4hardfire:
### Improvements

- Fix empty avatar state in the latest activity stream.

### Test

- [ ] View the latest activity stream of a group
- [ ] Make sure the empty avatar picture is shown for members without a picture in the sidebar.

image

### DIFF
--- a/lib/modules/eic_search/src/Plugin/Block/ActivityStreamBlock.php
+++ b/lib/modules/eic_search/src/Plugin/Block/ActivityStreamBlock.php
@@ -5,6 +5,7 @@ namespace Drupal\eic_search\Plugin\Block;
 use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Block\Annotation\Block;
 use Drupal\Core\Block\BlockBase;
+use Drupal\Core\Cache\Cache;
 use Drupal\Core\Datetime\DateFormatter;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\File\FileUrlGeneratorInterface;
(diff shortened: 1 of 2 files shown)


lib/modules/eic_search/src/Plugin/Block/ActivityStreamBlock.php
```diff
@@ -211,10 +212,27 @@ public function build() {
...
+      ],
+    ];
+
+    // Send members data and cache tags.
+    $members = [];
+    if ($show_members) {
+      $members = $this->getMembersData($group);
+      $cache['tags'] = [];
+      foreach ($members as $member) {
+        $cache['tags'] = Cache::mergeTags($member['cache_tags'], $member['cache_tags']);
```
lramarojaona, line 228: I'm getting this error here:
_TypeError: Argument 1 passed to Drupal\Core\Cache\Cache::mergeTags() must be of the type array, null given, called in /app/lib/modules/eic_search/src/Plugin/Block/ActivityStreamBlock.php on line 228 in Drupal\Core\Cache\Cache::mergeTags() (line 52 of core/lib/Drupal/Core/Cache/Cache.php)._

Does it make sense to merge the same array?
`Cache::mergeTags($member['cache_tags'], $member['cache_tags']);`

lramarojaona (changes requested):

lib/modules/eic_search/src/Plugin/Block/ActivityStreamBlock.php
4hardfire: That was a mistake

4hardfire: It should be ok now!

PR #1680, opened 2022-06-17
[EICNET-2548]fix: Add recommend source type and strict private visibility

Dromadoros:
How to test:

- [x] Go to a private group
- [x] Click on the recommend button in the group header
- [x] Be sure you only see member of the current group and not all TU


4hardfire (approved):
It works like a charm!

PR #1689, opened 2022-06-20
EICNET-2228: As TU I cannot reply on certain comments

4hardfire:
### Fixes

- Fix default value for commet.is_soft_delete and comment is_archived values sent to react.

### Test

- [ ] Import DB from QA
- [ ] Go to `/groups/edegem-v2/discussions/discussion-test-comments`
- [ ] Check if some of the old comments can still be replied

Before:

- image

After::

- image

### DIFF
--- a/lib/modules/eic_groups/src/Controller/DiscussionController.php
+++ b/lib/modules/eic_groups/src/Controller/DiscussionController.php
@@ -623,8 +623,8 @@ class DiscussionController extends ControllerBase {
       'edited_time' => $edited_time ?
         $this->t('Edited on @time', ['@time' => $edited_time ?: $created_time], ['context' => 'eic_groups']) :
         NULL,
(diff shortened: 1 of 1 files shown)


lib/modules/eic_groups/src/Controller/DiscussionController.php
```diff
@@ -623,8 +623,8 @@ private function renderCommentElement(CommentInterface $comment): array {
       'edited_time' => $edited_time ?
         $this->t('Edited on @time', ['@time' => $edited_time ?: $created_time], ['context' => 'eic_groups']) :
         NULL,
-      'is_soft_delete' => $soft_deleted,
-      'is_archived' => $archived,
+      'is_soft_delete' => $soft_deleted ?? 0,
```
Dromadoros, line 626: Can we just cast it as integer? (int) null will give 0 so should be ok

wdyt?

PR #1693, opened 2022-06-21
EICNET-900: Recommend content (improvements)

4hardfire:
### Improvements

- Show recommend flag in the 2nd place for groups and content
- Hide recommend flag when user cannot recommend entity
- Fix issue where recommend flag was shown twice in News and Stories
- Fix issue with recommend flag link text was not printed correctly.

### Test

- [ ] As TU, go to a News or a Story content type and make sure the "Recommend" flag is not shown twice and positioned at the 2nd place in the sidebar flags
- [ ] Go to a public group, and make sure the "Recommend" flag is positioned at the 2nd place in the group header
- [ ] Go to any content of a public group (discussion, document, video, gallery, etc.), and make sure the "Recommend" flag is positioned at the 2nd place in the sidebar
- [ ] As DA, go to `/admin/config/regional/translate`, search for "Recommend" and make sure you can translate the "Recommend" text that is used for the flag label

### DIFF
--- a/lib/modules/eic_recommend_content/eic_recommend_content.routing.yml
+++ b/lib/modules/eic_recommend_content/eic_recommend_content.routing.yml
@@ -3,7 +3,7 @@ eic_recommend_content.recommend:
   methods: [ POST ]
   defaults:
     _controller: 'Drupal\eic_recommend_content\Controller\RecommendContentController::recommend'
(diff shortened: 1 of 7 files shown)


Dromadoros (approved):
Nice

PR #1698, opened 2022-06-21
EICNET-2159: Make sure images styles are used where needed

4hardfire:


lramarojaona (changes requested):
Please check my comment in the ticket

lramarojaona:
Also it would be nice to specify `make build-front` in the steps, otherwise we get wrong results :)

4hardfire:
Improvements have been done according to the feedback.

lramarojaona (approved):
This is ok!

PR #1701, opened 2022-06-22
[EICNET-2535]fix: Do not use the search_api_reverse_reference but load correctly the gid from group content

Dromadoros:
How to test:

- [x] Re-index elements
- [x] Take DB from Migration
- [x] Go to /groups/eic-prizes-alumni-network/events and verify you have one event

The field  used in search api was: search_api_reverse_entity_references_group_content__entity_id:gid
But the query behind loaded all gid from group_content without checking group_type so we had 3 different results for the same entity_id


4hardfire (approved):
Seems ok to me!

PR #1702, opened 2022-06-23
EICNET-2542: Post Content shown for TU on Files page in Pending group (Fix cache issues)

4hardfire:
### Improvements

- Add group cache tags to the group library overview block;
- allow post content actions in group library if group is also in draft state.

### Test

- [ ] As TU, create a new group and leave it in pending state
- [ ] Go to the group library and make sure you don't have a "Post content" dropdown in the sidebar
- [ ] As SA/SCM, change the group state to draft
- [ ] As TU (owner), go to the group library once again and make sure you can now see the "Post content" dropdown
- [ ] As TU (non-member), go to the library of public group and make sure you cannot see the "Post content" dropdown
- [ ] As SA/SCM, change the group state to published
- [ ] As TU (owner), go to the group library once again and make sure you can now see the "Post content" dropdown
- [ ] As TU (non-member), go to the library of public group and make sure you cannot see the "Post content" dropdown


lramarojaona (approved):
Ok, it works!

PR #1705, opened 2022-06-23
[EICNET-1664] fix: user image in comment email

AlexisWalravens:
# How to test:

- [ ] Go to a discussion
- [ ] Submit a comment with invited users
- [ ] Run cron jobs
- [ ] Check the email sent in mailhog
- [ ] The user image inside the email should be round and not oval


Askerx9 (approved):
FE ok

PR #1707, opened 2022-06-23
[EICNET-2495]fix: Replace my groups by react component

Dromadoros:
How to test:

- [x] Go to profile user in my groups tab
- [x] Check that you can see correctly groups from the user page


4hardfire (approved):
It looks good to me! However, we still need to fix the "Events" as well. That should be done in the following ticket -> https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2567.

PR #1714, opened 2022-06-24
EICNET-226: Migrate "image" media type (Add default alt text)

4hardfire:
### Fixes

- Fix migration of alt texts for file images to media migration.

### Test

- [x] Run `drush mim upgrade_d7_file_image_to_media --update`
- [x] Go to `/media/3352/edit` and make sure it has an alt text (the same text as the media name)


lramarojaona (approved):
It's a bit weird to have such logic in Constants/Misc, but I guess it's ok,w e should just be careful if there is more and more logic in there we should create a new class.

It works!

PR #1715, opened 2022-06-24
EICNET-2562: Prevent legacy paragraphs to be silently populated.

lramarojaona:
### Tests

- [x] Create a global event
- [x] Edit it and make sure you can submit without error message saying the that the paragraph body field is required
- [x] You can also check the entity in devel and make sure that the `field_additional_content` is not populated


4hardfire (approved):
It works!

PR #1717, opened 2022-06-27
EICNET-390: Allow video medias in News and Stories documents

4hardfire:
### Improvements

- Allow video medias to be uploaded into field documents in News and Story CTypes.

### Test

- [x] As SA/SCM, create a story and a news and make sure you can upload video files in "Documents" field


lramarojaona (approved):
Ok!

PR #1718, opened 2022-06-27
EICNET-975: Show author in News and Story document teasers

4hardfire:
### Improvements

- Show author in News and Story document teasers.

### Test

- [x] As SA/SCM, create a new story and upload multiple documents
- [x] Check the story detail page and make sure author name is presented in each document teaser
- [x] Check the story detail page as anonymous and make sure author name is also presented in each document teaser but without a link to the user profile.


lramarojaona (changes requested):
It works, but we should align:

- Logged in user, it says: "Uploaded by <username>"
- Anonymous user, it only says: "<username>"

I guess we should always have "Uploaded by <username>"

4hardfire:
> It works, but we should align:
> 
> * Logged in user, it says: "Uploaded by "
> * Anonymous user, it only says: ""
> 
> I guess we should always have "Uploaded by "

Fixed!

PR #1720, opened 2022-06-27
EICNET-2447: Don't show events counter if events are not enabled on group type.

lramarojaona:
### Tests

- [ ] Clear the cache
- [ ] Go to a Group detail page
- [ ] Make sure you have events counter in the group header block
- [ ] Go to a Global event
- [ ] Make sure you don't have events counter in the group header block

### DIFF
--- a/lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
+++ b/lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
@@ -91,7 +91,7 @@ class EICGroupHeaderBlock extends BlockBase implements ContainerFactoryPluginInt
   /**
    * The solr search manager service.
    *
-   * @var SolrSearchManager
+   * @var \Drupal\eic_search\Service\SolrSearchManager
    */
   protected $solrSearchManager;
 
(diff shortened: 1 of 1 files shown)


lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
```diff
@@ -319,11 +326,14 @@ public function build() {
           'members' => $group_statistics->getMembersCount(),
           'comments' => $group_statistics->getCommentsCount(),
           'files' => $group_statistics->getFilesCount(),
-          'events' => !empty($results) ? $results['response']['numFound'] : 0,
```
4hardfire, line 322: @lramarojaona Couldn't we use the values from the $group_statistics object like: `$group_statistics->getEventsCount();` ? The number of events are supposed to be calculated already.
However, I found an issue where the number of events are not properly calculated in the group statistics. More specifically in `\Drupal\eic_group_statistics\Hooks\EntityOperations::groupContentInsert()` because there we are retrieving the total of events and increment it to the group statistics table. But we should be incrementing 1 by 1 and not the whole total of group events otherwise we will have wrong results.
Also, the decrement process is not done yet for group events.

4hardfire (changes requested):

lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
lramarojaona: @4hardfire ,

1. I guess we have this because of visibility issues (published/not published), but indeed it's not the same approach as the other statistics, I'll create this as a task in the backlog
2. This seems like a must before go-live, can you create tickets for that?

4hardfire: Here is the ticket for the implementation -> https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2593.

4hardfire (approved):
Ok for me!

PR #1721, opened 2022-06-27
EICNET-2066: 

4hardfire:
### Fixes

- Fix caching issue with highlight flag not shown when a GM becomes an Admin;
- Disable content recommendation when group is not published;
- fix coding standards.

### Test

- [ ] As GO/GA, go to a public group with discussions
- [ ] Make sure the you can recommend discussions
- [ ] Make sure you can highlight discussions. Also make sure GM cannot highlight a discussion.
- [ ] Make sure you can request group archival or group deletion from the group management dropdown
- [ ] As SA/SCM, archive the group
- [ ] As GO/GA/GM, make sure you cannot recommend any content in the group
- [ ] As GO/GA, make sure you can still highlight content in group. Also make sure GM cannot highlight any content.
- [ ] As GO/GA/GM, make sure you cannot request group archival (because the group is already archived
- [ ] As GO/GA, make sure you can still request group deletion from the group management dropdown. GM should not be able to request group deletion when the group is archived.

### DIFF
--- a/lib/modules/eic_groups/src/Access/HighlightGroupContentAccessCheck.php
+++ b/lib/modules/eic_groups/src/Access/HighlightGroupContentAccessCheck.php
@@ -4,14 +4,12 @@ namespace Drupal\eic_groups\Access;
 
 use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Routing\Access\AccessInterface;
-use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\Core\Session\AccountProxy;
 use Drupal\group\Entity\GroupInterface;
 use Drupal\node\NodeInterface;
-use Drupal\user\Entity\User;
 
 /**
- * Class HighlightGroupContentAccessCheck
+ * Provides an access checker for highlighting group content.
  *
  * @package Drupal\eic_groups\Access
  */
@@ -29,11 +27,17 @@ class HighlightGroupContentAccessCheck implements AccessInterface {
   ];
 
   /**
+   * Constructs a new HighlightGroupContentAccessCheck object.
+   *
    * @param \Drupal\Core\Session\AccountProxy $account
+   *   The current user account.
    * @param \Drupal\group\Entity\GroupInterface|null $group
+   *   The group entity.
    * @param \Drupal\node\NodeInterface|null $node
+   *   The group content node.
    *
    * @return \Drupal\Core\Access\AccessResultInterface
+   *   The access result.
    */
   public function access(
     AccountProxy $account,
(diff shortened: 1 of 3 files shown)


lib/modules/eic_groups/src/Access/HighlightGroupContentAccessCheck.php
```diff
@@ -46,7 +50,9 @@ public function access(
 
     $account = $account->getAccount();
     if (!$group->hasPermission('highlight group content', $account)) {
-      return AccessResult::forbidden('highlight group content permission is required');
+      return AccessResult::forbidden('highlight group content permission is required')
```
lramarojaona, line 53: @4hardfire , I think we should stop adding too much cache dependencies if it's not a costly action in terms of calculation. In this case it's just testing a permission, I don't think it's an issue if we don't have cache for this.

lib/modules/eic_groups/src/Access/HighlightGroupContentAccessCheck.php
```diff
@@ -55,10 +61,14 @@ public function access(
     ]);
 
     if (empty($group_content)) {
-      return AccessResult::forbidden('Invalid group content argument');
+      return AccessResult::forbidden('Invalid group content argument')
```
lramarojaona, line 64: Remove cache.

lib/modules/eic_groups/src/Access/HighlightGroupContentAccessCheck.php
```diff
@@ -55,10 +61,14 @@ public function access(
...
+        ->addCacheableDependency($group)
+        ->addCacheableDependency($account);
     }
 
-    return AccessResult::allowed();
+    return AccessResult::allowed()
```
lramarojaona, line 69: Remove cache.

lib/modules/eic_recommend_content/src/Access/RecommendContentAccessCheck.php
```diff
@@ -120,9 +132,22 @@ public function access(Route $route, AccountInterface $account, string $entity_t
...
+    $is_group_content = FALSE;
 
     switch ($moderation_state) {
       case DefaultContentModerationStates::PUBLISHED_STATE:
+        // If the entity belongs to a group and the group is not published, we
+        // deny access to recommend it.
+        if ($group = $this->eicGroupsHelper->getOwnerGroupByEntity($entity)) {
+          $is_group_content = TRUE;
+          if ($group->get('moderation_state')->value !== $moderation_state) {
+            $access->addCacheableDependency($account)
```
lramarojaona, line 145: Remove cache.

lramarojaona (changes requested):
It works, but I have few remarks.
Feel free to argue if you still think cache is needed there.

lib/modules/eic_groups/src/Access/HighlightGroupContentAccessCheck.php
4hardfire: If I remove the cache and the user becomes a GA, when I refresh the page the highlight button is still not shown. Should we add a max-age = 0?

lramarojaona (approved):
Tested again with your latest commits, it works!

PR #1723, opened 2022-06-27
EICNET-1710: Handle empty address parameters in the sidebar detail event

4hardfire:
### Improvements

- Handle empty address properties from event location address in the sidebar.

### Test

- [ ] Change address of a global event: try to leave the street address as empty and add only the postal code and the City
- [ ] Go to the event detail page and make sure there is no empty space before the postal code + city

Before:
image

- [ ] Edit the event again and leave the postal code and city as empty and add only the street address
- [ ] Go to the event detail page and make sure there is no empty space between the street address and the country name

Before:
image


lramarojaona (approved):
Ok it works alhtough `<br />` is for XHTML, with HTML5 `<br>` is ok

PR #1725, opened 2022-06-28
EICNET-2423: Fix comment notification template

4hardfire:
### Fixes

- Remove `<br>` tag between each tagged user in email notification

### Test

- [x] Go to a discussion detail
- [x] Add a comment and apply some styles (bold, underline, italic and a link) and tag multiple users
- [x] Run cron
- [ ] Make sure all the tagged users get an email and make sure the user image has the correct dimensions + the comment text is shown as html and not as plain text + make sure there is no `<br>` tag between each tagged user.


lramarojaona (approved):
Ok, it works, although it's not the profile pictire being sent but the empty avatar.

This is an improvement that can tackled in another ticket.

PR #1735, opened 2022-06-29
EICNET-2526: Update my invitations view.

lramarojaona:
### Tests

- [x] As TU make sure you are invited to a group
- [x] Go to `/users/<username>/invitations` and check that you have a group type column (named "Type")


4hardfire (approved):
Seems ok to me!

PR #1740, opened 2022-06-29
EICNET-2597: Whenever a user accepts his last invitation, he still sees the warning message that an invitation is awaiting his feedback

4hardfire:
### Improvements

- Decorate group invite event subscriber so that it avoids showing the invitations message on certain routes;
- Small improvements in code for better performance.

### Test

- [x] As SA/SCM, invite a user (with no invitations) to a group
- [x] Login as the user and make sure you have the pending invitations message in the banner
- [x] Go to the invitations page and accept the invitation
- [x] When you get redirected back to the invitations page you should only see the Success message

### DIFF
--- a/lib/modules/eic_groups/src/EventSubscriber/GinviteSubscriber.php
+++ b/lib/modules/eic_groups/src/EventSubscriber/GinviteSubscriber.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace Drupal\eic_groups\EventSubscriber;
+
+use Drupal\Core\Messenger\MessengerInterface;
+use Drupal\Core\Logger\LoggerChannelFactoryInterface;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\Url;
+use Drupal\ginvite\GroupInvitationLoader;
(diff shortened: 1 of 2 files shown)


lib/modules/eic_groups/src/EventSubscriber/GinviteSubscriber.php
```diff
@@ -0,0 +1,85 @@
...
+    ];
+    $route = $event->getRequest()->get('_route');
+
+    // @todo Doing this should already improve some performance however, we
+    // should create a function to query the invitations and limit the results
+    // to 1. This will avoid querying the whole table when we just want to know
+    // if the user has at least 1 pending invitation. Also, it would probably
+    // worth to create a separated block to show the information about pending
+    // invitations, to avoid strange behavior such as opening a new tab and
+    // navigate through the platform before clicking on accept invitation in
```
lramarojaona, line 61: This is a normal behavior of messages in Drupal, I don't think we should do something about that.

lramarojaona (approved):
Ok  it works!

PR #1742, opened 2022-06-30
EICNET-2539: No validation error message shown when required fields in inactive tabs are empty

4hardfire:
### Improvements

- Remove html5 validation from all forms;
- Show required field label on entity tree widget react component after user submits the form.

### Test

- [x] As TU (GM), try to add a discussion in a group
- [x] Leave the Topics and Regions & Countries fields empty
- [x] Click on save
- [x] the field group "Topics, countries & tags" is automatically focused and you see a required label on both Topics and Regions & Countries fields:

image

- [x] Try to create a gallery, story and a video and make sure the field "Contributors" is closed by default
- [x] Try to create a News and make sure the field "Paragraphs" is closed by default

### DIFF
--- a/lib/themes/eic_community/react/components/Field/EntityTree/index.js
+++ b/lib/themes/eic_community/react/components/Field/EntityTree/index.js
@@ -30,18 +30,18 @@ class EntityTree extends React.Component {
     this.addChip = this.addChip.bind(this);
     this.removeChip = this.removeChip.bind(this);
     // Uncomment line bellow if we use html5 form validation.
-    // this.submitForm = this.submitForm.bind(this)
+    this.submitForm = this.submitForm.bind(this)
     this.firstEl = React.createRef();
   }
 
   componentDidMount() {
-    // document.addEventListener('click', this.submitForm)
+    document.addEventListener('click', this.submitForm)
     this.search();
   }
 
   componentWillUnmount() {
     // Uncomment line bellow if we use html5 form validation.
-    // document.removeEventListener('click', this.submitForm)
+    document.removeEventListener('click', this.submitForm)
   }
 
   submitForm(e) {
--- a/lib/themes/eic_community/react/components/Field/EntityTree/index.js
+++ b/lib/themes/eic_community/react/components/Field/EntityTree/index.js
@@ -21,7 +21,7 @@ class EntityTree extends React.Component {
       isLoadingMore: false,
       page: 0,
       length: this.props.length,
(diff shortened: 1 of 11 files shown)


4hardfire:
@lramarojaona found this that might fix the issue -> https://www.drupal.org/project/field_group/issues/2969051

lramarojaona:
@4hardfire I think we should as well collapse paragraph fields that are not required, otherwise it looks like some fields are required even if they are not (both in groups and organisations, maybe check events as well).
This ticket looks like a good candidate to tackle that, wdyt?

4hardfire:
> @4hardfire I think we should as well collapse paragraph fields that are not required, otherwise it looks like some fields are required even if they are not (both in groups and organisations, maybe check events as well). This ticket looks like a good candidate to tackle that, wdyt?

Done!

lib/themes/eic_community/react/components/Field/EntityTree/index.js
```diff
@@ -21,14 +21,15 @@ class EntityTree extends React.Component {
...
-      hasError: false,
+      hasError: this.props.hasError,
     }
 
     this.search = this.search.bind(this);
     this.searchTerms = this.searchTerms.bind(this);
     this.searchUsers = this.searchUsers.bind(this);
     this.addChip = this.addChip.bind(this);
     this.removeChip = this.removeChip.bind(this);
+    // Uncomment line bellow if we use html5 form validation.
```
lramarojaona, line 32: @4hardfire ,
You re-enabled HTML5 validation right?
Do we need this? (Or do we need this comment?)

lib/themes/eic_community/react/components/Field/EntityTree/index.js
```diff
@@ -39,6 +40,7 @@ class EntityTree extends React.Component {
   }
 
   componentWillUnmount() {
+    // Uncomment line bellow if we use html5 form validation.
```
lramarojaona, line 43: @4hardfire ,
You re-enabled HTML5 validation right?
Do we need this? (Or do we need this comment?)

lramarojaona (changes requested):

lib/themes/eic_community/react/components/Field/EntityTree/index.js
4hardfire: HTML5 validation is on. I removed the comment.

lramarojaona (approved):

PR #1743, opened 2022-06-30
EICNET-2598: Groups without an image in /groups should show the default empty image

4hardfire:
### Fixes

Fix group and organisation teasers without an image in react component.

### Test

- [x] Go to `/groups` and make sure groups without an image show the default empty image. Make sure the same happens in `/organisations`


lramarojaona (commented):
Ok it works!

lramarojaona (approved):

PR #1750, opened 2022-07-04
[EICNET-2611]fix: Fix cypress issues with new comment from Quill

Dromadoros:
How to test:

- [x] Take DB from QA and run cypress test


4hardfire (approved):
All tests passed!

PR #1754, opened 2022-07-04
EICNET-1815: Critical actions admin view (Fix request deletion)

4hardfire:
### Fixes/Improvements

- Fix error when accept delete request `"InvalidArgumentException: Field comment_body is unknown"`;
- Remove related group content when node is deleted via request.

### Test

- [x] As TU, go to a discussion and request deletion
- [x] As SA/SCM, accept request
- [x] Make sure all group content related to the discussion have been removed
- [x] Make sure you can see a log message in `/admin/community/logs`

### DIFF
--- a/lib/modules/eic_flags/src/Service/DeleteRequestHandler.php
+++ b/lib/modules/eic_flags/src/Service/DeleteRequestHandler.php
@@ -11,7 +11,6 @@ use Drupal\eic_flags\RequestTypes;
 use Drupal\flag\FlaggingInterface;
 use Drupal\group\Entity\GroupContentInterface;
 use Drupal\group\Entity\GroupInterface;
-use Drupal\node\NodeInterface;
 use Drupal\user\UserInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


lib/modules/eic_flags/src/Service/DeleteRequestHandler.php
```diff
@@ -53,31 +54,33 @@ public function accept(
         break;
 
       case 'node':
+        $this->deleteNodeRelationships($content_entity);
```
lramarojaona, line 57: I don't think we need a batch for this, as we won't have much `group_content` entities per node.
In fact the code has been borken here: https://github.com/EIC-EA/EIC-community-D8/commit/ae1ffc9563d1e6785fe4d573215e12ac13445b12

I just tested with 
```php
      case 'node':
        $content_entity->delete();
        break;
```
and it works perfectly well. Group module will delete all related `group_content` entities in a hook.

lramarojaona (changes requested):

lib/modules/eic_flags/src/Service/DeleteRequestHandler.php
4hardfire: I reverted the code and added `$content_entity->delete();` as discussed.

lramarojaona (approved):
Ok!

PR #1757, opened 2022-07-04
EICNET-2160: Get rid of submit for approval and ask changes transitions in Organisations

lramarojaona:
### Tests

- [x] As SA/SCM/DA, create a user and add it as GA of an organisation
- [x] As GA of the organisation, create a News Article, don't submit
- [x] As GA, make sure you don't have "Submit for approval" button


4hardfire (approved):
It works! I also applied the same behavior for GMs in this PR #1747

PR #1760, opened 2022-07-04
EICNET-2332: Add alt text to field video.

4hardfire:
### Improvements

- Add alt text to field video in News, Story and Video.

### Test

- [x] As TU, try to create a News, Story and a Video
- [x] Make sure the video field has an alt text according to the ticket https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2332

### DIFF
--- a/config/sync/field.field.node.news.field_video.yml
+++ b/config/sync/field.field.node.news.field_video.yml
@@ -12,7 +12,7 @@ field_name: field_video
 entity_type: node
 bundle: news
 label: Video
-description: ''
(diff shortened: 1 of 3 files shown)


lramarojaona (approved):
Maybe a different approach would work as well in a `hook_form_alter`:

- Spot the field names we want to tackle
- Add the description dynamically

In this way whenever we would add the same video field somewhere, it would get the description automatically

But I guess it's ok for now, it works!

PR #1761, opened 2022-07-05
[EICNET-2596]fix: Use group bundle when checking for content in groups helper

korhankale:
## To test

- [x] Create a global event as draft
- [x] The "Get the conversation started" block should be displayed on the group home page
- [x] Create a discussion
- [x] The block shouldn't be displayed


4hardfire (changes requested):

4hardfire (approved):
It works!

PR #1763, opened 2022-07-05
EICNET-1438: Fix issue that breaks activity stream list after loading more results

4hardfire:
### Fixes

- Fix react activity stream result component when activity stream result type is undefined.

### Test

- [x] Import DB from DEV environment
- [x] Re-index solr
- [x] Go to `/groups/testing-playground/latest-activity` and click on load more until you load all the items
- [x] Make sure the list doesn't break


Dromadoros (approved):

Dromadoros:
Nice catch

PR #1765, opened 2022-07-05
EICNET-2602: Add missing permission for GO/GA to administer membership requests.

lramarojaona:
### Tests

- [x] Create a new group and set joining method to "Request"
- [x] As GO/GA (not SA/SCM/DA), make sure you have access to "Membership request" in the group dropdown
- [x] Create a new event
- [x] As GO/GA (not SA/SCM/DA), make sure you have access to "Membership request" in the group dropdown

### DIFF
--- a/config/sync/group.role.event-admin.yml
+++ b/config/sync/group.role.event-admin.yml
@@ -15,6 +15,7 @@ permissions:
   - 'access group_node overview'
   - 'access members management'
   - 'administer members'
+  - 'administer membership requests'
   - 'create group_node:discussion entity'
   - 'create group_node:document entity'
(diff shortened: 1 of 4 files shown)


4hardfire (approved):
It works. I just updated the firt step in the test scenario because the Membership requests link should only be available when the joining method is set to "Request".

lramarojaona:
> It works. I just updated the firt step in the test scenario because the Membership requests link should only be available when the joining method is set to "Request". Just so you know... the permissions are also handled by the joining method plugin -> `TuGroupMembershipRequest.php`

Ok, thanks!

PR #1767, opened 2022-07-05
Feature/eicnet 2595  archived content visibility

lramarojaona:
### Tests

- [x] As GM, create an archive request on a Discussion
- [x] As SA/SCM, Accept the archive request
- [x] As GM, make sure you can still access the Discussion
- [x] As GM, make sure you can't create new request archivals

### DIFF
--- a/config/sync/workflows.workflow.default.yml
+++ b/config/sync/workflows.workflow.default.yml
@@ -21,7 +21,7 @@ type_settings:
     archived:
       label: Archived
       weight: 2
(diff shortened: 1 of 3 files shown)


4hardfire (approved):
In the future we should add a class to save the constants for the news and stories moderation states.
It works!

lramarojaona:
> In the future we should add a class to save the constants for the news and stories moderation states. It works!

Agree, I added a todo

PR #1768, opened 2022-07-05
EICNET-2204: Group teaser archived status

4hardfire:
### Improvements

- Add moderation state in events overview;
- Send group moderation state to the theme variables so that it can be rendered as a tag.

### Todo

- [x] Re-factor group teaser twig template so that it renders the moderation state if not empty. The variable name is `group_state`.

### Test

**As Anonymous and Trusted user:**

- [x] Make sure there are public archived events in the platform
- [x] Go to /events
- [x] Make sure "Archived" label is shown for archived events
- [x] Go to the homepage and make sure you have at least 1 public archived group listed there
- [x] Make sure the "Archived" label is shown

### DIFF
--- a/lib/themes/eic_community/includes/preprocess/groups/event.inc
+++ b/lib/themes/eic_community/includes/preprocess/groups/event.inc
@@ -75,7 +75,7 @@ function eic_community_preprocess_group__event(&$variables) {
         }
 
         // Get group state.
-        $item['group_state'] = _eic_groups_get_moderation_state_tag($group);
+        $teaser['group_state'] = _eic_groups_get_moderation_state_tag($group);
 
         $teaser['type'] = [$teaser['type']];
       }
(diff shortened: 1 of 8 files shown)


lramarojaona (changes requested):
@4hardfire ,
In both cases (anonymous or TU), I don't see archived events in the homepage.

4hardfire:
> @4hardfire , In both cases (anonymous or TU), I don't see archived events in the homepage.

Done!

lramarojaona:
@4hardfire ,
As anonymous I still see the Archived label on the Homepage.

4hardfire:
> Anonymous still sees the Archived label on the homepage

@lramarojaona, I forgot to update the test scenarios. In fact, only the group visibility is shown for authenticated users because anonymous can only see public groups but the archived label should be shown for both.

lramarojaona (approved):
Ok, it works

PR #1795, opened 2022-07-13
EICNET-2642: Anonymous users see broken avatar images

4hardfire:
### Fixes

- Show empty avatar image when viewing user avatars in groups overview page and latest activity stream section.

### Test

- [ ] As anonymous, go to `/groups`
- [ ] Make sure the empty user avatar image is shown for each group teaser item
- [ ] Go to activity stream section of a group
- [ ] Make sure the empty user avatar image is shown for each activity stream item


lramarojaona (approved):
It works

PR #1804, opened 2022-07-13
EICNET-901: Show user display name in message notification about content recommendation

4hardfire:
### Improvements

- Show user display name in message notification about content recommendation.

### Test

- [x] As SA/SCM, try to recommend a group to internal and external users
- [x] Run cron
- [x] Make sure the users get a notification and the text "Dear <name>" (for platform users) or "Dear" (for external users) is presented in the notification email

### DIFF
--- a/config/sync/core.entity_form_display.message.notify_content_recommendation.default.yml
+++ b/config/sync/core.entity_form_display.message.notify_content_recommendation.default.yml
@@ -3,6 +3,7 @@ langcode: en
 status: true
 dependencies:
   config:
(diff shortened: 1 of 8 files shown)


lramarojaona (changes requested):
@4hardfire , I'm getting a PHP error:

_InvalidArgumentException: Field field_receiver_email is unknown. in Drupal\Core\Entity\ContentEntityBase->getTranslatedField() (line 587 of /app/web/core/lib/Drupal/Core/Entity/ContentEntityBase.php)._

lramarojaona (approved):
It works!

PR #1807, opened 2022-07-14
EICNET-2622: I cannot manage members as SCM/SA in a group

4hardfire:
### Improvements

- Add missing permission "Administer group members" to SA/SCM users in global events.

### Test

- [x] As SA/SCM (not DA), go to any global event page
- [x] Go to the "Manage members" page of the global event (from the Manage dropdown in the group header)
- [x] Make sure you can see the options to view/edit/delete membership of each group member

### Notes

- For old global events we need to rebuild permissions

### DIFF
--- a/config/sync/group.role.event-a416e6833.yml
+++ b/config/sync/group.role.event-a416e6833.yml
@@ -24,6 +24,7 @@ permissions:
   - 'access members overview'
   - 'access news overview'
   - 'access wiki overview'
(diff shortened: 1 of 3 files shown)


lramarojaona (approved):
It works

(I did the same here: https://github.com/EIC-EA/EIC-community-D8/pull/1798)

4hardfire:
This has been fixed in #1798

PR #1808, opened 2022-07-14
EICNET-2616: Already Member shown on the Members Pending page

4hardfire:
### Improvements

- Create custom view for the group pending membership requests page and disable the default one provided by groups module.

### Test

- [ ] As SA/SCM, create a public group and set joining method to "Request". Also publish the group
- [ ] As TU, request membership
- [ ] As SA/SCM, go to the pending membership requests page -> `/groups/group-name/members-pending` and make sure you can see the pending request from TU
- [ ] Edit the group and set joining method to "Open"
- [ ] As the previous TU, join the group
- [ ] As SA/SCM, go to the pending membership requests page again and make sure you don't see any membership requests from the previous TU because the user already joined the group.

### DIFF
--- a/config/sync/views.view.eic_group_pending_members.yml
+++ b/config/sync/views.view.eic_group_pending_members.yml
@@ -1,6 +1,6 @@
-uuid: 4d5e76e2-db40-468f-999f-76525f763f5b
+uuid: a1c63060-ca1a-49ad-8a69-da17c38d48b1
 langcode: en
-status: false
+status: true
 dependencies:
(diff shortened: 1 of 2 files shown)


lramarojaona:
@4hardfire , I can't see the pending membership requets in the view (step 3), although I can see it in the "All entities" list.

lramarojaona (approved):
Ok, it works!

I guess as an improvement, we should delete the memebrship request, but that's for another ticket

PR #1812, opened 2022-07-15
EICNET-1144: Show latest members on the group activity stream page (Hide for anonymous)

4hardfire:
### Improvements

- Hide latest members block from activity stream when current user is anonymous.

### Test

- [x] As anonymous, go to the activity stream section of a public group
- [x] Make sure you can't see the "Latest members" block
- [x] Login as any user, and make sure you can see the "Latest members" block


lramarojaona:
@4hardfire ,
I can't see the Latest members block.

lramarojaona (approved):
Ok, it works!

PR #1815, opened 2022-07-15
v1.0.0

lramarojaona:


lramarojaona:
We should merge from release branch and not develop, please use this PR instead: https://github.com/EIC-EA/EIC-community-D8/pull/1817

PR #1816, opened 2022-07-15
EICNET-1091: As a GM I can see when content has been shared inside my group

4hardfire:
### Fixes

- Fix wrong author name of the user that triggered the shared the content with another group.

### Test

- [x] As TU (member), try to share a discussion with another group. Make sure you are not the author of the discussion.
- [x] Go to the activity stream of the group where the discussion was shared
- [x] Make sure the author name of the message is the user that shared the content and not the author of the entity


lramarojaona (approved):
It works!

PR #1825, opened 2022-07-19
EICNET-2520: Allow users to re-request membership after being rejected.

lramarojaona:
### Tests

- [ ] As TU, go to a group and request membership
- [ ] As GO/GA/SA/SCM, go to the group and reject the membership
- [ ] As TU, go to the group and request membership again
- [ ] Make sure you membership request is taken into account

### DIFF
--- a/lib/modules/eic_groups/eic_groups.module
+++ b/lib/modules/eic_groups/eic_groups.module
@@ -358,6 +358,13 @@ function eic_groups_module_implements_alter(&$implementations, $hook) {
  * Implements hook_form_alter().
  */
 function eic_groups_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  // Get group types.
+  $group_types = [];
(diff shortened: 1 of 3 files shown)


4hardfire (commented):
It works quite well. However, I found an issue when accepting the membership request. Apparently I can set the owner role for the user and end up with 2 owners for the same group. In my opinion we should fix this.

4hardfire (changes requested):
I think there is an issue with the custom view that I made (http://eic.europa.eu/admin/structure/views/view/eic_group_pending_members)... I can't see the membership request for 2 of my groups. I believe the last filter in the filter criteria is breaking it. Could you have a look? That view was made in order to avoid showing membership requests of users that already are members of the group.

lramarojaona:
@4hardfire , I removed the check you mention, but the view is filetered on "pending" requests already so it should be ok

4hardfire (approved):
It's ok for now!

PR #1828, opened 2022-07-25
EICNET-2660: migrated bookpage challenges generates error with non admin user

4hardfire:
### Fixes

- Fix error when trying to access book pages as TU.

### Test

- [x] As SA/SCM, create a new book page and publish it
- [x] As TU, go to the book page and make sure there is no error 500 and you can see the detail page normally

### DIFF
--- a/lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
+++ b/lib/modules/eic_content/modules/eic_content_book/src/Hooks/EntityOperations.php
@@ -88,7 +88,11 @@ class EntityOperations implements ContainerInjectionInterface {
           $add_book_page_urls = $this->getBookPageAddFormUrls($entity);
 
           // If user can't access the route.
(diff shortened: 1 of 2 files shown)


lramarojaona (changes requested):
@4hardfire , it works but I'm getting a PHP notice:

_Notice: Trying to access array offset on value of type null in Drupal\eic_content_book\Plugin\Block\EICBookNavigationBlock->build() (line 106 of /app/lib/modules/eic_content/modules/eic_content_book/src/Plugin/Block/EICBookNavigationBlock.php)._

lramarojaona:
> @4hardfire , it works but I'm getting a PHP notice:
> 
> _Notice: Trying to access array offset on value of type null in Drupal\eic_content_book\Plugin\Block\EICBookNavigationBlock->build() (line 106 of /app/lib/modules/eic_content/modules/eic_content_book/src/Plugin/Block/EICBookNavigationBlock.php)._

Ok, I fixed it.

lramarojaona (approved):

PR #1846, opened 2022-07-27
EICNET-2670: As a user, I want a limit on group invitation reminders so that I am not spammed

4hardfire:
### Improvements

- Save number of times an invitation has been sent and the date of the last reminder;
- Improve the group invitation reminder cron job in order to skip sending reminders of old invitations;
- limit number of reminders to 3 and send every 3 days.
- New configurations settings: `cron_interval_group_skip_invite_reminder_time` and `cron_interval_group_invite_reminder_frequency_time`

### Test

For BE dev review:

- [x] Comment lines 296, 297, 298 of **modules/eic_groups/src/Hooks/CronOperations.php** file
- [x] Make sure you have old pending invitations in a group
- [x] Run cron
- [x] Make sure the users receive a notification about the reminder
- [x] Do a DB query to change the value in the table **group_content__field_invitation_reminder_count**: column **field_invitation_reminder_count_value** = 3
- [x] Clear the cache and run cron again
- [x] Make sure the reminder for the invitations with the field_invitation_reminder_count_value = 3 is not sent
- [x] Do a DB query to change the value in the table **group_content__field_invitation_reminder_date**: Column **field_invitation_reminder_date_value** = <unix-timestamp-of-3-days-ago>
- [x] Clear the cache and run cron again
- [x] Make sure the reminder for the invitations with the field_invitation_reminder_date_value = 3-days-ago is not sent

For QA:

- To test this you should follow the acceptance criteria in the ticket https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2670.


config/sync/field.field.group_content.event-group_invitation.field_invitation_reminder_count.yml
```diff
@@ -0,0 +1,25 @@
+uuid: a6ecc22c-4cda-45ac-b5d6-97da9112fcd4
```
lramarojaona, line 1: Would be way better to create those fields through `hook_entity_bundle_field_info()` so that the site won't break if they add a new Group type (they might add other group types in the future like Challenges and Projects).

lib/modules/eic_groups/src/Hooks/CronOperations.php
```diff
@@ -295,11 +297,37 @@ private function processGroupInvitationsReminder() {
       return;
     }
 
+    // Date from which we want to skip invitation reminders to avoind sending
+    // reminders for old invitations.
+    $skip_invitation_reminder_time = new DrupalDateTime('today -' . Settings::get('cron_interval_group_skip_invite_reminder_time', 86400 * 30) . ' days');
```
lramarojaona, line 302: You should create a Config for that and not a setting.
See: https://drupal.stackexchange.com/questions/227983/how-to-know-when-to-use-config-or-settings-overrides-in-settings-php

Also your calculation is wrong here:
`86400 * 30) . ' days'`
You don't need 86400 cause you are using `days`

lib/modules/eic_groups/src/Hooks/CronOperations.php
```diff
@@ -295,11 +297,37 @@ private function processGroupInvitationsReminder() {
...
+
+    // Frequency date to send reminders. By default every 3 days a reminder
+    // will be sent for the same invitation.
+    $reminder_frequency_time = new DrupalDateTime('today -' . Settings::get('cron_interval_group_invite_reminder_frequency_time', 86400 * 3) . ' days');
```
lramarojaona, line 306: You should create a Config for that and not a setting.
See: https://drupal.stackexchange.com/questions/227983/how-to-know-when-to-use-config-or-settings-overrides-in-settings-php

Also your calculation is wrong here:
`86400 * 30) . ' days'`
You don't need 86400 cause you are using `days`

lramarojaona (changes requested):

lib/modules/eic_groups/src/Hooks/CronOperations.php
4hardfire: Done!

lib/modules/eic_groups/src/Hooks/CronOperations.php
4hardfire: Done!

config/sync/field.field.group_content.event-group_invitation.field_invitation_reminder_count.yml
4hardfire: This is done. When installing group_invitation content in a group, the fields configurations are automatically added.

lramarojaona (approved):
Ok, it works!

PR #1853, opened 2022-07-28
Feature/eicnet 2683 matomo

lramarojaona:
### Tests

- [x] Check that `matomo` module is enabled: `/admin/config/system/matomo`
- [x] Check that `oe_webtools_analytics` module is disabled: `/admin/config/system/oe_webtools_analytics`

We can't really test more than this on our side since we don't have acces to it.


web/sites/default/default.settings.local.php
```diff
@@ -210,6 +210,13 @@
 $config['smtp.settings']['smtp_from'] = '';
 $config['smtp.settings']['smtp_fromname'] = '';
 
+/**
+ * Matomo settings.
+ */
+$config['matomo.settings']['site_id'] = getenv('MATOMO_SITE_ID');
+$config['matomo.settings']['url_http'] = 'http://' . getenv('MATOMO_URL');
+$config['matomo.settings']['url_https'] = 'https://' . getenv('MATOMO_URL');
```
Dromadoros, line 218: We should add these env in .env.dist

Dromadoros (approved):

web/sites/default/default.settings.local.php
lramarojaona: Ok, I guess you `.env.docker`, just fixed that.

PR #1860, opened 2022-07-28
EICNET-2679: Respect field format.

lramarojaona:
### Tests

- [x] Go to `/admin/content/block-content`
- [x] Edit the "Homepage - Stay up to date block" block
- [x] Make sure you have some HTML tags in the paragraphs and save
- [x] Go to the homepage and check that thos HTML tags are interpreted correctly and not raw outputed


4hardfire (approved):
It works!

PR #1877, opened 2022-08-01
EICNET-2698: Wiki tab is missing on production env

4hardfire:
### Fixes

- Fix issue where group wiki section was not published after migrating groups.

### Test

- [ ] Update group/event migrations -> `drush mim upgrade_d7_node_complete_group --update` and `drush mim upgrade_d7_node_complete_event_site --update`
- [ ] Make sure the wiki section of migrated groups and events is published

### Regression tests:

- [ ] As SA/SCM, create a new public group as draft
- [ ] Edit the group and make sure the wiki feature is enable and publish the group
- [ ] Make sure as anonymous you can see the wiki tab


lramarojaona (approved):
Ok, it works

PR #1895, opened 2022-08-11
EICNET-2734: [User profile] A trusted user should be able to edit his role.

4hardfire:
### Fixes

- Disable job title field for SMED users.

### Test

- [x] As TU (with SMED ID), try to edit your profile information
- [x] Make sure the Job title field is readonly
- [x] As SA/SCM, edit the profile information of the previous TU
- [x] Make sure you can edit the Job title field
- [x] As TU (without SMED ID), try to edit your profile information
- [x] Make sure you can edit the Job title field. This should only happen for users without SMED ID.

### DIFF
--- a/lib/modules/eic_user/src/Hooks/FormAlter.php
+++ b/lib/modules/eic_user/src/Hooks/FormAlter.php
@@ -218,14 +218,11 @@ class FormAlter implements ContainerInjectionInterface {
     /** @var \Drupal\profile\Entity\Profile $profile */
     $profile = $form_state->getFormObject()->getEntity();
     $user = $profile->getOwner();
-    $is_admin = UserHelper::isPowerUser($this->currentUser);
+    $current_user_is_admin = UserHelper::isPowerUser($this->currentUser);
+    $is_smed_user = UserHelper::isSmedUser($user);
 
     // If the user has a SMED ID, we make the job title field as readonly.
-    if (
-      !$is_admin &&
-      $user->hasField('field_smed_id') &&
-      !$user->get('field_smed_id')->isEmpty()
-    ) {
+    if (!$current_user_is_admin && $is_smed_user) {
       $form['field_vocab_job_title']['#disabled'] = TRUE;
     }
   }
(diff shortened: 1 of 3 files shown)


lib/modules/eic_user/src/Hooks/FormAlter.php
```diff
@@ -188,6 +204,34 @@ public function submitBulkInviteUsers(array $form, FormStateInterface $form_stat
...
+   *
+   * @param array $form
+   *   An associative array containing the structure of the form.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
+   * @param string $form_id
+   *   The form ID.
+   */
+  public function alterProfileMemberForm(array &$form, FormStateInterface $form_state, string $form_id) {
+    $user = \Drupal::routeMatch()->getParameter('user');
```
lramarojaona, line 218: It would be better to get the user from the form directly, and I guess we don't need to test the entity as this is the user form:
```
  /** @var \Drupal\user\UserInterface $type */
  $user = $form_state->getFormObject()->getEntity();
```

lramarojaona (changes requested):

lib/modules/eic_user/src/Hooks/FormAlter.php
4hardfire: Done!

lib/modules/eic_user/src/Hooks/FormAlter.php
```diff
@@ -188,6 +204,32 @@ public function submitBulkInviteUsers(array $form, FormStateInterface $form_stat
...
+   *   The form ID.
+   */
+  public function alterProfileMemberForm(array &$form, FormStateInterface $form_state, string $form_id) {
+    /** @var \Drupal\profile\Entity\Profile $profile */
+    $profile = $form_state->getFormObject()->getEntity();
+    $user = $profile->getOwner();
+    $is_admin = UserHelper::isPowerUser($this->currentUser);
+
+    // If the user has a SMED ID, we make the job title field as readonly.
+    if (
```
lramarojaona, line 224: We could put this logic in UserHelper, such as
```
UserHelper::isSmedUser(Account $user): bool{}
```
So we can re-use this, as I believe there will be more stuff like this in the future

lramarojaona (changes requested):

lib/modules/eic_user/src/Hooks/FormAlter.php
4hardfire: I agree with you. It's done!

lramarojaona (approved):

PR #1896, opened 2022-08-11
EICNET-2720: Switch 'Recently updated' from timestamp to changed time.

lramarojaona:
### Tests

- [x] Go to `/admin/community/overview-pages/3/edit` and `/admin/community/overview-pages/5/edit` and enable "Recently updated"
- [x] Go to `/groups` and sort by Recently updated
- [x] Now go to an old group and update it
- [x] Go back to `/groups` and sort by Recently updated
- [x] Check that the updated group shows first
- [x] Repeat the same for `/events`

### Notes

- Since the overview configuration is content, it has to be manually updated in PROD


4hardfire (approved):
It works!

PR #1898, opened 2022-08-11
EICNET-2741: Fix group_permissions module when group is not saved yet.

lramarojaona:
### Tests

- [x] As TU (not DA), go to `/events` and click "Add event"
- [x] Make sure you can fill in the form and save without any PHP error


4hardfire (changes requested):
I still have an error when creating an Event as TU:
`Drupal\Core\Entity\EntityStorageException: Cannot load GroupContent entities for an unsaved group. in Drupal\group\Entity\Storage\GroupContentStorage->loadByGroup() (line 70 of /var/www/html/web/modules/contrib/group/src/Entity/Storage/GroupContentStorage.php).`

4hardfire (approved):
It works!

PR #1899, opened 2022-08-11
EICNET-2766: [Organisation] Trusted should not join organisation without approaval of the admnistrator/owner

4hardfire:
### Fixes

- Remove group permission "Join group" from outsiders and trusted users in group type organisation;
- Create hook_update to remove group permission from existing organisations.

### Test

- [x] As TU (non-member), make sure you cannot see the Join button when viewing an organisation. The memberships should be managed via SMED webservices.


lramarojaona (approved):
It works

PR #1900, opened 2022-08-11
EICNET-2754: Moderated workflow: allow Publish transistion from unpublished.

lramarojaona:
### Tests

- [x] As GM (no SA/SCM/DA), go to an Event
- [x] Create a Story and save as Unpublished
- [x] Edit the Story again ane make sure you can save it as Published


4hardfire (approved):
It works!

PR #1903, opened 2022-08-12
EICNET-2778: [Event] News and stories created within an event should only be visible to the member of that event

4hardfire:
### Fixes

- Add extra pre filter in NewsStorySourceType used in global overview of News and Stories in order to discard News and Stories that belong to groups.

### Test

- [x] As SA/SCM, create a News inside a global event
- [x] Go to the global overview page of News and Stories -> `/articles`
- [x] Make sure only global news are shown in the results. News that belong to groups should not be shown.


lramarojaona (approved):
It works!

PR #1912, opened 2022-08-17
[EICNET-2781]feat: Do not allow group content (from event) to be shown for non-members (only SOLR)

Dromadoros:
How to test:

- [x] Reindex solr
- [x] As anonymous go to a public event that have content like discussions
- [x] Verify you can see the detail page
- [x] Go to discussions tab now and verify that you cannot see content
- [x] Now join the group and verify that you can see different discussions

### DIFF
--- a/lib/modules/eic_search/src/Search/DocumentProcessor/ProcessorMessage.php
+++ b/lib/modules/eic_search/src/Search/DocumentProcessor/ProcessorMessage.php
@@ -212,17 +212,11 @@ class ProcessorMessage extends DocumentProcessor {
 
       if ($group = Group::load($node_group_id)) {
         $group_follows = $this->getFollowUidByFlag('follow_group', $group);
-        $group_parent_label = $group->label();
-        $group_parent_url = $group->toUrl()->toString();
         $group_parent_id = $group->id();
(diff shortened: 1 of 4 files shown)


lib/modules/eic_search/src/Search/DocumentProcessor/ProcessorMessage.php
```diff
@@ -212,6 +212,17 @@ public function process(Document &$document, array $fields, array $items = []):
 
       if ($group = Group::load($node_group_id)) {
         $group_follows = $this->getFollowUidByFlag('follow_group', $group);
+        $group_parent_label = $group->label();
+        $group_parent_url = $group->toUrl()->toString();
+        $group_parent_id = $group->id();
+        $group_is_published = $group->isPublished();
+        $group_type = $group->getGroupType()->id();
```
lramarojaona, line 219: Is there a reason to include `label`, `url`, and `is_published`?
If we don't use them, it would be better to drop that, as this is subject to change and then re-index is needed.

lramarojaona (commented):
It works, but I have a question.

lramarojaona (approved):
Ok, it works perfectly!

PR #1913, opened 2022-08-17
[EICNET-2768]feat: Hide related groups in news

Dromadoros:
How to test:

- [x] Edit/create a news and verify that related groups is no more in the form display


lramarojaona (approved):
Perfect!

PR #1914, opened 2022-08-17
[EICNET-2787]feat: Disable bookmark content

Dromadoros:
How to test:

- [x] Go to different node (not group_content) and verify bookmark action is not there anymore


lramarojaona (approved):
It works

PR #1934, opened 2022-08-26
[EICNET-2802]feat: Implement also profile update into the user one

Dromadoros:
### Improvements

- Created a re-usable service to handle social_link fields.

### Tests

- [x] With Postman, trigger the `User - CREATE` request
- [x] Check that you have a new user created with populated profile information
- [x] Now trigger the `User - UPDATE` request
- [x] Check that the user was updated and profile as well (you should see _updated_ text in `About you` section of the profile)


4hardfire (approved):
It works!

4hardfire (changes requested):
The profile resource configurations need to be committed.

PR #1937, opened 2022-08-31
Feature/eicnet 2817 group user management permissions

lramarojaona:
### Improvements

- Created a new block `eic_page_contextual_action_block` plugin to display page contextual links.

### Tests

- [x] As SA/SCM, go to a group members page /groups/<group-name>/members
- [x] Make sure you have a "Add member" link and that you can access the page by clicking
- [x] Run `drush updb -y`
- [x] As GO/GA (non SA/SCM), go to an existing group/organisation/event
- [x] Make sure permissions follow the scheme described in the ticket: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2817

To easily test the udpate, you can use the following in eic_groups_update_9010, eic_groups_update_9011 and eic_groups_update_9012:
EICNET-2817


lramarojaona:
> It works!
> 
> Just a small note: when changing the group member role of a user to a group admin, there is a cache issue in the group header. The user cannot see the updated links in the dropdown until the cache is cleared.

I jjust fixed this

PR #1941, opened 2022-09-05
EICNET-2779: [Event] A none logged user should not see pending events

4hardfire:
### Fixes

- Do not show group content of pending groups in global search (including content inside organisations).

### Test

- [x] As SA/SCM, create a public group and leave it as pending
- [x] Create content in the group
- [x] Make sure content is shown in the global search only for GO/GA/SA/SCM
- [x] As anonymous, make sure the content is not shown or any other user besides GO


lramarojaona (approved):
Ok, it works

PR #1943, opened 2022-09-06
Feature/eicnet 2832 user smed check fix

lramarojaona:
### Tests

- [ ] Login through EU Login as a user that is not `USER_VALID` or `USER_APPROVED_COMPLETE`
- [ ] Make sure you receive the custom message (see https://github.com/EIC-EA/EIC-community-D8/blob/develop/lib/modules/eic_user/modules/eic_user_login/src/Service/SmedUserManager.php#L125)
- [ ] Make sure you don't get the `Your account is blocked or has not been activated. Please contact a site administrator.` message


Sigtechnica:
looks good, interesting approach!

PR #1944, opened 2022-09-06
EICNET-2797: Disable visibility sorting.

lramarojaona:
### Tests

- [x] As SA, go to `/admin/community/groups/groups`, `/admin/community/groups/organisations` and `/admin/community/groups/events`
- [x] Make sure you can't sort on the visibility column in the table


4hardfire (approved):
It works!

PR #1945, opened 2022-09-07
EICNET-2817: need to reinitialize the array

Sigtechnica:
to clean up the list of role ids.


Sigtechnica:
as discussed with @lramarojaona , the array was keeping the information of previous groups

PR #1947, opened 2022-09-08
EICNET-2673: Homepage statistics is not updated for organizations

4hardfire:
### Fixes

- Disable access check when querying statistics for facts and figures;
- Revert previous fix that was using search api to retrieve the counters.

### Test

- [x] Run cron
- [x] Go to the homepage and make sure the statistics in Facts and Figures block are updated
- [x] Try to add a new group and publish it
- [x] Run cron
- [x] Make sure the new group has been incremented to Facts and Figures statistics.


lramarojaona (approved):
It works!

PR #1948, opened 2022-09-08
Feature/eicnet 2801 profile about you default format

lramarojaona:
### Improvements

- Added https://www.drupal.org/project/allowed_formats module

### Tests

- [x] Check that when creating a user through UI, you cannot select full_html in the "About you" field


4hardfire (approved):
It works!

PR #1949, opened 2022-09-08
EICNET-2840: Allow power users to edit first and last name.

lramarojaona:
### Tests

- [x] Log in as DA/SA/SCM
- [x] Find a user that is managed by cas module (user should have an entry in the `authmap` table)
- [x] Make sure you can edit the first and last name


4hardfire (approved):
It works!

PR #1951, opened 2022-09-08
EICNET-2798: Filters on a the topics do not point to correct taxonomy

4hardfire:
### Fixes

- Add cache url.path as cache context in topic statistics block.
- Fix issue where News content type was not shown in Stories overview page.

### Test

- [x] As any user, go to /topics
- [x] Select a topic
- [x] In the topic detail page check the statistics block and click in one of them
- [x] Make sure you get redirected to the overview page with the correct topic as a filter
- [x] Go to a different topic detail page
- [x] Make sure the statistics are different and correct according to the selected topic
- [x] Click in one of the statistics and make sure you get redirected to the overview page with the correct topic as a filter

### Test stories overview

- [x] As any user, go to /stories
- [x] Make sure you can see stories and news

### DIFF
--- a/lib/modules/eic_search/src/Search/Sources/NewsStorySourceType.php
+++ b/lib/modules/eic_search/src/Search/Sources/NewsStorySourceType.php
@@ -134,7 +134,7 @@ class NewsStorySourceType extends SourceType {
    * @inheritDoc
    */
   public function getPrefilteredContentType(): array {
-    return ['news', 'story'];
+    return ['story'];
   }
 
   /**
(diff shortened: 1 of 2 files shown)


lib/modules/eic_search/src/Search/Sources/NewsStorySourceType.php
```diff
@@ -134,7 +134,7 @@ public function ableToPrefilteredByGroup(): bool {
    * @inheritDoc
    */
   public function getPrefilteredContentType(): array {
-    return ['story'];
+    return ['news', 'story'];
```
lramarojaona, line 137: @4hardfire ,
Whe should not show News in the Stories overview, this is a fix we made few weeks ago, following the client's requirement.

lramarojaona (changes requested):

lib/modules/eic_search/src/Search/Sources/NewsStorySourceType.php
4hardfire: Reverted back!

Dromadoros (approved):

PR #1953, opened 2022-09-09
Feature/eicnet 2790 missinghp block

lramarojaona:
### Remark

- Wait for feedback before merging: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2790

### Improvements

- Created a pathauto pattern for Page CT

### Tests

- [x] As SA, create 2/3 Page content
- [x] Go to `/admin/content/block-content`, edit the "Homepage - USP Tiles" block and reference the pages you created
- [x] As anonymous go to the homepage
- [x] Make sure there is a block named "As an EIC beneficiary, access a one-stop-shop" with the tiles of your pages

### DIFF
--- a/config/sync/core.entity_form_display.node.page.default.yml
+++ b/config/sync/core.entity_form_display.node.page.default.yml
@@ -14,6 +14,7 @@ dependencies:
     - content_moderation
     - field_group
     - link
+    - media_library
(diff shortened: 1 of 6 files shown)


4hardfire (approved):
it works!

PR #1958, opened 2022-09-13
Issue EICNET-2826: updated response messages when user is blocked, based on the smed status

Sigtechnica:
….


4hardfire (approved):
Looks good to me!

PR #1966, opened 2022-09-20
EICNET-2715: [Stories] Manage content display wrong information for draft (Rework)

4hardfire:
### Fixes

- Fix issue with method isPublished() from ArchiveRequestHandler when checking if user entity is published.

### Test

- [ ] As anonymous, go to the login page -> /user/login
- [ ] Try to login as a non CAS user
- [ ] Make sure there is no error and the homepage loads correctly -> `Error: Call to undefined method Drupal\group\Entity\GroupContent::isPublished() in Drupal\eic_flags\Service\ArchiveRequestHandler->isPublished()`


Sigtechnica:
tested member page of an organisation works again /user/login aswell  (only for accp and test)

PR #1980, opened 2022-09-27
EICNET-2715: [Stories] Manage content display wrong information for draft (Rework)

4hardfire:
### Improvements

- Improve logic to check if an entity is published and can be requested for archival.

### Test

- [x] As TU, make sure you get no error when you navigate to published group detail page
- [x] As TU, make sure you get no error when you navigate to a group members overview page
- [x] As TU, make sure you get no error when you navigate to published group content node


lramarojaona (approved):
Ok, it works!

PR #1981, opened 2022-09-27
Feature/eicnet 2853 group permission revisions

lramarojaona:
### Tests

- [x] Install a dump from PROD
- [x] Get the result of `drush sqlq "SELECT count(*) FROM group_permission_revision;"`
- [x] Run `drush updb -y`
- [x] Get the result of `drush sqlq "SELECT count(*) FROM group_permission_revision;"` and make sure you have the exact same number as the number of existing groups (+/- 7000 items)
- [x] Edit a group and enable/disable some features
- [x] Get the result of `drush sqlq "SELECT count(*) FROM group_permission_revision;"` and make sure you have the exact same number as the number of existing groups (no new revision should be created)

### TODO

- Check if cache must be explictely cleared, or if it's already done upon the save() function.

### DIFF
--- a/composer.patches.json
+++ b/composer.patches.json
@@ -48,7 +48,8 @@
       "3302469 - Allow other modules to alter the calculated permissions": "https://www.drupal.org/files/issues/2022-08-09/group-calculated_permissions_alter-3302469-4.patch"
     },
     "drupal/group_permissions": {
-      "3302469 - Allow other modules to alter the calculated permissions": "https://www.drupal.org/files/issues/2022-08-12/group_permissions-cache_performance_issue-3301865-7.patch"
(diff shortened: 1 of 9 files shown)


4hardfire (changes requested):
The last step failled. After I edit a group and disable/enable some permissions, for some reason I endup with one more revision. The number of groups on my side is 7116 and now I have 7117 group permission revisions.

4hardfire:
@lramarojaona if you edit the following group -> `/groups/coaching/edit` and disable 2 features, I end up with 1 more revision.

lramarojaona:
> @lramarojaona if you edit the following group -> `/groups/coaching/edit` and disable 2 features, I end up with 1 more revision.

Ok, I fixed it, can you re-try?

4hardfire (approved):
Ok it works!

PR #1983, opened 2022-09-27
EICNET-2823: Don't assign roles upon cas user registration.

lramarojaona:
### Tests

- [x] As SA, go to `/admin/config/eic/smed` and make sure **Allow users to register within Drupal** is checked
- [x] As SA, go to `/admin/config/eic/smed` and make sure **Check/sync user** is unchecked
- [x] As anonymous go to the site, click **Member access** and try to login
- [x] As SA, make sure a user has been created and the user doesn't have any role assigned

### DIFF
--- a/config/sync/cas.settings.yml
+++ b/config/sync/cas.settings.yml
@@ -29,8 +29,7 @@ user_accounts:
   email_assignment_strategy: 1
   email_hostname: localhost
   email_attribute: email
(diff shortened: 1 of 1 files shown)


4hardfire (changes requested):
This failed in the last step. The user was indeed created without any role but the user status is set to Active. Can you have a look?
image

PR #1987, opened 2022-09-29
EICNET-2747: As a participant (trusted user) to a SMED Event i should no be able leave an event

4hardfire:
### Improvements

- Alter entity.group.leave route and add custom access check to validate if a group is a SMED group before allowing user to leave group;
- Create hook update to add missing group permission "delete memberships" to GO/GA;
- Add logic to deny access to delete GAs from SMED groups.

### Test

- [x] Create a global event (with SMED ID) via WS
- [x] Add some GMs and GAs to the event
- [x] As GO/GA, go to event members page -> `/events/<event-name>/members`
- [x] Make sure you can only delete GMs
- [x] Make sure you cannot leave the event as GO/GA/GM
- [x] As SA/SCM, edit the eventand remove SMED ID
- [x] As GA/GM, make sure you can leave the event
- [x] As GO/GA, make sure you can remove GMs and GAs -> `/events/<group-name>/members`


lib/modules/eic_groups/modules/eic_group_membership/eic_group_membership.module
```diff
@@ -507,12 +508,25 @@ function eic_group_membership_group_content_access(EntityInterface $entity, $ope
         break;
       }
 
+      // Admins cannot be removed from SMED groups.
+      if (
```
lramarojaona, line 512: This condition is not ok.
If the user is is a power user, the condition does not apply, then a GO/GA can delete a GA if the user is a power user.

lib/modules/eic_groups/modules/eic_group_membership/eic_group_membership.module
```diff
@@ -507,12 +508,25 @@ function eic_group_membership_group_content_access(EntityInterface $entity, $ope
...
+        break;
+      }
+
       $current_member = $group->getMember($account);
 
       // We deny access if the user is not a power user or doesn't have
       // permission to administer group members.
-      $has_delete_permission = ($account->hasPermission('administer members') ||
-        $account->hasPermission('delete memberships'));
+      $has_delete_permission = ($current_member->hasPermission('administer members') ||
```
lramarojaona, line 528: This is wrong, if I'm a power user not part of the group (outsider), I'm getting this error:
`Error: Call to a member function hasPermission() on bool in eic_group_membership_group_content_access() (line 528 of /app/lib/modules/eic_groups/modules/eic_group_membership/eic_group_membership.module). `

In fact we should check the permissions of both member roles and outsider roles.

This small module might be very ufseful: https://www.drupal.org/project/group_outsider_in

See: https://www.drupal.org/project/group/issues/2884662

lib/modules/eic_groups/src/EICGroupsHelper.php
```diff
@@ -1149,4 +1149,14 @@ public static function getGroupDefaultFeatures(Group $group) {
     return $config->get('default_features') ?? [];
   }
 
+  /**
+   * Checks if a group is a SMED group.
+   *
+   * @param \Drupal\group\Entity\GroupInterface $group
+   *   The group entity.
+   */
+  public static function isSmedGroup(GroupInterface $group) {
```
lramarojaona, line 1158: For this we can use `Drupal\eic_webservices\Utility\EicWsHelper::isCreatedThroughSmed()`.

If you think it's more convenient to have it in `EICGroupsHelper`, then just make this method a wrapper method of `Drupal\eic_webservices\Utility\EicWsHelper::isCreatedThroughSmed()`.

lramarojaona (changes requested):

lib/modules/eic_groups/modules/eic_group_membership/eic_group_membership.module
4hardfire: I changed the condition but we still assume that SA/SCM users can delete everyone (except the group owner).

lib/modules/eic_groups/modules/eic_group_membership/eic_group_membership.module
4hardfire: I refactored the code and installed the contrib module. Can you validate @lramarojaona?

lib/modules/eic_groups/src/EICGroupsHelper.php
4hardfire: Done!

lramarojaona (approved):
Ok, it works!

PR #1991, opened 2022-09-30
EICNET-2875: mails from contact form are stuck in the queue

4hardfire:
### Fixes

- Add missing condition to check if field_sender references anonymous user when preprocessing global contact messages.

### Test

- [ ] Import DB from PROD
- [ ] Run drush cim and drush cr
- [ ] Run `drush queue:run eic_message_notify_queue`
- [ ] Make sure the stucked messages are sent


Sigtechnica:
looks ok !

PR #1994, opened 2022-09-30
EICNET-2697: Do not link users to their profile for anonymous users.

lramarojaona:
### Tests

- [x] As TU, go to a group's About page
- [x] Make sure Group owner and Administrators have a link to their profile
- [x] As Anonymous go to a group's About page
- [x] Make sure Group owner and Administrators don't have a link to their profile


4hardfire (approved):
It works!

PR #1995, opened 2022-09-30
EICNET-2658: Hide the From textfield from the user contact form.

lramarojaona:
### Tests

- [ ] As TU, go to a user profile
- [ ] Click on _Direct message_ button
- [ ] On the user contact form, make sure there is no _From_ textfield


4hardfire (approved):
It works!

PR #1998, opened 2022-10-03
Feature/eicnet 2807 invitee email access

lramarojaona:
### Tests

- [x]  As GO/GA, go to your group
- [x] Invite an existing platform user and an external user
- [x] Make sure you don't see the platform user email in the invitations list
- [x] Make sure you do see the external user email in the invitations list
- [x] As SA/SCM, go to the same group
- [x] Make sure you see email address for both users


4hardfire (approved):
It works!

PR #2002, opened 2022-10-05
Feature/eicnet 2722 flag access check

lramarojaona:
### Improvements

- Moved our custom logic for `isPublished()` and `isArchived()` to a new `Drupal\eic_moderation\ModerationHelper` service
- Refactored the code in theme for testing the access to share

### Tests

- [x] As SA, create a Public group and publish it, make sure you can like and share on social media
- [x] Create a Document, Discussion, Video, Gallery and Event as Draft (unpublished)
- [x] Make you sure you cannot like and share on all of them
- [x] Publish them and make sure you can like and share
- [x] Unpublish the group
- [x] Make sure you cannot like and share the Group, Document, Discussion, Video, Gallery and Event


4hardfire (changes requested):
Code seems fine but some steps didn't pass:

- `As SA, create a Public group and publish it, make sure you can like and share on social media`
When the group is pending I still see the share link. Also, for some reason the like button is never shown even if the group is published

- `Make you sure you cannot like and share on all of them`
After creating a group content, the share button is always shown (even in draft). The like button is ok.

- `Make sure you cannot like and share the Group, Document, Discussion, Video, Gallery and Event`
The like button is never shown in the group header (even when group is published) and the share button is always shown independently of the group status. For the group content, the like button is ok but the share button is always shown independently of the node moderation state or group moderation state.

lib/modules/eic_moderation/src/ModerationHelper.php
```diff
@@ -0,0 +1,128 @@
+<?php
```
4hardfire, line 1: I would put this file inside eic_moderation/src/Services

4hardfire (changes requested):

lib/modules/eic_moderation/src/ModerationHelper.php
lramarojaona: Well, we already have lots of Helper classes that are the src root

lramarojaona:
> Code seems fine but some steps didn't pass:
> 
>     * `As SA, create a Public group and publish it, make sure you can like and share on social media`
>       When the group is pending I still see the share link. Also, for some reason the like button is never shown even if the group is published
> 
>     * `Make you sure you cannot like and share on all of them`
>       After creating a group content, the share button is always shown (even in draft). The like button is ok.
> 
>     * `Make sure you cannot like and share the Group, Document, Discussion, Video, Gallery and Event`
>       The like button is never shown in the group header (even when group is published) and the share button is always shown independently of the group status. For the group content, the like button is ok but the share button is always shown independently of the node moderation state or group moderation state.

@4hardfire , there was some code messing up tieh flags, so the like was always removed.

I fixed the otcher checks.

Can you re-test?

lib/themes/eic_community/includes/preprocess/blocks/block--eic-group-header-block.inc
```diff
@@ -198,12 +196,6 @@ function eic_community_preprocess_eic_group_header_block(array &$variables) {
       }
     }
 
-    // We set the recommend flag in the 2nd position.
-    if ($name === 'recommend_content') {
-      $flags[1] = $flag;
-      continue;
-    }
-
```
4hardfire, line 206: Is this removal intended? I remember this was previously requested by the client, to have the Recommend flag in the second position.

4hardfire (changes requested):

4hardfire:
> > Code seems fine but some steps didn't pass:
> > ```
> > * `As SA, create a Public group and publish it, make sure you can like and share on social media`
> >   When the group is pending I still see the share link. Also, for some reason the like button is never shown even if the group is published
> > 
> > * `Make you sure you cannot like and share on all of them`
> >   After creating a group content, the share button is always shown (even in draft). The like button is ok.
> > 
> > * `Make sure you cannot like and share the Group, Document, Discussion, Video, Gallery and Event`
> >   The like button is never shown in the group header (even when group is published) and the share button is always shown independently of the group status. For the group content, the like button is ok but the share button is always shown independently of the node moderation state or group moderation state.
> > ```
> 
> @4hardfire , there was some code messing up tieh flags, so the like was always removed.
> 
> I fixed the otcher checks.
> 
> Can you re-test?

This is ok!

lib/modules/eic_moderation/src/ModerationHelper.php
4hardfire: ok

PR #2006, opened 2022-10-06
[EICNET-2830]fix: Allow trusted user to create tag on fly

Dromadoros:
How to test:

- [x] As a trusted user create an Event group content from a Group
- [x] Add tag that does not exist and press enter
- [x] Save the new node and check you have no error


lramarojaona (approved):
It works!

PR #2007, opened 2022-10-06
EICNET-2885: On user profile edition Country should be empty not afghanistan by default

4hardfire:
### Fixes

- Apply patch to fix issue with required address fields selecting the first country in the list by default.

### Test

- [x] As SA/SCM, create a new user
- [x] Make sure the country option in the Address field is set to " - Select- "

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1848,7 +1848,8 @@
                     }
                 },
                 "patches_applied": {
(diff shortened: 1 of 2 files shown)


lramarojaona (approved):
It works!

PR #2012, opened 2022-10-07
EICNET-2667: I can still manage members/memberships/invitations of a blocked group

4hardfire:
### Fixes

- Create custom views access plugin to restrict access to based on the group permission and moderation state;
- Apply views custom access plugin in manage members page and manage invitations page;
- Create custom view EIC Group Invitations and replace it with the default one provided by the group module.

### Test

- [x] As GO/GA, make sure you cannot access "Manage invitations" and "Manage members" pages when the group is blocked/pending/refused
- [x] Make sure you cannot edit/delete group members when the group is blocked/pending/refused/archived
- [x] Make sure you cannot access "Manage members" when the group is archived (the logic was already like this, I just improved it and used the custom views access plugin that I created)
- [x] As  SA/SCM, make sure you can always access "Manage invitations" and "Manage members" pages

### DIFF
--- a/config/sync/views.view.eic_group_invitations.yml
+++ b/config/sync/views.view.eic_group_invitations.yml
@@ -1,19 +1,18 @@
-uuid: 7b940c0a-3ee6-4b34-98e5-0e1b43558a92
+uuid: d963a8e7-fc7f-477c-96f2-a79b061a5e81
 langcode: en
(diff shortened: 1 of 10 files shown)


lramarojaona (approved):
Ok, it works!

PR #2013, opened 2022-10-10
Feature/eicnet 2878 contact form

lramarojaona:
### Tests

- [x] As anonymous, go to the global contact form `/contact`
- [x] Fill in the form and submit
- [x] Make sure email is sent to the email defined in the selected category (see `/admin/structure/taxonomy/manage/contact_category/overview`)
- [x] Make sure no email is sent to info@example.org

### Deployment procedure

- Make sure info@example.org is removed after config import, otherwise remove it manually


4hardfire (approved):
It works!
Just a side note: for some reason after running drush cim the config doesn't show up in the console to update. Then I manually removed the recipients from the contact form configurations form using the UI and it worked. I tried to run drush cex and there are no configuration changes so I suppose it's a bug with config import command or something. This means that we might have to do the same in PROD.

lramarojaona:
> It works! Just a side note: for some reason after running drush cim the config doesn't show up in the console to update. Then I manually removed the recipients from the contact form configurations form using the UI and it worked. I tried to run drush cex and there are no configuration changes so I suppose it's a bug with config import command or something. This means that we might have to do the same in PROD.

Ok, thanks, I'll add it in the deployment procedure for Siggy

PR #2020, opened 2022-10-12
Feature/eicnet 2894 fix notification links

lramarojaona:
### Improvements

- Created a default redirect route to My Settings page (useful when we don't know the user ID)

### Tests

- [x] As TU, make sure you have some topics of interest and that your interest notifications are enabled (/user/<uid>/settings)
- [x] As SA, create a Story with the same topic and publish it
- [x] Run cron
- [x] Make sure you receive a notification of this type sub_content_interest_published and that has a link to `<domain-name>/user/settings` when clicking on `Manage subscriptions`

### DIFF
--- a/lib/modules/eic_user/eic_user.routing.yml
+++ b/lib/modules/eic_user/eic_user.routing.yml
@@ -102,3 +102,10 @@ eic_user.get_notification_settings:
   requirements:
     notification_type: 'groups|events|organisations'
     _user_is_logged_in: 'TRUE'
+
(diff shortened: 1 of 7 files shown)


4hardfire (changes requested):
The Manage settings link in the notification works properly.
However, I found multiple issues in the email:

1. The subject contains all the text from the body as well.
2. There are some extra fields shown in the body of the email.

image

I think we should hide the fields in the view modes `Notify - Email body` and `Notify - Email subject`

PR #2021, opened 2022-10-12
EICNET-2461: Sharing content to another group should be filtered on the availability of that CT in that specific group

4hardfire:
### Fixes

- Exclude gallery nodes from being shared in organisations.

### Test

- [ ] As SA/SCM, go to a public group
- [ ] Make sure the group contains galleries
- [ ] Try to share a gallery with another group and make sure you cannot see/select any organisation from the dropdown list.


lramarojaona:
@4hardfire , I'm not sure about that one.
Last discussion we had with the client was to align most of the group types features and I think we should stop creating constraints.
Let's discuss in the ticket.

PR #2022, opened 2022-10-12
EICNET-2745: Closed joining method for events (Rework)

4hardfire:
### Improvements

- Enable closed joining method in Global events + Fix error when saving private groups.

### Test

- [x] As SA/SCM, edit an event
- [x] Make sure the joining method Closed can be selected
- [x] Set group visibility to private
- [x] Make sure there is no error when you save the group


lramarojaona (approved):
Ok, it works!

PR #2024, opened 2022-10-12
EICNET-2758: [Event] No one should be allowed to comment on a draft story

4hardfire:
### Fixes

- Hide comments block and access to block api endpoints when content is in draft or unpublished state.

### Todo

- [x] Move methods from ContentModerationManager to EICModerationHelper after the following PR has been merged #2002

### Test

- [x] As SA/SCM, create a draft News in a global event
- [x] Make sure you cannot post or reply to comments
- [x] Publish the News and make sure you can post or reply to comments now
- [x] Unpublish the News and make sure you cannot post or reply to comments
- [x] Make sure the same behaviour happens with all group content. Whenever the content is in draft, users cannot post or reply to comments.

### DIFF
--- a/lib/modules/eic_groups/src/Controller/DiscussionController.php
+++ b/lib/modules/eic_groups/src/Controller/DiscussionController.php
@@ -15,10 +15,8 @@ use Drupal\Core\File\FileUrlGeneratorInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface;
 use Drupal\eic_comments\CommentsHelper;
-use Drupal\eic_content\Constants\DefaultContentModerationStates;
 use Drupal\eic_flags\RequestStatus;
-use Drupal\eic_groups\EICGroupsHelper;
-use Drupal\eic_moderation\Service\ContentModerationManager;
+use Drupal\eic_moderation\ModerationHelper;
 use Drupal\eic_search\Service\SolrDocumentProcessor;
 use Drupal\eic_user\UserHelper;
 use Drupal\flag\FlaggingInterface;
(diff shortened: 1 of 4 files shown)


lib/modules/eic_groups/src/Controller/DiscussionController.php
```diff
@@ -655,13 +752,15 @@ private function renderCommentElement(CommentInterface $comment): array {
...
+   *   The group content node ID.
    *
-   * @return bool
-   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
-   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   * @return GroupInterface|bool
+   *   The group entity.
    */
-  private function isGroupArchived(int $node_id): bool {
+  private function loadGroupFromNode(int $node_id): bool {
```
lramarojaona, line 763: We already have a similar function in `Drupal\eic_groups\EICGroupsHelper::getGroupFromContent()`

lramarojaona (changes requested):
It works, but I have 2 remarks.

lib/modules/eic_groups/src/Controller/DiscussionController.php
4hardfire: Done!

lib/modules/eic_groups/src/Plugin/Block/CommentsFromDiscussionBlock.php
```diff
@@ -134,22 +142,24 @@ public static function create(
    *   The plugin_id for the plugin instance.
    * @param mixed $plugin_definition
    *   The plugin definition.
-   * @param \Drupal\eic_groups\EICGroupsHelper $groups_helper
+   * @param EICGroupsHelper $groups_helper
```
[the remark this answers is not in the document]
4hardfire: Fixed!

lramarojaona (approved):
Ok, it works!

PR #2027, opened 2022-10-13
EICNET-2637: When going to the wiki tab inside a global event the breadcrumbs say group

4hardfire:
### Fixes

- Fix breadcrumb when viewing wiki page of a global event.

### Test

- [x] As GM, go to a wiki page of a global event
- [x] Make sure the breadcrumb shows `Home > ... > Events >` and not `Home > ... > Groups >`

### DIFF
--- a/lib/modules/eic_groups/src/EICGroupsHelper.php
+++ b/lib/modules/eic_groups/src/EICGroupsHelper.php
@@ -1144,4 +1144,37 @@ class EICGroupsHelper implements EICGroupsHelperInterface {
     return $config->get('default_features') ?? [];
   }
 
+  /**
+   * Gets group bundle label.
+   *
+   * @param \Drupal\group\Entity\GroupInterface $group
+   *   The group entity.
+   * @param bool $return_plural
+   *   Return as plural.
+   *
+   * @return string
+   *   The group label.
+   */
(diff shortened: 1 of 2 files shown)


lramarojaona:
@4hardfire , The bradcrumb shows "Events" but still links to `/groups`

4hardfire:
> @4hardfire , The bradcrumb shows "Events" but still links to `/groups`

Should be fixed. Can you re-test?

lib/modules/eic_groups/src/EICGroupsHelper.php
```diff
@@ -1144,4 +1144,37 @@ public static function getGroupDefaultFeatures(Group $group) {
...
+   *
+   * @param \Drupal\group\Entity\GroupInterface $group
+   *   The group entity.
+   * @param bool $return_plural
+   *   Return as plural.
+   *
+   * @return string
+   *   The group label.
+   */
+  public function getGroupBundleLabel(GroupInterface $group, bool $return_plural = FALSE) {
```
lramarojaona, line 1158: That's nice!

lramarojaona (approved):
Ok, it works!

lib/modules/eic_groups/src/EICGroupsHelper.php
4hardfire: It might be useful in some other places as well :).

PR #2028, opened 2022-10-13
EICNET-2753: Remove email from entity user reference view.

lramarojaona:
### Tests

- [x] As TU, create a story
- [x] Try to add a platform user as Contributor
- [x] Make you you don't get a 500 error and get some results shown

### Remarks

- This is a workaround since we removed the permission for TUs to view other users' email address


4hardfire (approved):
It works!

PR #2032, opened 2022-10-17
[EICNET-2368]chore: Display visibility label for events

korhankale:
- [x] Update a global event and make it restricted
- [x] The restricted label should be display on the event's teaser in the event overview
- [x] The restricted label should be display on the group header


lramarojaona (approved):
Ok, it works!

PR #2033, opened 2022-10-17
EICNET-2898: Validate user entity and send out custom message if not valid.

lramarojaona:
### Tests

With Postman.

- [x] Try to create a user with missing `name` value. You should get a message: `Unprocessable Entity: validation failed. Account name is empty.`
- [x] Try to create a user with missing `mail` value. You should get a message: `Unprocessable Entity: validation failed. Email address is empty.`


4hardfire (approved):
It works!

PR #2037, opened 2022-10-18
EICNET-XXXX: chore: Fix missing dependency on install.

lramarojaona:
- [x] Make a fresh install
- [x] Make sure you don't get stuck on 
image


4hardfire (approved):
It works!

PR #2040, opened 2022-10-20
Feature/eicnet 2913 org membership job title

lramarojaona:
### Tests

- [x] Go to the organisation manage members page `/organisations/european-innovation-council-eic/members`
- [x] Give some job titles to some members, including Owner and Admin
- [x] Go the organisation detail page
- [x] Make sure you see the members (max 6), ordered by the job title's weight (`/admin/structure/taxonomy/manage/job_titles/overview`)
- [x] Edit the organisation
- [x] Add some items under the `Team` tab
- [x] Go the organisation detail page
- [x] Make sure you the iteams you just added instead of the previous items
- [x] Log in as GO/GA (non power user)
- [x] Edit a membership and make sure you cannot change the job title

### DIFF
--- a/config/sync/core.entity_form_display.group_content.organisation-group_membership.default.yml
+++ b/config/sync/core.entity_form_display.group_content.organisation-group_membership.default.yml
@@ -3,6 +3,7 @@ langcode: en
 status: true
 dependencies:
   config:
+    - field.field.group_content.organisation-group_membership.field_vocab_job_title
     - field.field.group_content.organisation-group_membership.group_roles
     - group.content_type.organisation-group_membership
   module:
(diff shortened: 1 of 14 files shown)


4hardfire (changes requested):
Should we also make sort by taxonomy name alphabetically when the weight field is the same. After a site install the weights are all set to 0 and for some reason the sort was not according to what I was seeing in the taxonomy overview UI.
Can you have a look?

lramarojaona:
> Should we also make sort by taxonomy name alphabetically when the weight field is the same. After a site install the weights are all set to 0 and for some reason the sort was not according to what I was seeing in the taxonomy overview UI. Can you have a look?

You're right, this is fixed.

4hardfire (approved):
It works!

Sigtechnica:
verified it works!

PR #2042, opened 2022-10-20
EICNET-1921: Activity stream shows undefined link for deleted content

4hardfire:
### Improvements

- Create hook_update to delete activity stream messages with references to non-existing content.

### Test

- [x] Import DB from PROD
- [x] Run hook_update
- [x] Make sure the activity stream messages that point to non-existing content were deleted
- [x] Try to create a discussion and post it in the activity stream
- [x] Try to update the discussion and post it in the activity stream
- [x] Go to the latest activity of the group
- [x] Make sure you have 2 items
- [x] Delete the discussion
- [x] Make sure any message related to the discussion is shown in the activity stream
- [x] Check the database table `message__field_referenced_node` to see if there is still 2 references to the deleted node
- [x] Run cron
- [x] Make sure the rows were removed from the database

### DIFF
--- a/lib/modules/eic_messages/eic_messages.install
+++ b/lib/modules/eic_messages/eic_messages.install
@@ -30,7 +30,7 @@ function eic_messages_update_9001(&$sandbox) {
   }
 
   $batch_builder = (new BatchBuilder())
-    ->setTitle(t('Processing Batch to modify group permissions.'))
+    ->setTitle(t('Processing Batch to delete messages with references to non-existing content.'))
     ->setFinishCallback('_eic_messages_batch_finished')
     ->setInitMessage(t('Batch is starting'))
     ->setProgressMessage(t('Processed @current out of @total.'))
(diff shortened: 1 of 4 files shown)


lib/modules/eic_messages/eic_messages.install
```diff
@@ -0,0 +1,58 @@
...
+  $query->leftJoin('node', 'n', 'n.nid = frn.field_referenced_node_target_id');
+  $query->isNull('n.nid');
+  $results = $query->execute()->fetchAllAssoc('mid');
+
+  if (!$results) {
+    return;
+  }
+
+  $batch_builder = (new BatchBuilder())
+    ->setTitle(t('Processing Batch to modify group permissions.'))
```
lramarojaona, line 33: Message should be updated here.

lramarojaona (changes requested):

lib/modules/eic_messages/eic_messages.install
4hardfire: Fixed!

lramarojaona (approved):
Ok, it works!

PR #2048, opened 2022-10-24
Feature/eicnet 2847 story type

lramarojaona:
### Improvements

- Added new taxonomy to content generator

### Tests

- [x] Create a new Story
- [x] As SA, go to the Stories overview page (/admin/community/overview-pages) and enable the "Type" facet
- [x] Go to `/articles` and make sure you have a "Type" facet

### DIFF
--- a/config/sync/core.entity_form_display.node.story.default.yml
+++ b/config/sync/core.entity_form_display.node.story.default.yml
@@ -18,6 +18,7 @@ dependencies:
     - field.field.node.story.field_tags
     - field.field.node.story.field_video
     - field.field.node.story.field_vocab_geo
+    - field.field.node.story.field_vocab_story_type
     - field.field.node.story.field_vocab_topics
     - node.type.story
   module:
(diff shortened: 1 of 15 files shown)


4hardfire (commented):
@lramarojaona For some reason when I tried to save the overview page it saved but the type facet wasn't showing as checked when I try to edit again. I had to choose another source type and then put back the Story source type to reset the configuration of the overview page and then it worked!
What should we do with existing stories since the story type field is mandatory?

lramarojaona:
> @lramarojaona For some reason when I tried to save the overview page it saved but the type facet wasn't showing as checked when I try to edit again. I had to choose another source type and then put back the Story source type to reset the configuration of the overview page and then it worked! What should we do with existing stories since the story type field is mandatory?

@4hardfire , I think this is a known issue, not related to the facet itself. There's nothing to do for old content.

4hardfire (approved):

PR #2054, opened 2022-10-27
EICNET-2736: User account] remove text "explaining" email field since irrelevant

4hardfire:
### Improvements

- Remove helper text from email field in user profile.

### Test

- [x] As SA/SCM, edit a user account
- [x] Make sure the email field doesn't show any helper text


lramarojaona (approved):
It works, but maybe we could make this conditional based on `UserHelper::isSmedUser()`

4hardfire:
> It works, but maybe we could make this conditional based on `UserHelper::isSmedUser()`

@lramarojaona shouldn't this be removed for all users? I don't understand why we need that condition.

lramarojaona:
> isSmedUser

Some users might be non EU Login users. Btw, `UserHelper::isSmedUser()` is not correct. We should probably check through the cas module. Probably `Drupal\cas\Service\CasUserManager::getCasUsernameForAccount()` ?

lramarojaona:
> > It works, but maybe we could make this conditional based on `UserHelper::isSmedUser()`
> 
> @lramarojaona shouldn't this be removed for all users? I don't understand why we need that condition.

@4hardfire , Some users might be non EU Login users. Btw, `UserHelper::isSmedUser()` is not correct. We should probably check through the cas module. Probably D`rupal\cas\Service\CasUserManager::getCasUsernameForAccount()` ?

4hardfire:
> > > It works, but maybe we could make this conditional based on `UserHelper::isSmedUser()`
> > 
> > 
> > @lramarojaona shouldn't this be removed for all users? I don't understand why we need that condition.
> 
> @4hardfire , Some users might be non EU Login users. Btw, `UserHelper::isSmedUser()` is not correct. We should probably check through the cas module. Probably D`rupal\cas\Service\CasUserManager::getCasUsernameForAccount()` ?

@lramarojaona It should be fixed. Can you have a look?

PR #2057, opened 2022-10-27
Feature/eicnet 2714 block node add page

lramarojaona:
### Tests

- [x] Login as SA/SCM
- [x] Make sure you can access `/node/add` page
- [x] Login as TU
- [x] Make sure you canno access `/node/add` page


4hardfire (approved):
It works!

PR #2058, opened 2022-10-27
EICNET-2728: Make topics mandatory for Group GT.

lramarojaona:
### Tests

- [x] As TU, create a group
- [x] Make sure Topics field is mandatory


4hardfire (approved):
It works!

PR #2060, opened 2022-10-28
Feature/eicnet 2926 archived group join

lramarojaona:
### Tests

- [x] As SA invite User A in a group
- [x] As TU User B, request membership to the group
- [x] As SA, archive the group
- [x] As User A, try to accept or decline the invitation
- [x] Check that you get an Access denied with a message: "This %group_type is archived. You cannot perform this action."
- [x] As GO/GA (non power user) of the group, try accept or decline the membership
- [x] Check that you get an Access denied with a message: "This %group_type is archived. You cannot perform this action."

### DIFF
--- a/lib/modules/eic_groups/eic_groups.services.yml
+++ b/lib/modules/eic_groups/eic_groups.services.yml
@@ -39,6 +39,7 @@ services:
       - { name: access_check, applies_to: _orphan_group_content_access_check }
   eic_groups.archived_route_access_checker:
     class: Drupal\eic_groups\Access\ArchivedRouteAccessCheck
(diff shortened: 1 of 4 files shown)


4hardfire (approved):
It works!
Just a small improvement for the future: As GO/GA I never see the button to accept or decline membership request when the group is archived and the Warning message is showing anyway. Maybe it becomes confusing to the user:
image

lramarojaona:
> It works! Just a small improvement for the future: As GO/GA I never see the button to accept or decline membership request when the group is archived and the Warning message is showing anyway. Maybe it becomes confusing to the user: image

You're right, I added a condition to show this message for the current route only.

PR #2062, opened 2022-11-02
EICNET-2664: As SA/SCM, I want an option in the CMS so that I can manage the roles of the platform users

4hardfire:
### Improvements

- Create new permission to administer user roles;
- Allow site administrators to administer user roles except assigning site_admin and administrator roles to users.

### Test

- [x] As SA, try to edit a user account
- [x] Make sure you can only assign the roles "Trusted user", "Service authentication" and "Content administrator"

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -87,6 +87,7 @@
         "drupal/redirect": "^1.6",
         "drupal/redis": "^1.5",
         "drupal/restui": "^1.19",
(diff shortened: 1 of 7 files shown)


4hardfire:
> Instead of creating new code, there is a module for that: https://www.drupal.org/project/role_delegation

Done!

lramarojaona (approved):
Ok, it works!

PR #2064, opened 2022-11-03
EICNET-2749: As TU, I want a visual indication in a group I requested to join so that I can tell my request is pending

4hardfire:
### Improvements

- Prevent users from accessing group_membership_request group when there is already an opened or pending request.

### Todo

- We need to work on the Front end styles when the button changes from "Request membership" to "Request sent". Currently it has the same look and feel but we should show it as if it was disabled (grey background)

### Test

- [x] As SA, create a public group and set the joining method to "Request"
- [x] As TU, go to the group overview page
- [x] Make sure you see the button "Request membership"
- [x] Click on "Request membership" and confirm your request
- [x] Make sure you have been redirected to the group overview page and you see the button "Request sent"
- [x] As SA, go to the list of membership requests of the group -> `/groups/<group-name>/members-pending`
- [x] Reject the membership request from the TU
- [x] As the TU, make sure you can see the button "Request membership" and also make sure you are able to request group membership again

### DIFF
--- a/lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
+++ b/lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
@@ -281,7 +281,6 @@ class EICGroupHeaderBlock extends BlockBase implements ContainerFactoryPluginInt
             'value' => 'disabled',
             ],
           ],
-          'disabled' => TRUE,
         ];
       }
     }
--- a/lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
+++ b/lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
@@ -266,7 +266,7 @@ class EICGroupHeaderBlock extends BlockBase implements ContainerFactoryPluginInt
 
     $joining_method = $this->eicGroupsHelper->getGroupJoiningMethod($group);
     if ($joining_method === 'tu_group_membership_request') {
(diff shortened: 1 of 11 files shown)


lenglebert:
To do by Alexis

AlexisWalravens:
@4hardfire I setup your branch and requested access to a group, but now only the join button is shown.
Is there something I need to do besides the `make setup` ?

image

lib/modules/eic_groups/src/Plugin/Block/EICGroupHeaderBlock.php
```diff
@@ -260,6 +264,20 @@ public function build() {
...
     }
 
+    $joining_method = $this->eicGroupsHelper->getGroupJoiningMethod($group);
+    if ($joining_method === 'tu_group_membership_request') {
+      $cacheable_metadata->addCacheTags(["membership_request:{$this->currentUser->id()}"]);
+
+      // Shows the "Request sent" button if the user already request group membership.
+      if (!$membership && $has_sent_membership_request) {
+        $operation_links[] = [
+          'title' => $this->t('Request sent', [], ['context' => 'eic_groups']),
```
arismag: According to https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2749 the "Request sent" should change to "Pending approval"
image

arismag:
We should style the button when pending like https://ec.europa.eu/component-library/v3.13.0/playground/ec/?path=/story/components-button--ghost&args=disabled:!true

arismag (approved):

arismag:
image

PR #2068, opened 2022-11-07
EICNET-2923: Fix processing of discussions.

lramarojaona:
### Tests

- [x] Make sure you have one disussion with comments and one without
- [x] Make sure the discussion authors and commenters all have a user picture
- [x] Update both discussions
- [x] Go to /groups/<group-name>/discussions
- [x] Make sure you see the author picture for both discussions and the commenter picture for the one that has comments


4hardfire (approved):
It works!

PR #2109, opened 2023-06-21
[EICNET-2964]fix: Index for a group the correct last activity timestamp and not always the changed field

Dromadoros:


lramarojaona (approved):
Ok, it works

PR #2110, opened 2023-06-22
[EICNET-2843]feat: Set digest frequency daily by default

Dromadoros:


lramarojaona (approved):
Ok, Daily is now the default

PR #2111, opened 2023-06-22
[EICNET-2953]feat: If user is unsubscribed or blocked do not send email

Dromadoros:


lramarojaona (commented):
@Dromadoros ,
Does this apply to both digests _AND_ notifications?

PR #2115, opened 2023-07-03
[EICNET-2967]feat: implement sensitive visibility + solr

Dromadoros:
- Added a new visibility type for group: Sensitive
- A new role: Sensitive. User with role sensitive can see sensitive group
- You can only be member of sensitive group by invitation
- Group/node/download access only if you are allow to see sensitive content


Sigtechnica:
code looks ok, will merge and we will test on our test server

PR #2117, opened 2023-07-10
[EICNET-2955]fix: If no joining method available for event, do not show empty box

Dromadoros:
- [ ] Set an event as closed joining method and "public" visibility
- [ ] Go on the event page as anonymous
- [ ] Be sure you do not have empty box in the header

image


lramarojaona:
Not sure why, but I can't access the event page (as anonymous and TU) as soon as the joining method is set to Closed.

lramarojaona (approved):
Ok, I did a workaround locally and now it works.

PR #2128, opened 2023-07-14
[EICNET-2903]feat: Do not show blocked or removed user in email digest

Dromadoros:
- [ ] Create a content in a group with a certain user
- [ ] Block this user
- [ ] Go to Test digest UI (/admin/config/eic/digest)
- [ ] Run digest for a user that is in the group and will receive the node created by a user
- [ ] Verify that you cannot see anymore the header with user on mail notification

image


lramarojaona (approved):
Perfect, it works!

PR #2130, opened 2023-07-14
Feature/eicnet 2835 latest news stories

lramarojaona:
### Tests

- [ ] Create a News or a Story
- [ ] Make sure it appears first in the homepage in the "All News & Stories" section
- [ ] Edit an old Story or News and make it "Sticky"
- [ ] Make sure it appears first on the homepage

### To do

- [ ] There's something wrong with the display if you compare it to the old block
- [ ] Remove the old block from the homepage


Sigtechnica:
talked with ludovic and should be ok, just to remove the old block

PR #2142, opened 2023-07-25
[EICNET-2972]feat: Add queue to remove items + check if there is media to remove + sensitive permission

Dromadoros:
How to test:

**Delete group and his content by queue**
- [x] Get a group where you have a lot of content in it
- [x] Delete the group (keep an eye about some content title to see if it has not been removed directly)
- [x] Normally content should not be delete on the group delete action
- [x] Wait for cron or trigger it (drush core:cron)
- [x] Verify content has been removed

**Sensitive deletion**
- [x] Change a group visibility to sensitive
- [x] Verify you can still delete the group

**Delete media when it's only used once**
- [x] Add a document to a group with a *new* media in it
- [x] Delete the group
- [x] Run cron
- [x] Verify document has been removed and also the media

### DIFF
--- a/lib/modules/eic_groups/src/Plugin/QueueWorker/RemoveGroupContentApi.php
+++ b/lib/modules/eic_groups/src/Plugin/QueueWorker/RemoveGroupContentApi.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace Drupal\eic_groups\Plugin\QueueWorker;
+
+use Drupal\Core\Queue\QueueWorkerBase;
+use Drupal\media\MediaInterface;
+use Drupal\node\Entity\Node;
+use Drupal\node\NodeInterface;
+
+/**
+ * Defines 'eic_group_remove_content' queue worker.
+ *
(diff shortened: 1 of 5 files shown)


lib/modules/eic_groups/src/Plugin/QueueWorker/RemoveGroupContentApi.php
```diff
@@ -0,0 +1,53 @@
...
+   * {@inheritdoc}
+   */
+  public function processItem($data) {
+    $node = Node::load($data);
+
+    if (!$node instanceof NodeInterface) {
+      return;
+    }
+
+    if ($node->bundle() === 'document') {
```
lramarojaona, line 31: @Dromadoros , here I think we should be generic:

- Search for any entity reference field
- For each Media entity reference, look if it's being used somewhere else
- If not, delete the media

lramarojaona (changes requested):
@Dromadoros ,
It works, but please see my comment

PR #2157, opened 2024-02-06
EICNET-2989

arismag:
The library is attached only on the active admin theme:

image
image
image

On the frontend theme, the library is not attached:
image


karimkancel:
To be tested in DEV

PR #2164, opened 2024-02-19
EICNET-2960: Add Youtube channel in "Follow Us" block

arismag:
Add youtube option on social media follow links and add corresponding styling option.


arismag:
In addition, these PR request manual steps for adding the youtube link to the custom social block on every instances. There are instructions on the ticket https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2960 on the Proposed solution sections.

karimkancel (approved):

PR #2184, opened 2024-04-05
[EICNET-2946]fix: If group is archived do not notified invitation

arismag:
This was the original solution.


arismag:
This was a temp PR in order to resolve a problem with the merge conflicts. Correct one for this ticket in: https://github.com/EIC-EA/EIC-community-D8/pull/2185

PR #2186, opened 2024-04-10
EICNET-2997

dwb17:
Create a custom module in the steps of the https://www.drupal.org/project/require_on_publish that will allow user to submit an incomplete story or news item as long as the entity is in draft state.

This issue also covers issues
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2872
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/EICNET-2756


arismag:
@dwb17 The validation works mostly as expected. However some thinks to note:

- Minor: Not sure if should also be allowed to save as unpublished.
- Minor: When disabling the module maybe the default required field should be used as a fallback.
- I think the same think should happen regardless the content type.
- **Most important:** There is no indication on the fields label that the field is required. You are aware of what fields are required only on submission action, which is not a good UX scenario and not how the module require_on_publish works, since it does handle this with different icon as shown in the screenshot below. 

I think that just by having the asterisk in a dark orange color to indicate that the field is still a required field but can be saved as draft.  And I guess the solution would still need to work for the group fields as well. If there is a way to still pass the required or form-required classes of the default Required functionality or new ones (ex. non-draft-required).

Let's discuss this further.

image

Also there was a discussion about the limit_validation_errors. I think this is how webforms handles the save draft and back actions. Dunno how relevant if for our case, I am leaving it here for the reference.

arismag (approved):
All good! I just changed the asterisk color to blue as on the required_on_publish module. The difference wasn't so clear as seen in the screenshots below with the initially proposed color. 

Screenshot_20240416_151152
Screenshot_20240416_151118

arismag:
After discussing it with Karim, we should apply the same to all required fields of "News" content type.

PR #2187, opened 2024-04-11
EICNET-2725: Revoke permission to join organization for SA and SCM

arismag:


arismag:
Still need to write the update hook for the permissions.

arismag:
@dwb17 the revoke is inspired from https://git.drupalcode.org/project/flood_control/-/blob/334a7737157923d897b444eb435a9329a64db66b/flood_control.install

PR #2189, opened 2024-04-17
EICNET-2740: SMED Webservice: Allow SMED to update the SMED ID of a user

dimitriskr:


arismag:
@karimkancel hey Karim,  I've assigned it to you since this requires SMED to be tested, ;)

PR #2190, opened 2024-04-17
EICNET-2995: re-implement original way to manage news & stories

dimitriskr:


arismag:
We are waiting a commit of @Sigtechnica that resolves a conflict in the config.

PR #2199, opened 2024-05-20
EICNET-2599: Add 'Countries & Regions' in story content type full view mode

arismag:
EICNET-2599: Add 'Countries & Regions' in story content type full view mode

### DIFF
--- a/lib/themes/eic_community/includes/preprocess/content/node--news--story.inc
+++ b/lib/themes/eic_community/includes/preprocess/content/node--news--story.inc
@@ -357,7 +357,7 @@ function _eic_community_display_geo(&$variables) {
     if (is_int($key)) {
       /** @var \Drupal\taxonomy\Entity\Term $term */
       $term = $value['#options']['entity'];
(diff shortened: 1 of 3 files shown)


lib/themes/eic_community/includes/preprocess/content/node--news--story.inc
```diff
@@ -346,6 +347,42 @@ function _eic_community_display_topics(&$variables) {
...
+ * Preproccess Countries and Regions for full news/story.
+ */
+function _eic_community_display_geo(&$variables) {
+  $geo = [];
+
+  foreach ($variables['elements']['field_vocab_geo'] as $key => $value) {
+    if (is_int($key)) {
+      /** @var \Drupal\taxonomy\Entity\Term $term */
+      $term = $value['#options']['entity'];
+      $is_parent = $term->get('parent')->target_id == 0 ?? false;
```
dwb17: `false` should be `FALSE` when written in PHP

arismag: ooooooops! You are right, thanks! Fixed!

PR #2201, opened 2024-05-21
EICNET-3072: change token group_content:my_invitations_link to site:e…

dimitriskr:
…ic-user-invitations-page-url


arismag:
@Sigtechnica hey man! What's the status of this fix? I think you checked it with Dimitris, is thould have the green light for testing/production right?

PR #2212, opened 2024-06-19
EICNET-3081: Replace group like text stats with appropriate icon for anonymous users

arismag:
EICNET-3081: Replace group like text stats with appropriate icon for anonymous users


arismag:
image

PR #2292, opened 2025-01-22
EICNET-3124-and-EICNET-3131

arismag:
…r/jsonpath to 0.8.


Sigtechnica:
I ran locally on my laptop a fresh install : 

- make setup, was successful
- make update also successful

PR #2329, opened 2025-04-11
EICNET-3256: User profile completion status

dimitriskr:


lib/modules/eic_dashboards/src/Plugin/views/field/UserProfileCompletionStatus.php
```diff
@@ -0,0 +1,57 @@
...
+
+use Drupal\Core\Database\Connection;
+use Drupal\eic_user\UserHelper;
+use Drupal\user\UserInterface;
+use Drupal\views\Plugin\views\field\FieldPluginBase;
+use Drupal\views\ResultRow;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Field handler to flag the node type.
```
dwb17, line 12: Update the text here.

lib/modules/eic_dashboards/src/Plugin/views/field/UserProfileCompletionStatus.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace Drupal\eic_dashboards\Plugin\views\field;
+
+
+use Drupal\Core\Database\Connection;
```
dwb17, line 6: Are you using this class, remove otherwise

lib/modules/eic_dashboards/src/Plugin/views/field/UserProfileCompletionStatus.php
```diff
@@ -0,0 +1,57 @@
...
+    // Leave empty to avoid a query on this field.
+  }
+
+  /**
+   * @inheritdoc
+   */
+  public function render(ResultRow $values) {
+    $user = $values->_entity;
+    if ($user instanceof UserInterface) {
+      $is_completed = $this->userHelper->isUserProfileCompleted($user);
```
dwb17: Are we going to check on this, as it might be expensive to test for each record if the entity has validation errors. I thought we were going to provide a light version of isUserProfileCompleted, that would perform a SQL query on specific fields to ensure they are populated.

dwb17 (changes requested):

PR #2336, opened 2025-04-28
EICNET-3187

arismag:


lib/modules/eic_dashboards/src/Services/MembersStatistics.php
```diff
@@ -77,6 +78,79 @@ public function getTotalMembers(): array|int {
...
+      ];
+    }
+
+    return $members;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getPlatformMembersLoggedPastDays(int $days): int {
```
dwb17: I believe a default value should be specified

lib/modules/eic_dashboards/src/Services/MembersStatistics.php
```diff
@@ -77,6 +78,79 @@ public function getTotalMembers(): array|int {
...
+    $timestamp = strtotime("-$days days");
+
+    $query = $this->entityTypeManager->getStorage('user')->getQuery();
+
+    return $query->condition('status', 1)
```
dwb17: I do not think I saw the direct query being returned in other methods. We should have a consistent way of returning.

PR #2349, opened 2025-05-12
EICNET-3343: Accumulative dashboards 

dimitriskr:


lib/modules/eic_dashboards/eic_dashboards.deploy.php
```diff
@@ -0,0 +1,208 @@
...
+    $sandbox,
+    $entity_query,
+    50,
+    $entity_type_id,
+    DashboardsDatabase::DISCUSSIONS_DASHBOARD_TYPE
+  );
+
+}
+
+function _eic_dashboards_populate_database_batch_helper(array &$sandbox, QueryInterface $entity_query, $entities_per_batch, $entity_type_id, $dashboard_type) {
```
dwb17, line 166: Add a phpdoc to this function.
It is odd that the batch_size is a parameter, is there a reason for that?

lib/modules/eic_dashboards/eic_dashboards.deploy.php
```diff
@@ -0,0 +1,208 @@
...
+  $ids = $entity_query
+    ->range($sandbox['current'], $entities_per_batch)
+    ->execute();
+  if (empty($ids)) {
+    $sandbox['#finished'] = 1;
+    return;
+  }
+
+  foreach ($ids as $id) {
+    $entity = \Drupal::entityTypeManager()->getStorage($entity_type_id)->load($id);
```
dwb17, line 188: Would it be possible to use a direct SQL query? It would speed up the whole process immensely.

lib/modules/eic_dashboards/eic_dashboards.install
```diff
@@ -21,3 +22,115 @@ function eic_dashboards_update_10001(&$sandbox) {
...
+
+    ];
+    $schema_api->createTable('eic_dashboards', $schema);
+  }
+}
+
+/**
+ * Implements hook_update_N().
+ */
+function eic_dashboards_update_10003(&$sandbox) {
```
dwb17, line 74: Why is this needed, it is not the first time we're creating the table.

lib/modules/eic_dashboards/src/Services/DashboardCumulativeService.php
```diff
@@ -0,0 +1,234 @@
...
+    $data = [];
+    foreach ($results as $key => $row) {
+      $data[$key]['id'] = $this->dateFormatter->format(strtotime($row->date), 'custom', 'M Y');
+      $data[$key]['count'] = (int) $row->count;
+    }
+
+    return $data;
+  }
+
+  public function fn($bundle, $startDate, $endDate, $dashboard_type, $range = 10) {
```
dwb17, line 171: Need a phpdoc here, not sure what is happening in this function.

PR #2357, opened 2025-05-27
EICNET-3364: accumulative count dashboards count fix

dimitriskr:


lib/modules/eic_dashboards/src/Drush/Commands/GenerateCumulativeStatsMonthDrushCommands.php
```diff
@@ -78,137 +78,146 @@ public function generateAllStatsPreviousMonth() {
...
-    if ($options['all-stats']) {
-      $dashboard_type_to_generate = $this->dashboardCumulativeService->getAllDashboardTypes();
-    }
-    else {
-      if (in_array($options['dashboard-type'], $this->dashboardCumulativeService->getAllDashboardTypes())) {
-        $dashboard_type_to_generate = [$options['dashboard-type']];
+    $this->io()->warning("This action is destructive. Data will be calculated differently.");
+    $confirm = $this->confirm("Are you sure you want to proceed?");
+    if ($confirm) {
+
```
dwb17, line 84: Empty lines

lib/modules/eic_dashboards/src/Drush/Commands/GenerateCumulativeStatsMonthDrushCommands.php
```diff
@@ -78,137 +78,146 @@ public function generateAllStatsPreviousMonth() {
...
+          else {
+            $dashboard_results[$monthKey]++;
+          }
         }
-        else {
-          $dashboard_results[$monthKey]++;
+        foreach ($dashboard_results as $monthKey => $count) {
+          $this->dashboardCumulativeService->insertOrUpdate($dashboard_type, $monthKey, $count);
         }
+        $this->dashboardCumulativeService->calculatePastStats($dashboard_type);
```
dwb17, line 219: Is this correct, I though we would remove this or no?

dimitriskr: > Is this correct, I though we would remove this or no?

In this command, I don't think so. We can, but we didn't say it would change here

dwb17 (approved):

PR #2360, opened 2025-06-16
EICNET-3170: Add node total views on page header

arismag:
EICNET-3170: [Story] Number of views of an article visible for each article


arismag:
Changes made in

    Total view added on page header
    Icon changed on follow flag
    Width of hero header has changed in order the new info fits in well.

image

PR #2380, opened 2025-09-04
EICNET-3386: Onboarding automations

dimitriskr:


config/sync/views.view.test.yml
dwb17, line 1: I think you need to rename this view

PR #2392, opened 2025-10-24
EICNET-3162: create a command to manually update the group permission…

dimitriskr:
…s with the given arguments and options


dimitriskr:
I have created a Drush command that can be used in general, that can add or remove permissions in a role, on the group_permission entity

For example, the following command

`drush eic_groups:update-permissions organisation member --add="edit organisation description" `
will load all group permission entities for 'organisation' group type and add the "edit organisation description" permission to role "organisation-member"

Respectively it can remove a permission by adding `–remove="permission string"`

PR #2405, opened 2026-03-17
EICNET-3500 - Resync code changes / improvements from DDP

Sigtechnica:


Sigtechnica:
Will merge this request and then we will we what should not be part of the core.

